PR #1, opened 2019-08-22
CIP1: Validator-run attestation service

nambrot:
The [lightweight identity protocol](https://docs.celo.org/celo-codebase/protocol/identity) allows users to map (the hash of) their phone numbers to account addresses, greatly improving the user experience of sending value on the Celo platform.

To do so in a decentralized manner, we rely on validators running what we currently call a verification pool to observe attestation requests and fullfill them by sending them via a text message provider like Twilio. However, the service we created expects celo-blockchain to do the decryption and attestation message signing which may not be ideal. Further, the verification pool should be agnostic to the infrastructure it is run on, whereas now it is tightly coupled to Google Cloud Platform products Firebase Cloud Function and Firebase Database.

This CIP outlines work to get the identity protocol to a more realistic production setting to possibly test on a Games of stakes betanet.


CIPs/0001.md
```diff
@@ -0,0 +1,48 @@
+# CIP 1: Validator-run attestation service
+
+- Date: 2019-08-19
+- Author: @nambrot
+
+### Overview
+
+The [lightweight identity protocol](https://docs.celo.org/celo-codebase/protocol/identity) allows users to map (the hash of) their phone numbers to account addresses, greatly improving the user experience of sending value on the Celo platform.
+
+To do so in a decentralized manner, we rely on validators running what we currently call a verification pool to observe attestation requests and fullfill them by sending them via a text message provider like Twilio. However, the service we created expects celo-blockchain to do the decryption and attestation message signing which may not be ideal. Further, the verification pool should be agnostic to the infrastructure it is run on, whereas now it is tightly coupled to Google Cloud Platform products Firebase Cloud Function and Firebase Database.
```
asaj: Attestation pool?

CIPs/0001.md
```diff
@@ -0,0 +1,48 @@
...
+- Get closer to a realistic production setting
+- Make the validator experience of providing attestations as smooth as possible
+- Make the verification process reliable and performant
+
+### Proposed Solution
+
+We propose the creation of a service called the attestation service. Its features would be as follows:
+
+**Accept Reveals via HTTPS**
+We currently have a reveal transaction whose sole purpose it is to encrypt an phone number (identifier) to the validator. That is unncessary to do on-chain. While this technically opens up the protocol to DDos attacks, we think the trade-off is worth it, as it will remove one of three transactions in the whole process, thus speed it up, and reduce gas costs. Validators need to register their service endpoint on the Attestations Contract
```
asaj: nit: "sole purpose it is to communicate an encrypted phone number to"...

asaj: I wouldn't say this opens the protocol up to DDoS, but instead that we will no longer be able to leverage the DDoS prevention built in to the blockchain layer for this particular aspect of attestations

asaj: Why register this with `Attestations` rather than `Validators`?

CIPs/0001.md
```diff
@@ -0,0 +1,48 @@
...
+
+**Specify a separate encryption key**
+We currently use the consensus key of a validator to both decrypt the phone number and sign the attestation message. Consensus carries much more weight than Attestations, so we should separate the keys to reflect that. We would need slight modifications to the Attestations contract to support the signing with the encryption key as set by the validator in the contract.
```
asaj: Same question here. Would this fit better in the `Validators` contract? The lines seem to be a bit blurred

TEMPLATE.md
```diff
@@ -0,0 +1,40 @@
...
+Aim for anywhere from a few sentences to a couple paragraphs. The goal is to provide background (i.e. what is the current state of things) and a description of the problem that this design doc is addressing.
+
+### Goals
+
+- A bulleted list of goals.
+
+
+### Proposed Solution
+
+Explains in detail the proposed solution. Should be the longest section of the doc and contain one or more subsections as needed. way that lends itself to many different hosting setups, reflecting the diversity of our validators and their setups.
```
asaj: ??

TEMPLATE.md
```diff
@@ -0,0 +1,40 @@
...
+
+If applicable, discuss alternative approaches and why they were not suitable.
+
+### Open Questions
+
+If applicable, discuss any unsolved issues that the author may wish to seek feedback on
+
+### Future Work
+
+If applicable, describe future work implied by this doc (e.g. the “Simplifications For v0.1” section in the initial identity spec)
```
asaj: Remove the example here

CIPs/0001.md
nambrot: As per usual, always taking naming suggestions. Right now, I'm suggesting to call it attestation service, as the pool part would be happening within #2

CIPs/0001.md
nambrot: This specific URL seems specific to the mechanics of the `Attestations` contract to letting an "entity" provide attestations. While those entities are currently Validators, I see a future where there will be alternate attestation issuers like NGOs, IGOs, etc.

CIPs/0001.md
nambrot: I agree the lines seem to be blurred. I believe we already have an encryption key on Attestations (for encrypting transaction comments), and I'm not sure we need to increase the number of keys, i.e. we should be able to just reuse this key? It's also less work because that's what we already do?

TEMPLATE.md
nambrot: Oops

CIPs/0001.md
nambrot: Though now that I'm thinking about this, this makes a ton of sense to pair with the #4 and actually store the key along side the service URL in the metadata of #4? @timmoreton

CIPs/0001.md
```diff
@@ -0,0 +1,48 @@
...
+- Get closer to a realistic production setting
+- Make the validator experience of providing attestations as smooth as possible
+- Make the verification process reliable and performant
+
+### Proposed Solution
+
+We propose the creation of a service called the attestation service. Its features would be as follows:
+
+**Accept Reveals via HTTPS**
+We currently have a reveal transaction whose sole purpose it is to communicate an encrypted phone number (identifier) to the validator. That is unncessary to do on-chain. While we will no longer be able to leverage the DDoS prevention built in to the blockchain layer, we think the trade-off is worth it, as it will remove one of three transactions in the whole process, thus speed it up, and reduce gas costs.
```
nambrot: We have to somehow communicate the service endpoint URL, imo we should reuse the metadata endpoint from #4, what do you think @timmoreton

nambrot:
Considering it merged as we are working on it!

PR #4, opened 2019-09-16
[WIP] CIP 3: Claims and metadata

timmoreton:
* A proposal for linking addresses with real-world identities that can be verified off chain
* A proposal for storing metadata for principals such as validators on a website linked in this way

### DIFF
--- a/CIPs/0003.md
+++ b/CIPs/0003.md
@@ -1,0 +1,134 @@
+# CIP [0003]: Identity Claims and Metadata
+
+- Date: 2019-09-19
+- Author: @timmoreton
+- Status: WIP
+
+## Overview
+
+This CIP describes a general mechanism that allows principals on the Celo platform to claim real-world identities and maintain rich, unverified metadata. Metadata may be public, or optionally, encrypted and accessible only to the Celo Foundation. Third parties can verify these identity claims, and use the associated unverified metadata as appropriate. 
+
+There are a number of principals that provide resources to or deploy services on the Celo platform that may wish to assert identities or provide metadata. These include, but are not limited to:
+* Validators
+* Validator groups
+* Smart contract owners
+* DApp developers
+* Creators of Governance proposals
+* Oracles 
+
+Use cases include:
+* Tools want to present public metadata supplied by a validator or validator group as part of a list of candidate groups, or a list of current elected validators.
+* Governance Explorer UIs may want to present public metadata about the creators of governance proposals
+* The Celo Foundation receives notice of a security vulnerability and wants to contact elected validators to facilitate them to make a decision on applying a patch.
+* A DApp makes a request to the Celo Wallet for account information or to sign a transaction. The Celo Wallet should provide information about the DApp to allow the user to make a decision whether to accept the transaction or not.   
+
+## Goals
+
+- Integrate with the existing Attestations structure, rather than maintain separate identifer fields in other contracts, so that the same identity claims and metadata can be linked from any of a principals' activities  
+- Allow principals to make an extensible range of identity claims, and support clients verifying the veracity of these 
+- Allow principals to supply unverified extensible, structured metadata off-chain
+- Allow tools to choose what metadata they require and how they present it
+- Support both public (unencrypted) metadata and metadata accessible only to the Celo Foundation
+
+## Proposed Solution
+
+### Claims
+
+The Attestations contract is extended as follows:
+*  Add a `struct Claim` that comprises: 
+   * `service` - Required. Initially supported values for `service` include `primary`, `github`, and `twitter`.
+   * `identifier` - Optional. The identifier being claimed on the service being claimed.  
+   * `data` - Optional. Additional data, e.g. a 
+   URL, to locate a linked signed reference to the claim. 
+* A field array `Claim[] claims` is added to the `Account` struct. 
+
+All claims are optional.
+
+### Primary Claims and Public Metadata
+
+For claims with `service` set to `primary`, `identifier` is a domain name. `data` is an optional path prefix, such that `https://<identifier>/<data>/identity.json` is a file in the following format:
+
+```
+{
+  "address": "<address>",
+  "claim": {
+    "service": "primary",
+    "identifier": "..",
+    "ctime": 1231222312,
+    "expiry" : 31536000, 
+  },
+  "signature": "<sig>",
+}
+```
+
+TODO - detail exact format. 
+
+Clients should verify the SSL/TLS certificate against a chain of trust and check it was issued for the domain and `identifier` value in question. 
+
+Public metadata for a principal with a `primary` claim may be obtained by first verifying the claim then retrieving the resource at `<url>/<address>-metadata.json`.
+
+Example structure of this file:
+
+```
+{
+  "address" : "<address>",
+  "unvalidated": {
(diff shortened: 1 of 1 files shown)


CIPs/0003.md
```diff
@@ -0,0 +1,134 @@
...
+    "service": "primary",
+    "identifier": "..",
+    "ctime": 1231222312,
+    "expiry" : 31536000, 
+  },
+  "signature": "<sig>",
+}
+```
+
+TODO - detail exact format.
```
nambrot: To remove?

CIPs/0003.md
```diff
@@ -0,0 +1,134 @@
...
+- Support both public (unencrypted) metadata and metadata accessible only to the Celo Foundation
+
+## Proposed Solution
+
+### Claims
+
+The Attestations contract is extended as follows:
+*  Add a `struct Claim` that comprises: 
+   * `service` - Required. Initially supported values for `service` include `primary`, `github`, and `twitter`.
+   * `identifier` - Optional. The identifier being claimed on the service being claimed.
```
nambrot: I would be in favor of encoding the `service` into the identifier itself. So for phone numbers it would be `tel://+49301112223333`, twitter it would be `twitter://nambrot` and for domains it would be `dns://nambrot.com`, because for attestations, issuers could just attest to the identifier without having to separately specify the service.

CIPs/0003.md
```diff
@@ -0,0 +1,134 @@
...
+   * `identifier` - Optional. The identifier being claimed on the service being claimed.  
+   * `data` - Optional. Additional data, e.g. a 
+   URL, to locate a linked signed reference to the claim. 
+* A field array `Claim[] claims` is added to the `Account` struct. 
+
+All claims are optional.
+
+### Primary Claims and Public Metadata
+
+For claims with `service` set to `primary`, `identifier` is a domain name. `data` is an optional path prefix, such that `https://<identifier>/<data>/identity.json` is a file in the following format:
```
nambrot: Related to the above comment, what was the reasoning to particularly denote a DNS claim as primary?

nambrot: Both Google and Keybase look like the prove ownership of a domain via TXT Record. 

I wonder if they do that because it might be technically more secure than http(s), as some websites might allow users to host data at controllable paths? A mitigation could be what Let's encrypt and DNT do, to use [RFC 5785](https://tools.ietf.org/html/rfc5785) with `domain/.well-known`. Fun fact stellar also uses that folder https://www.stellar.org/developers/guides/concepts/stellar-toml.html

Or because a naked top level domain might prescribe users more infrastructure demands with a webserver than simply editing a TXT record? On the other hand, making HTTP calls is probably easier than querying for TXT records, though there are now DNS over HTTPS servers, so it would actually preserve privacy a little better this way.

nambrot: Mostly as a personal anecdote, but with my personal setup at nambrot.com, it would kind of a pain to host it on my specific path. I'd rather prefer to be able to specify under which path (any domain) I would like to host my public metadata at.

It separates the concern of proving a claim from hosting data with the identity of that claim.

CIPs/0003.md
```diff
@@ -0,0 +1,134 @@
...
+  },
+  "signature": "<sig>",
+}
+```
+
+TODO - detail exact format. 
+
+Clients should verify the SSL/TLS certificate against a chain of trust and check it was issued for the domain and `identifier` value in question. 
+
+Public metadata for a principal with a `primary` claim may be obtained by first verifying the claim then retrieving the resource at `<url>/<address>-metadata.json`.
```
nambrot: Do we need the metadata to be hosted
1. Under the actual domain
2. Under a specified path

Mostly as a personal anecdote, but with my personal setup at nambrot.com, it would kind of a pain to host it on my specific path. I'd rather prefer to be able to specify under which URL (any domain) I would like to host my public metadata at. What do you think about just adding a `metadata` field to the `Account` struct

It separates the concern of proving a claim from hosting data with the identity of that claim.

CIPs/0003.md
```diff
@@ -0,0 +1,134 @@
...
+- Allow principals to make an extensible range of identity claims, and support clients verifying the veracity of these 
+- Allow principals to supply unverified extensible, structured metadata off-chain
+- Allow tools to choose what metadata they require and how they present it
+- Support both public (unencrypted) metadata and metadata accessible only to the Celo Foundation
+
+## Proposed Solution
+
+### Claims
+
+The Attestations contract is extended as follows:
```
nambrot: Actually, what do you think about hosting claims entirely off-chain? Thinking more about it, there seems to be no reason to really put this in a smart contract other than being a convenient spot, and the storage costs do not seem worth it?

timmoreton: I think it could support future on-chain verification of these claims. Maybe phone numbers are just a special case where we've been able to provide an on-chain verification.

nambrot: I think the attestation of those claims can be done with the existing attestations struct, and the claims being hosted off-chain shouldn't impact this?

nambrot: Phone numbers are special because they are not easily verifiable in a "permissionless" way which is why we delegate that task to the protocol. Twitter/DNS can be verified by anyone easily.

CIPs/0003.md
```diff
@@ -0,0 +1,134 @@
...
+
+TODO confirm encryption method
+
+### Later: claims for other services
+
+For claims with `service` set to `github`, `identifier` is the Github handle, and `data` identifies the raw contents of a gist, such that `https://gist.githubusercontent.com/<identifier>/<data>` provides a file in the given format. Claim verifiers should take care to disallow relative path segments (e.g `/../`) in `data`. 
+
+For claims with `service` set to `twitter`, `data` is the identifer of a tweet (such that it could be retrieved at `https://api.twitter.com/1.1/statuses/show.json?id=<id>`) whose body has the same format as above.
+
+TODO - twitter requires an API key to retrieve tweets in json format. This is a centralization burden on claim verifiers.
```
nambrot: I think in the worst case, we don't even need to use the API? We could just pull the tweet from the HTML page and grep for the signature?

CIPs/0003.md
```diff
@@ -0,0 +1,134 @@
+# CIP [0003]: Identity Claims and Metadata
+
+- Date: 2019-09-19
+- Author: @timmoreton
+- Status: WIP
+
+## Overview
+
+This CIP describes a general mechanism that allows principals on the Celo platform to claim real-world identities and maintain rich, unverified metadata. Metadata may be public, or optionally, encrypted and accessible only to the Celo Foundation. Third parties can verify these identity claims, and use the associated unverified metadata as appropriate.
```
cla-bel: I would caution about putting responsibilities on the Celo foundation:
1/ not yet clear what the roles and responsibilities of the foundation will be, recommendations received from other projects is to limit the scope to resource allocation only (no process or day-to-day management, which should be outsourced)
2/ I'd be in favor of planning for a day where the foundation doesn't even need to exist anymore (a la bitcoin)

CIPs/0003.md
pensebien: This can be nice.

CIPs/0003.md
yasoo655: yasoo65556777@gmail.com

PR #6, opened 2020-02-11
CIP-0005 

kevjue:


CIPs/0005.md
```diff
@@ -0,0 +1,68 @@
...
+## Goals
+
+- Reduce bandwidth requirements of the full nodes.
+- Maximize availability of the IBFT overlay network.
+
+## Proposed Solution
+
+The announce protocol has to be able to deal with two fundamental faulty scenarios.  For a given validator pair, when only one of the validators has the correct enodeURL of the other (the asymmetric ignorance case) and when both of the validators don't have the correct enodeURL of the other (the mutual ignorance case).
+
+To deal with these cases, this proposal proposes adding the following high level components to the existing Announce protocol.  1) Add the capability for a validator to directly communicate to another remote validator it's assigned enodeURL, 2) add fast discovery of validator enode mapping changes, and 3) add capability for a validator to query another validator for it's assigned enodeURL via a gossiped message.  Details of these components are in the subsections below.
```
m-chrzan: Typo: it's -> its.

CIPs/0005.md
```diff
@@ -0,0 +1,68 @@
...
+
+The asymmetric ignorance case can then be handled by having the validators directly communicate to each remote validator it's assigned enodeURL whenever the enodeURL mapping is updated.  Note that this will need to happen for all remote validators, even if the assigned URL is not updated for a subset of the validators, since all validators will need to know what is the updated version number.  Also, this direct communication will need to be handled when there is and is not an existing connection between validators.
+
+The mutual ignorance case can be handled by having each validator monitor when a validator enode mapping changes, and when it realizes that it's changed, it will gossip a query message to the validator for the updated assigned enodeURL.  That validator will then directly communicate it to the querying validator.
+
+### Direct communication of assigned enodeURL
+
+This component will include the following changes:
+
+1)  Announce message format will need to be changed so that only one authenticated and encrypted enodeURL (along with the version number) can be put in the message payload
```
m-chrzan: Rather than describing the difference from the current implementation, I think this CIP would be clearer with explicit descriptions of the intended protocols.

CIPs/0005.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Fast discovery of changed enodeURL mapping
+
+This component will require all full nodes and validators to store a copy of a table that will contain the latest enodeURL mapping version number for each elected validator.  Whenver a validator updates it's enodeURL mapping, it will need to gossip a message containing the updated version number, and all nodes should update their table.
+
+Whenever a connection is established between two nodes, they will synchronize their tables so that both tables will have the latest version numbers.  This will handle the case of when a node just joins the P2P network and when a network partition has been repaired.  In those cases, the nodes throughout the network should quickly synchronize their tables.
+
+### EnodeURL query message
+
+When a validator realizes that it's assigned enodeURL for a remote validator has been updated, that validator should gossip a enodeURL query message destined to the remote validator.  Once the destined validator receives the message, it will send a direct announce message to the originating validator.
```
m-chrzan: Typos: it's -> its, a -> an.

CIPs/0005.md
nambrot: Agreed

CIPs/0005.md
```diff
@@ -0,0 +1,74 @@
...
+
+The mutual ignorance case can be handled by having each validator monitor when a validator enode mapping changes, and when it realizes that it's changed, it will gossip a query message to the validator for the updated assigned enodeURL.  That validator will then directly communicate it to the querying validator.
+
+### Direct communication of assigned enodeURL
+
+This component will include the following changes:
+
+1)  Announce message format will need to be changed so that only one authenticated and encrypted enodeURL (along with the version number) can be put in the message payload
+2)  On enodeURL mapping update, that validator will need to directly send an announce message to all validators. In the case of when the validator already has a connection with the remote validator, the originating validator should simply send an announce message to it.  In the case when that connection doesn't exist, then the announce message will need to be sent as part of the connection establishing handshake.
```
nambrot: What does enodeURL mapping update mean?

nambrot: What is the benefit of directly sending the announce message? Purely speed? There are no actual bandwidth savings, right?

nambrot: Will a direct announce receiver gossip the announce further?

CIPs/0005.md
```diff
@@ -0,0 +1,74 @@
...
+
+Issue https://github.com/celo-org/celo-monorepo/issues/2667 tracks this component.
+
+### Fast discovery of changed enodeURL mapping
+
+This component will require all full nodes and validators to store a copy of a table that will contain the latest enodeURL mapping version number for each elected validator.  Whenver a validator updates it's enodeURL mapping, it will need to gossip a message containing the updated version number, and all nodes should update their table.
+
+Whenever a connection is established between two nodes, they will synchronize their tables so that both tables will have the latest version numbers.  This will handle the case of when a node just joins the P2P network and when a network partition has been repaired.  In those cases, the nodes throughout the network should quickly synchronize their tables.
```
nambrot: How does this differ from the current implementation?

CIPs/0005.md
```diff
@@ -0,0 +1,74 @@
...
+
+Issue https://github.com/celo-org/celo-monorepo/issues/2668 tracks this component.
+
+### EnodeURL query message
+
+When a validator realizes that it's assigned enodeURL for a remote validator has been updated, that validator should gossip a enodeURL query message destined to the remote validator.  Once the destined validator receives the message, it will send a direct announce message to the originating validator.
```
nambrot: How does the origin validator realize that?

nambrot: What's the benefit of this component?

CIPs/0005.md
kevjue: > What does enodeURL mapping update mean?
Each validator will maintain a mapping of remote validator -> enodeURL that they should use to connect to the originating validator.  Whenever that mapping is updated, then the originating validator will gossip out an updated version number.






> What is the benefit of directly sending the announce message? Purely speed? There are no actual bandwidth savings, right?

There actually will be a ton of bandwidth savings for full nodes, since the announce message will not be gossiped throughout the network.



> Will a direct announce receiver gossip the announce further?

No, the receiver will not.

CIPs/0005.md
kevjue: The difference is that in the current implementation, it will store all of the assigned enodeURLs for all of the elected validators.  So basically will store a O(N<sup>2</sup>) table.  Also, each row of the table is wide (e.g. has many bytes).  And more importantly, since this table is gossiped, bandwidth of receiving and sending this full table is very high (table is received up to C times), where C is the number of connections of the full node.

With this change, each full node will store a table that is O(N) size.  And each row is relatively small.  They only contain an address, version number (64 bit), and signature.  So smaller than a transaction.

CIPs/0005.md
kevjue: > How does the origin validator realize that?

Via the gossiped version number msg described in the above subsection.

> What's the benefit of this component?

This component is actually necessary.  Once a validator realizes that it's assigned enodeURL from a remote validator has potentially changed.  Then the origin validator will need to gossip this message.

CIPs/0005.md
```diff
@@ -0,0 +1,74 @@
...
+
+The asymmetric ignorance case can then be handled by having the validators directly communicate to each remote validator it's assigned enodeURL whenever the enodeURL mapping is updated.  Note that this will need to happen for all remote validators, even if the assigned URL is not updated for a subset of the validators, since all validators will need to know what is the updated version number.  Also, this direct communication will need to be handled when there is and is not an existing connection between validators.
+
+The mutual ignorance case can be handled by having each validator monitor when a validator enode mapping changes, and when it realizes that it's changed, it will gossip a query message to the validator for the updated assigned enodeURL.  That validator will then directly communicate it to the querying validator.
```
nategraf: I don't quite understand the flow here. How does "each validator monitor when a validator enode mapping changes"? Isn't the who point of the announce scheme to push a message to let everyone know it has changed?

CIPs/0005.md
```diff
@@ -0,0 +1,74 @@
...
+
+There were two variants to this proposals that were considered.
+
+1) Not including the fast discovery of changed enodeURL mapping, and instead have validators send enodeURL query messages once they realize that they are not connected to an elected validator.  The proposal will not handle the case when a validator is down as elegantly.  In that case, all the other validators will continuously be gossiping query messages until it's back up, which could result in high bandwidth usage.
+
+2) Persisting the enodeURL query gossip message for all full nodes/validators and have them be regossiped on peer connection.  This will reduce any need to regossip query messages and I think can be considered to be added later.  This does add significant complexity to the proposal above and probably handles a case that rarely happens.  If we find out that it happens more then expected, then this feature could be added.
+
+## Risks
+
+These protocol changes has a direct impact on consensus since the IBFT overlay network depends on this protocol.
```
nategraf: I think we should be more clear here that this proposal is not interoperable with the existing announce protocol and this change could not be deployed to a live network as proposed.

CIPs/0005.md
```diff
@@ -0,0 +1,74 @@
...
+This component will include the following changes:
+
+1)  Announce message format will need to be changed so that only one authenticated and encrypted enodeURL (along with the version number) can be put in the message payload
+2)  On enodeURL mapping update, that validator will need to directly send an announce message to all validators. In the case of when the validator already has a connection with the remote validator, the originating validator should simply send an announce message to it.  In the case when that connection doesn't exist, then the announce message will need to be sent as part of the connection establishing handshake.
+
+Issue https://github.com/celo-org/celo-monorepo/issues/2667 tracks this component.
+
+### Fast discovery of changed enodeURL mapping
+
+This component will require all full nodes and validators to store a copy of a table that will contain the latest enodeURL mapping version number for each elected validator.  Whenver a validator updates it's enodeURL mapping, it will need to gossip a message containing the updated version number, and all nodes should update their table.
```
asaj: Will non-validators be able to "spoof" these messages and trick full nodes into believing the version number is higher than it is?

CIPs/0005.md
asaj: Are the version numbers signed by the validators? Otherwise full nodes could trick other full nodes into accepting larger version numbers

CIPs/0005.md
```diff
@@ -0,0 +1,74 @@
...
+- Status: WIP
+
+## Overview
+
+This proposal contains changes to Celo's announce protocol.  That protocol's main goal is to distribute each validator's enodeURL(s) to all the other validators, so that they can create direct TCP connections among each other (effectively create an overlay network among the validators).  This overlay network is then used by Celo's IBFT consensus algorithm to reach consensus on block proposals.
+
+The validator enodeURLs do need to be encrypted when transmitted over the wire, as we want to minimize the number of entities that are aware of it to reduce the change of a DOS attack on the network's validators.  To that end, the enodeURLs needs to be encrypted with each of the remote validators' public key when sent over the wire.  Consequently, each validator will need to share data that is O(N<sup>2</sup>) size where N is the number of elected validators (Note that the validators will also share it's enodeURL(s) with the top 10 validators that are just below the election threshold cutoff.  Throughout this document, the phrase "elected validators" will include those 10).
```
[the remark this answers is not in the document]
kevjue: > How do we account for a key rotation? We'll want to make sure that if a new validator signer is authorized midway through an epoch, for the rest of the epoch we include the current validator signer & the new validator signer in this set

Good point.  The most straightforward solution is to double the size of the val_enode_table.  Will need to think if there is a more effective solution.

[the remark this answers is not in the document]
kevjue: > Are connections always maintained amongst the entire set (even those not currently elected)?

Ya, My thinking is since we're reducing the number of nodes that will announce (e.g. it's elected + 10), mind as well make the persisted connection logic easier.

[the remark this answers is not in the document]
kevjue: > Iirc running an election is pretty expensive of an operation, how often do you expect this set to be calculated? Cached similar to what is done with the active and registered validators?

Yes.  It'll be cached.

CIPs/0005.md
kevjue: You're right that it doesn't need to be encrypted.  My thought is that it'll be easier to code by having one format for the direct announce and gossiped announce.

CIPs/0005.md
kevjue: This message will contain a signature from the originating validator, so wouldn't be able to be spoofed.

CIPs/0005.md
kevjue: Yes, these messages will be signed.

CIPs/0005.md
kevjue: > Rather than describing the difference from the current implementation, I think this CIP would be clearer with explicit descriptions of the intended protocols.

As the messages are added within Geth, I'll update this CIP to include those message formats.

PR #7, opened 2020-05-03
CIP [0006] Proposal to allow voting directly through releasegold contract

zviadm:


tromer:
Don't you need to also proxy `election:activate`, `election:revoke` and `governance:revokeupvote`?

Also, would it make sense to use polymorphism and abstract interfaces (say `IGoldLocker`, `IGovernanceVoter`, `IElectionVoter`) that would be implemented by ReleaseGold, but also any other future contact that implements similar proxying? Then UIs and other contacts can use the abstract interfaces, and thus automatically support any future such contract.

zviadm:
> Don't you need to also proxy `election:activate`, `election:revoke` and `governance:revokeupvote`?
> 
> Also, would it make sense to use polymorphism and abstract interfaces (say `IGoldLocker`, `IGovernanceVoter`, `IElectionVoter`) that would be implemented by ReleaseGold, but also any other future contact that implements similar proxying? Then UIs and other contacts can use the abstract interfaces, and thus automatically support any future such contract.

You are right `electionActivate` and `governanceRevokeUpVote` will also be needed. Election revoke functions already exist in the release gold contract.

nambrot:
@zviadm Thank you for submitting this CIP.

As you and @tromer already found out, the surface area of what voting functions would have to be exposed on `ReleaseGold` is rather large. I believe the original intention behind leveraging the authorization of voter keys in `Accounts` was to reduce said surface areas.

I would definitely be interested in hearing what risks/downsides you are seeing in the current version of `ReleaseGold`. If you would like to have both the beneficiary and the voter key on the same hardware wallet to achieve your desired goal, then that currently seems possible? (albeit not recommended as there should be different access patterns to both keys)

zviadm:
> @zviadm Thank you for submitting this CIP.
> 
> As you and @tromer already found out, the surface area of what voting functions would have to be exposed on `ReleaseGold` is rather large. I believe the original intention behind leveraging the authorization of voter keys in `Accounts` was to reduce said surface areas.
> 
> I would definitely be interested in hearing what risks/downsides you are seeing in the current version of `ReleaseGold`. If you would like to have both the beneficiary and the voter key on the same hardware wallet to achieve your desired goal, then that currently seems possible? (albeit not recommended as there should be different access patterns to both keys)

Think of website like `celoist.com`, and amount of extra effort they will have to go through to implement voting through web UI for release gold contracts. 

For regular addresses they can just connect to the ledger wallet and send vote transactions through `forno` node. For release gold contracts though, they will need to implement concept of authorizing a vote signer first, explain that concept to the user, explain that they need to put that vote signer also on ledger and then send votes through that authorized signer. 

This will significantly increase complexity of voting flow. Considering everyone in initial setup will be operating with release gold contracts, i think it is going to create additional complexity barrier for people to actually lock gold and vote properly. Especially voting on governance proposals. 

There is another use case too. It is very useful to have this type of setup:
-> using vote signer address for automatic voting for groups (through a script for example)
-> using address on ledger to vote manually for governance proposals. 
This type of setup is no longer possible with releasegold contract because you can only have one authorized voter, and if it is on ledger it cant be used in automation, and if it is not on the ledger it can't be used through nice easy to use web UI.

asaj:
> As you and @tromer already found out, the surface area of what voting functions would have to be exposed on ReleaseGold is rather large. I believe the original intention behind leveraging the authorization of voter keys in Accounts was to reduce said surface areas.

+1, using the vote signer keeps the contracts forward compatible with any changes to governance and proof of stake. Otherwise, to support new functionality, every RG contract would need to be upgraded with every change to the protocol, which isn't feasible.

@zviadm I'm wondering if there are other ways to reduce this complexity? For folks who are only looking to vote, they can authorize their beneficiary as the vote signer, which means they don't have to manage extra keys. 

> There is another use case too. It is very useful to have this type of setup:
> -> using vote signer address for automatic voting for groups (through a script for example)
> -> using address on ledger to vote manually for governance proposals.
> This type of setup is no longer possible with releasegold contract because you can only have one authorized voter, and if it is on ledger it cant be used in automation, and if it is not on the ledger it can't be used through nice easy to use web UI.

I don't fully understand why in this scenario they would want to vote with a ledger for governance proposals, why not just use the vote signer that they're using for automated voting?

I would imagine that folks who are setting up automatic voting scripts can probably deal with the added complexity of managing a separate vote signer.

CIPs/0006.md
```diff
@@ -0,0 +1,48 @@
+# CIP [0006]: Implement ElectionVote, GovernanceUpvote and GovernanceVote functions in releasegold contract
+
+- Date: 2020-05-03
+- Author: @zviadm
+- Status: DRAFT
```
YazzyYaz, line 5: Hi @zviadm thank you so much for submitting a Celo CIP!

Can you update your draft to follow the new CIP Template here: https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

Once the missing items are added and it follows the new template format, I'll merge the draft. Let me know if you'd like me to help you out on guidance on the new template.

YazzyYaz (changes requested):

CIPs/0006.md
zviadm: This CIP is no longer relevant, since most of the ReleaseGold owners have figured out how to manage their locking and voting. And there isn't huge amounts of new ReleaseGold contracts issued anymore. 
Thus we can just close this CIP

PR #9, opened 2020-05-06
[CIP 0007] WIP Downtime Slasher

gastonponti:
CIP 0007

Affordable Downtime Slasher

### DIFF
--- a/CIPs/0007.md
+++ b/CIPs/0007.md
@@ -1,0 +1,108 @@
+# CIP [0007]: Affordable Gas Downtime Slasher
+
+- Date: 2020-05-05
+- Author: @gastonponti (with a lot of feedback and ideas from @m-chrzan)
+- Status: WIP
+
+## Overview
+
+This proposal contains a new DowntimeSlasher contract.
+
+The actual DowntimeSlasher tries to verify that the validator signed at least a block in all the window that is required to be validated which it leads to running out of gas for big windows.
+
+The main goal of this new contract is to generate a tool that would allow any user to slash a validator for every possible block windows without gas limitations. 
+
+For the actual window of 1/2 epoch, the DowntimeSlasher won't finish with a "run out gas" exception. Without this new tool for slashing, there is no a possible way to penalize a validator for downtime.
+
+## Goals
+
+Make a new contract that:
+- Allows the user to slash a validator if it was down for a specific consecutive block window
+- The implementation works for every possible block window (required in the initialization)
+
+## Proposed Solution
+
+Create a new contract funtion that divides the downtime verification in valid slots.
+
+The user that wants to slash a particular validator will have to call a function from the contract `S_n` times (`S_n` could be asked to the contract itself) to verify each slot.
+
+Then the user must generate another contract call to verify that the validator to be slashed was down for every consecutive slot.
+
+### Process
+
+1) The user calls the contract to know how many slots are required (`S_n`) and how many blocks must have every slot (`B_n`)
+
+2) The user requires to validate `S_n` slots
+    - The user calls the contract for the slot validation `[StartBlockNumber + (i * B_n), StartBlockNumber + ((1 + i) * B_n) )` with `0 <= i < S_n`
+
+3) The user calls the slash method with the `StartBlockNumber` (internally will verify that every slot validation was performed and succeded)
+
+### Variation of the proposed solution
+
+- Every user generates the validation slots for itself. The first user to gather all the validations and trigger the slashing, collects the reward
+
+- The slots validated works for every user in the community. When a slashing is performed, every user that generated a validated slot that is used to demostrate the downtime, receives a portion of the reward 
+
+## Alternative Solutions
+
+### Validator's Uptime in the epoch
+
+The Celo blockchain for reward purposes, calculates an uptime for each validator. The validator’s uptime for the epoch is the proportion of blocks in the epoch for which it is ‘up’: U = counter / (epoch_size - 11).
+
+One possible approach, it's to slash the validators if they are not covering a certain umbral.
+
+#### Pros
+
+The uptime calculation it's part of the Celo protocol. This not only allow us to save gas from calculating a possible slashing, but also could let the same blockchain to perform the slashes. This would make a safer net, instead of requiring a so rigurous control of the users.
+
+In the actual scenario, if no user is controling, no one get slashed.
+
+#### Cons
+
+The validator's uptime is calculated at the end of the epoch, that has two potential problems:
+
+- The slashing for an specific validator will be performed in the next epoch. That allows a validator that is not performing well in the epoch X, to be reelected for the epoch X+1, because the tool used for that (the uptime) was added after the election was completed.
+
+- The uptime is a relation that has meaning tied to an specific epoch as whole. It won't give you a detail information of what happened.
+
+For example, if the community sets the lower umbral as 50% this two scenarios have the same uptime:
+ * A Validator is down for 1/2 epoch consecutive, and then signs every block for the next 1/2 epoch
+ * A Validator signs 1 every 2 packages
+
+Another problem is the calculation between epoch for reelected validators.
+Using the same umbral of the other example, if the validator has in the epoch X = 75% uptime, and in the epoch X+1 = 70% uptime, this two numbers could hide a more that a 1/2 epoch consecutive blocks downtime. This will happen if the 25% of the downtime in the X is at the ending of the epoch, and the other 30% of the downtime for the epoch X+1 is at the beginning. 
+
+### Challenge the validator
+
+This solutions consist in solving a challenge between a user and the validator. 
+
+The user could claim that the validator has been down for an specific period of blocks. In which case, the validator has an specific amount of time to bring a proove where the blockchain could verify that the validator was up. For example 1 or N blocks that where signed by the validator.
+
+#### Pros
+
+From the blockchain perspective, a simple and easy way for validating the uptime.
+
+#### Cons
+
+- Will require that each validator controls and follows the challenges that it has open against.
+
+- The chain could have a flood of challenges every epoch. To avoid indiscriminated challenges we could require from the user to place a bid, that the validator could win if it prooves its inocence. This way we could avoid non-checked downtime challenges, and the validator could have a good incentive to adopt this alternative solution.
+
+### Why were discarded the other solutions?
+
+The objective of this CIP is to fix a tool that today doesn't work properly. Changing the way to calculate in which situations could a validator get slashed, requires a deeper discussion that don't have to be rushed.
+
+If we change the rules, with a tool broken, we are basically forcing a result, and it's not the purporse of this CIP.
+The author highly recommend to take in count these other ways of slashing for the future, without the rush that need brings.
+
+These other tools could be complementary to the actual (and working properly) one. Cover the different aspects of what we believe is a validator working as it should, it's the key to setting the standards of excellence that we seek.
+
+
+## Useful Links
+
+* [Epoch uptime information](https://docs.celo.org/celo-codebase/protocol/proof-of-stake/epoch-rewards/validator-rewards)
+* Links to related CIPs or other documents
+
+## Implementation
+
+* To follow.


CIPs/0007.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Date: 2020-05-05
+- Author: @gastonponti (with a lot of feedback and ideas from @m-chrzan)
+- Status: WIP
+
+## Overview
+
+This proposal contains a new DowntimeSlasher contract.
+
+The actual DowntimeSlasher tries to verify that the validator signed at least a block in all the window that is required to be validated which it leads to running out of gas for big windows.
```
nategraf, line 11: We would like any readers unfamiliar with the issue that prompted this proposal, and who may be external to cLabs, to have the context to understand and provide comments on the proposal.
```suggestion
Currently, slashable downtime is defined as a window of blocks, greater than a predefined length, in which no block's parent aggregated seal includes a signature from the validator in question. It's currently implementation checks and executes slashing in a single transaction which requires an amount of gas linear in the required window size.

Under its current implementation, the gas required to evaluate a downtime event may far exceed the block gas limit. In the Release Candidate 1 network, the downtime slashing window is configured to be 5760 blocks, which would require roughly 13.5 million gas to evaluate, while the block gas limit is set to 10 million.
```

CIPs/0007.md
```diff
@@ -0,0 +1,108 @@
...
+
+The main goal of this new contract is to generate a tool that would allow any user to slash a validator for every possible block windows without gas limitations. 
+
+For the actual window of 1/2 epoch, the DowntimeSlasher won't finish with a "run out gas" exception. Without this new tool for slashing, there is no a possible way to penalize a validator for downtime.
+
```
nategraf, line 16:
```suggestion
```
See above comments.

CIPs/0007.md
```diff
@@ -0,0 +1,108 @@
...
+## Goals
+
+Make a new contract that:
+- Allows the user to slash a validator if it was down for a specific consecutive block window
+- The implementation works for every possible block window (required in the initialization)
```
nategraf, line 21: What does this mean? Could you rephrase this?

CIPs/0007.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Every user generates the validation slots for itself. The first user to gather all the validations and trigger the slashing, collects the reward
+
+- The slots validated works for every user in the community. When a slashing is performed, every user that generated a validated slot that is used to demostrate the downtime, receives a portion of the reward 
+
+## Alternative Solutions
+
+### Validator's Uptime in the epoch
+
+The Celo blockchain for reward purposes, calculates an uptime for each validator. The validator’s uptime for the epoch is the proportion of blocks in the epoch for which it is ‘up’: U = counter / (epoch_size - 11).
```
nategraf, line 50: The uptime number also uses the 12-block sliding window, so this description is not quite accurate.

CIPs/0007.md
```diff
@@ -0,0 +1,108 @@
...
+
+One possible approach, it's to slash the validators if they are not covering a certain umbral.
```
nategraf, line 52:
```suggestion
One possible approach, is to slash the validators if they do not meet a certain threshold for uptime in an epoch.
```

CIPs/0007.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Cons
+
+The validator's uptime is calculated at the end of the epoch, that has two potential problems:
+
+- The slashing for an specific validator will be performed in the next epoch. That allows a validator that is not performing well in the epoch X, to be reelected for the epoch X+1, because the tool used for that (the uptime) was added after the election was completed.
+
+- The uptime is a relation that has meaning tied to an specific epoch as whole. It won't give you a detail information of what happened.
+
+For example, if the community sets the lower umbral as 50% this two scenarios have the same uptime:
```
nategraf, line 68: English work is "threshold" :)
```suggestion
For example, if the community sets the lower threshold as 50% this two scenarios have the same uptime:
```

CIPs/0007.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Challenge the validator
+
+This solutions consist in solving a challenge between a user and the validator. 
+
+The user could claim that the validator has been down for an specific period of blocks. In which case, the validator has an specific amount of time to bring a proove where the blockchain could verify that the validator was up. For example 1 or N blocks that where signed by the validator.
+
+#### Pros
+
+From the blockchain perspective, a simple and easy way for validating the uptime.
```
nategraf, line 83: Implementing a challenge-response scheme would be far from simple in practice. I don't think we have a fleshed out enough idea of what this would look like to really include it as a proposal in this CIP.

CIPs/0007.md
```diff
@@ -0,0 +1,108 @@
...
+### Process
+
+1) The user calls the contract to know how many slots are required (`S_n`) and how many blocks must have every slot (`B_n`)
+
+2) The user requires to validate `S_n` slots
+    - The user calls the contract for the slot validation `[StartBlockNumber + (i * B_n), StartBlockNumber + ((1 + i) * B_n) )` with `0 <= i < S_n`
+
+3) The user calls the slash method with the `StartBlockNumber` (internally will verify that every slot validation was performed and succeded)
+
+### Variation of the proposed solution
```
asaj, line 40: Which of these are you proposing?

gastonponti: Well, I'm not totally decided, and I want to discuss it a little.

For my point of view, the "user has to verify everything" it's the approach that would be easier for the community to accept because it's not changing anything (same as the old one).
That's why I was thinking to use that. But I would like some opinions on that.

(And you are totally right, I should write which option I'm proposing clearer)

CIPs/0007.md
```diff
@@ -0,0 +1,108 @@
+# CIP [0007]: Affordable Gas Downtime Slasher
+
+- Date: 2020-05-05
+- Author: @gastonponti (with a lot of feedback and ideas from @m-chrzan)
+- Status: WIP
+
+## Overview
+
+This proposal contains a new DowntimeSlasher contract.
```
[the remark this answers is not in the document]
gastonponti: Totally agree

CIPs/0007.md
kevjue: I like the 1st proposal.  It's much more intuitive (which will incidentally make implementation much more easier) and I would think that the added functionality of the 2nd proposal will hardly be utilized.  E.g. it will almost always be one one user that generates all of the validation slots.  I suppose that we could also upgrade the smart contract later on down the road if there is alot of community demand for the 2nd proposal.

CIPs/0007.md
```diff
@@ -0,0 +1,108 @@
...
+
+One possible approach, it's to slash the validators if they are not covering a certain umbral.
+
+#### Pros
+
+The uptime calculation it's part of the Celo protocol. This not only allow us to save gas from calculating a possible slashing, but also could let the same blockchain to perform the slashes. This would make a safer net, instead of requiring a so rigurous control of the users.
+
+In the actual scenario, if no user is controling, no one get slashed.
+
+#### Cons
```
kevjue, line 60: Seems like one way to address most (if not all the cons) is that every validator will check for uptime at *every* block, and not just at the end of the epoch.

mrsmkl:
Because we are now making changes to slashing, I think we should now consider whether the penalties are appropriate:
* Penalty for validator: If it is 100 cGLD, most of the penalty will probably be from score dropping to zero
* Penalty for groups: Because the slashing multiplier is applied to voter rewards, I think this means that the group is basically dead for the month until multiplier is reset.

Also randomized sampling should be considered. Unless it has some vulnerability, I think it should be used since it requires least changes to the current code.

nategraf:
After a seconds inspection, I agree that the randomized sampling approach is worth considering. 

In the case of algorithm that checks `n` blocks of a slashable window of size `w`, of which the validator in question signed `s` blocks (0 if the validator was down the whole time). The following gives the probability that a validator will be slashed in a given slashing attempt.
```
1 - (1 - ((w-s) / w) ^ n)
```

> Note: If `s` is `0`, the validator will be slashed with probability `1`

As an example, with `n = 4000` and `s / w = 0.5` (i.e. 5% uptime) the slashing probability in a given attempt is `7.84 * 10 ^ -90`, which is much less than the probability of a successfully discovering a Keccak256 pre-image with a single random guess, with we already assume is impossible.

Here is a psuedo-solidity algorithm for such a sampling-based approach
```solidity
function isValidatorSlashable(address validator, uint256 startBlock, uint256 w, uint256 n) {
  // Seed selection discussed below.
  uint256 r = seed();
  bool slashable = true;
  for (uint256 i = 0; i < n; i++) {
    r = uint256(keccak256(r))
    // Index practically uniform in [0, w), with a bias of less than w * (2 ^ -256) << 2^-192
    uint256 index = r % w;
    slashable &= isValidatorDownAt(validator, startBlock + index);
  }
  return slashable;
}
```

Such an algorithm would have an overhead of a `SHA3`, an `MSTORE` and a `MOD` operation, which is collectively less than `50` gas per loop iteration, over the current approach. Currently, we can fit about 5000 loops iterations into the `10M` block gas limit and each iteration costs a minimum of `800` gas for the precompile call. As a rough estimate this would mean at least `4700` samples could be taken for the same gas cost. The `4000` estimate quoted above was taken from this calculation.

Seed selection could be based on the core Random contract, but does not need to be and could be based on the block number alone or even a fixed string. Using the Randomness contract would ensure that the slashing samples could not be precomputed and thus adds to the difficulty of slashing validators that are not 100% down during a given window, although the difference is slight and is not overall important to security.

Another variation on this would keep a bitmap of length `w` in memory of blocks that have already been checked, and resample when a repetition is detected. This would further reduce the probability of a validator being slashable to
```
1 - (1 - ((w-s)! / (w^n * (w - s - n)!)))
```

Gas cost for keeping this array, which would be no more than `ceiling(17280 / 256) = 68` words is small, so this would not be gas cost prohibitive. It would however make the gas cost of the function execution probabilistic, and therefore uncomputable within the gas limit with certain seeds.

gastonponti:
Updated with the actual implementation.
Reduced a lot the text to focus, and go directly to the point.
@asaj @nategraf

CIPs/0007.md
```diff
@@ -0,0 +1,65 @@
+# CIP [0007]: Affordable Gas Downtime Slasher
+
+- Date: 2020-05-05
+- Author: @gastonponti (with a lot of feedback and ideas from @m-chrzan)
+- Status: Implemented
```
YazzyYaz, line 5: This needs to be changed to Draft status

CIPs/0007.md
```diff
@@ -0,0 +1,65 @@
...
+
+Under its current implementation, the gas required to evaluate a downtime event may far exceed the block gas limit. In the Release Candidate 1 network, the downtime slashing window is configured to be 5760 blocks, which would require roughly 13.5 million gas to evaluate, while the block gas limit is set to 10 million.
+
+## Goal
+
+- Allow the user to slash a validator if it was down for a specific consecutive block window without running out of gas.
+
+## Proposed Solution
+
+Create a new contract funtion that divides the downtime verification in validation intervals.
```
YazzyYaz, line 23: function* typo

CIPs/0007.md
```diff
@@ -0,0 +1,65 @@
...
+
+The user that wants to slash a particular validator will have to call a function from the contract that will generate a proof with the validators that signed at least one block for the specific interval set by the caller.
+
+The user will require to repeat this first process the amount of times necessary to cover with overlapping or adjacent intervals, at least the `slashableDowntime` amount of blocks set in the contract.
+
+Then the user must generate another contract call listing all the intervals required to verify that the validator to be slashed was down for those intervals, and that also covers the `slashableDowntime` window.
+
+### Note
+
+This process allows the user to validate a interval that will contain the data of every validator. This means that if for example, three validators where down in the same period, the same interval could be used to slash the three of them.
```
YazzyYaz, line 33: an* interval

CIPs/0007.md
```diff
@@ -0,0 +1,65 @@
...
+### Some numbers
+
+Tested with intervals of 5000 blocks (each one required less than 10 million gas), using 1000 intervals for the actual slash (required ~6.5 million gas) which covers 5M blocks.
+Even if is not necessary right now, this solution will cover possible governable incrementations of the `slashableDowntime` window.
+
+### Variation of the proposed solution
+
+- The slots validated could work for every user in the community. When a slashing is performed, every user that generated a validated slot that is used to demostrate the downtime, receives a portion of the reward. But to address this, changes are required in other contracts to calculate and distribute the rewards.
+
+## Alternative Solution
```
YazzyYaz, line 45: I would remove this entire section. Alternative solutions should be proposed under their own respective CIPs

YazzyYaz (changes requested):

CIPs/0007.md
nategraf: Just to make sure we are on the same page, this is implemented and being deployed with contract release 1. Is the status flipped once the contract is finally deployed?

CIPs/0007.md
```diff
@@ -0,0 +1,65 @@
...
+- Author: @gastonponti (with a lot of feedback and ideas from @m-chrzan)
+- Status: Implemented
+
+## Overview
+
+This proposal includes improvements to Celo's downtime slashing, such that it can be evaluated within the block gas limit.
+
+In order to promote a stable network, the Celo protocol includes slashing for downtime such that if a validator fails to participate in consensus for a given amount of time, they may be slashed. More information about slashing in the Celo protocol can be found on [docs.celo.org](https://docs.celo.org/celo-codebase/protocol/proof-of-stake/penalties)
+
+Currently, slashable downtime is defined as a window of blocks, greater than a predefined length, in which no block's parent aggregated seal includes a signature from the validator in question. It's currently implementation checks and executes slashing in a single transaction which requires an amount of gas linear in the required window size.
```
nategraf, line 13: The definition did not change. Removing the work "currently" would help clarify that.
```suggestion
Slashable downtime is defined as a window of blocks, greater than a predefined length, in which no block's parent aggregated seal includes a signature from the validator in question. It's currently implementation checks and executes slashing in a single transaction which requires an amount of gas linear in the required window size.
```

CIPs/0007.md
```diff
@@ -0,0 +1,65 @@
...
+The user that wants to slash a particular validator will have to call a function from the contract that will generate a proof with the validators that signed at least one block for the specific interval set by the caller.
+
+The user will require to repeat this first process the amount of times necessary to cover with overlapping or adjacent intervals, at least the `slashableDowntime` amount of blocks set in the contract.
+
+Then the user must generate another contract call listing all the intervals required to verify that the validator to be slashed was down for those intervals, and that also covers the `slashableDowntime` window.
+
+### Note
+
+This process allows the user to validate a interval that will contain the data of every validator. This means that if for example, three validators where down in the same period, the same interval could be used to slash the three of them.
+The gas required to generate the `I_n` intervals will be spent only in the first validation.
```
nategraf, line 34:
```suggestion
> Note: This process allows the user to validate a interval that will contain the data of every validator. This means that if for example, three validators where down in the same period, the same interval could be used to slash the three of them.
> The gas required to generate the `I_n` intervals will be spent only in the first validation.
```

nategraf (approved):
Overall, looks good to me.

CIPs/0007.md
YazzyYaz: I now see that Alternative Solutions was specified in the current template, so no worries, please keep this section :)

CIPs/0007.md
gastonponti: Exactly

YazzyYaz (changes requested):
Hi @gastonponti thanks for writing this, currently with the activation of the new CIP process, can you ensure that the CIP specification matches the new template here: https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

PR #10, opened 2020-05-07
Set status of CIP3

nambrot:
Should be implemented now


Istanbulexpat:
Why am i getting every notification on this - and yet it says I'm ignoring and not listening - and unsubscribe link in not valid for my account?

SovereignXpid:
Can someone tell me how to unsubscribe – it’s not working.

PR #11, opened 2020-05-08
CIP8: Off-chain storage

nambrot:
In CIP3, the Celo protocol has started with a simple off-chain storage mechanism mainly focused on claims an account can make about their off-chain identity and communicate to other participants in the network. Over time, there have been additional use cases which have similar access patterns that prompted the creation of `Metadata` in the first place. I.e. data that does not require the expensive cost of "hard consensus", instead just needs to be authenticated by a single party.

Examples include:
- Transaction metadata such as comments or encrypted messaging in general
- Profile information such as name or a profile picture
- Unidirectional attestations like an audit or badges won in The Great Celo Stake Off
- Signed statements/claims by a party for relase binaries/docker images/package hashes

This CIP should be ready for initial comments now. It should be worth noting that this CIP only proposes the general storage mechanism. Specific applications should generally specify their own schema on top of this mechanism and considered out of scope for this CIP (except maybe a reference application/schema)

To Add:
- [x] multiple storage roots and namespaces 
- [x] separate out signatures from files
- [x] allow for expiry of files

### DIFF
--- a/CIPs/0008.md
+++ b/CIPs/0008.md
@@ -1,0 +1,73 @@
+# CIP 0008: Expand Metadata to general off-chain storage
+
+- Date: 2020-05-05
+- Author: @nambrot
+- Status: WIP
+
+## Overview
+
+In CIP3, the Celo protocol has started with a simple off-chain storage mechanism mainly focused on claims an account can make about their off-chain identity and communicate to other participants in the network. Over time, there have been additional use cases which have similar access patterns that prompted the creation of `Metadata` in the first place. I.e. data that does not require the expensive cost of "hard consensus", instead just needs to be authenticated by a single party.
+
+Examples include:
+- Transaction metadata such as comments or encrypted messaging in general
+- Profile information such as name or a profile picture
+- Unidirectional attestations like an audit or badges won in The Great Celo Stake Off
+- Signed statements/claims by a party for relase binaries/docker images/package hashes
+
+## Goals
+
+- Expand storage and indexing capabilities of this off-chain data
+- Encode the notion of decryption keys
+- Lightweight messaging system
+
+
+## Proposed Solution
+
+We propose to add another claim, called the `STORAGE` claim:
+```json
+{
+  "type": "STORAGE",
+  "timestamp": 1588886114907,
+  "address": "https://example.com/celo_storage"
+}
+```
+
+The address should be referred to as the root. The expectation here is that data can accessed via `root/${dataAddress}`. For private data, this should be encrypted to the recipient. Here are some common ways of modeling applications into this format:
+
+- For transaction metadata, the account can encrypt the metadata and put it in a file with the transaction hash as the metadata. Further CIPs could specify a standard for structured metadata.
+    - `dataAddress`: `transactionHash`
+    - file contents: Encrypted, structured transaction metadata (comments, or otherwise)
+- Data in relation to other accounts could be communicated to other accounts with:
+    - `dataAddress`: `otherAccountAddress`
+    - file contents: Structured data such as:
+        - `sessionKeys`: The session key(s) encrypted with dataEncryptionKey of `otherAccount` on `Accounts.sol`
+        - `attestations`: These could be signed data attestating things such an audit or badges
+- Profile information could be stored on this system with:
+    - `dataAddress`: `profileName` or `profilePicture`
+    - file contents: the data unencrypted, or encrypted to session keys available to the recipients
+- Releases:
+    - `dataAddress`: `hashOfRelease`
+    - file contents: The signature of the key on the hash of the release
+
+If no assocation between to the `dataAddress` should be leaked, the `dataAddress` itself could be encrypted. The recipient would then just have to decrypt all `dataAddress`s (which could be available in an index file).
+
+
+## Alternative Solutions
+
+I considered doing storing all data on IPFS, however the tooling around IPFS as well as IPNS does not seem user-friendly enough to make a substantial number of users use this directly (or centralize anyways)
+
+## Risks
+
+More reliance on off-chain data and misunderstanding such reliance could lead to inconsistency or poor usability. Ultimately this is traded off against cost of storing data on-chain or in centralized services.
+
+## Useful Links
+
+* [The original Metadata CIP](https://github.com/celo-org/CIPs/blob/master/CIPs/0003.md)
+* Links to related CIPs or other documents
+
+## Implementation
+
+* To follow.
+* Or a link to a PR.
+
+Add any implementation considerations here. For all proposals going through the governance process, this section should reference the code implementing the proposed change. It’s recommended to get community feedback before writing any code.
--- a/CIPs/0008.md
+++ b/CIPs/0008.md
@@ -57,7 +57,7 @@ All file contents should always be signed by an authorized signer of the account
 
 ## Alternative Solutions
 
-I considered doing storing all data on IPFS, however the tooling around IPFS as well as IPNS does not seem user-friendly enough to make a substantial number of users use this directly (or centralize anyways). A generic "HTTP-based file system" is proposed as it is the lowest common denominator. It can be served by a simple HTTP server, even popular VCS static page hosting or specialized service providers.
+I considered doing storing all data on IPFS, however the tooling around IPFS as well as IPNS does not seem user-friendly enough to make a substantial number of users use this directly (or centralize anyways). A generic "HTTP-based file system" is proposed as it is the lowest common denominator. It can be served by a simple HTTP server, even popular VCS static page hosting or specialized service providers. The lock-in effect is minimal since data can be easily migrated, and multiple copies could even be hosted with multiple roots. Thus availabillity concerns relative to on-chain storage can be mitigated relatively well.
 
 ## Risks
 
--- a/CIPs/0008.md
+++ b/CIPs/0008.md
@@ -54,7 +54,7 @@ If no assocation between to the `dataAddress` should be leaked, the `dataAddress
 
 ## Alternative Solutions
 
-I considered doing storing all data on IPFS, however the tooling around IPFS as well as IPNS does not seem user-friendly enough to make a substantial number of users use this directly (or centralize anyways)
+I considered doing storing all data on IPFS, however the tooling around IPFS as well as IPNS does not seem user-friendly enough to make a substantial number of users use this directly (or centralize anyways). A generic "HTTP-based file system" is proposed as it is the lowest common denominator. It can be served by a simple HTTP server, even popular VCS static page hosting or specialized service providers.
 
 ## Risks
 
--- a/CIPs/0008.md
+++ b/CIPs/0008.md
@@ -33,15 +33,15 @@ We propose to add another claim, called the `STORAGE` claim:
 }
 ```
 
-This http address should be referred to as the root. The expectation here is that data can accessed via `root/${dataPath}`. A root can specify a REGEX with the `filterdataPaths` key to indicate what kind of data it stores. An account can make multiple `STORAGE` claims, both to increase availability, as well as effectively sharding this off-chain data. This CIP does not specify what to do in case of conflicts.
+This http address should be referred to as the root. The expectation here is that data can accessed via `root/${dataPath}`. A root can specify a [glob schema](https://github.com/isaacs/minimatch) with the `filterdataPaths` key to indicate what kind of data it stores. An account can make multiple `STORAGE` claims, both to increase availability, as well as effectively sharding this off-chain data. This CIP does not specify what to do in case of conflicts.
 
 Here are some common ways of modeling applications into this format:
 
 - For transaction metadata, the account can encrypt the metadata and put it in a file with the transaction hash as the metadata. Further CIPs could specify a standard for thee structure of that metadata.
-    - `dataPath`: `/transactions/transactionHash`
+    - `dataPath`: `/transactions/${transactionHash}`
     - file contents: Encrypted, structured transaction metadata (comments, or otherwise)
 - Data in relation to other accounts could be communicated to other accounts with:
-    - `dataPath`: `/accounts/otherAccountAddress`
+    - `dataPath`: `/accounts/${otherAccountAddress}`
     - file contents: Structured data such as:
         - `sessionKeys`: The session key(s) encrypted with dataEncryptionKey of `otherAccount` on `Accounts.sol`
         - `attestations`: These could be signed data attestating things such an audit or badges. An attested account could also store the attestations from the attestor here, as a client can just verify the signature of the attestor.
@@ -49,12 +49,12 @@ Here are some common ways of modeling applications into this format:
     - `dataPath`: `/profile/name` or `/profile/picture`
     - file contents: the data unencrypted, or encrypted to session keys available to the recipients
 - Releases:
-    - `dataPath`: `/releases/hashOfRelease`
+    - `dataPath`: `/releases/${hashOfRelease}`
     - file contents: The signature of the key on the hash of the release
 
 If no assocation between to the `dataPath` should be leaked, the `dataPath` itself could be encrypted. The recipient would then just have to decrypt all `dataPath`s (which could be available in an index file).
 
-All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster. Signatures should be available under `${dataPath}.signature`. It is worth noting that this is only protecting against an adversary like the file hoster trying to provide unauthenticated plaintext. It does not protect against the adversary withholding or returning outdated data. If the order, including the existence, of the data is critical, then the developer should consider falling back to on-chain data, as consensus on the current view of the data is required. A possible partial workaround is to add expiry to the schema.
+All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster. Signatures should be available under `${dataPath}.signature`. It is worth noting that this is only protecting against an adversary like the file hoster trying to provide unauthenticated plaintext. It does not protect against the adversary withholding or returning outdated data. Additionally, this does not protect against the adversary swapping out files. If data is being used for authorization, developers should consider adding the datapath itself to the plaintext to be signed. If the order, including the existence, of the data is critical, then the developer should consider falling back to on-chain data, as consensus on the current view of the data is required. A possible partial workaround is to add expiry to the schema.
 
 ## Alternative Solutions
 
--- a/CIPs/0008.md
+++ b/CIPs/0008.md
@@ -52,6 +52,8 @@ Here are some common ways of modeling applications into this format:
     - `dataAddress`: `/releases/hashOfRelease`
     - file contents: The signature of the key on the hash of the release
 
+Application developers should consider using standards such as JSON-LD to not reinvent the wheel
+
 If no assocation between to the `dataAddress` should be leaked, the `dataAddress` itself could be encrypted. The recipient would then just have to decrypt all `dataAddress`s (which could be available in an index file).
 
 All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster. Signatures should be available under `${dataAddress}.signature`. It is worth noting that this is only protecting against an adversary like the file hoster trying to provide unauthenticated plaintext. It does not protect against the adversary withholding or returning outdated data. If the order, including the existence, of the data is critical, then the developer should consider falling back to on-chain data, as consensus on the current view of the data is required. A possible partial workaround is to add expiry to the schema.
--- a/CIPs/0008.md
+++ b/CIPs/0008.md
@@ -12,7 +12,7 @@ Examples include:
 - Transaction metadata such as comments or encrypted messaging in general
 - Profile information such as name or a profile picture
 - Unidirectional attestations like an audit or badges won in The Great Celo Stake Off
-- Signed statements/claims by a party for relase binaries/docker images/package hashes
+- Signed statements/claims by a party for release binaries/docker images/package hashes
 
 ## Goals
 
--- a/CIPs/0008.md
+++ b/CIPs/0008.md
@@ -29,34 +29,32 @@ We propose to add another claim, called the `STORAGE` claim:
   "type": "STORAGE",
   "timestamp": 1588886114907,
   "address": "https://example.com/celo_storage",
-  "filterDataAddresses": ".*"
+  "filterdataPaths": ".*"
 }
 ```
 
-This http address should be referred to as the root. The expectation here is that data can accessed via `root/${dataAddress}`. A root can specify a REGEX with the `filterDataAddresses` key to indicate what kind of data it stores. An account can make multiple `STORAGE` claims, both to increase availability, as well as effectively sharding this off-chain data. This CIP does not specify what to do in case of conflicts.
+This http address should be referred to as the root. The expectation here is that data can accessed via `root/${dataPath}`. A root can specify a REGEX with the `filterdataPaths` key to indicate what kind of data it stores. An account can make multiple `STORAGE` claims, both to increase availability, as well as effectively sharding this off-chain data. This CIP does not specify what to do in case of conflicts.
 
 Here are some common ways of modeling applications into this format:
 
 - For transaction metadata, the account can encrypt the metadata and put it in a file with the transaction hash as the metadata. Further CIPs could specify a standard for thee structure of that metadata.
-    - `dataAddress`: `/transactions/transactionHash`
+    - `dataPath`: `/transactions/transactionHash`
     - file contents: Encrypted, structured transaction metadata (comments, or otherwise)
 - Data in relation to other accounts could be communicated to other accounts with:
-    - `dataAddress`: `/accounts/otherAccountAddress`
+    - `dataPath`: `/accounts/otherAccountAddress`
     - file contents: Structured data such as:
         - `sessionKeys`: The session key(s) encrypted with dataEncryptionKey of `otherAccount` on `Accounts.sol`
         - `attestations`: These could be signed data attestating things such an audit or badges. An attested account could also store the attestations from the attestor here, as a client can just verify the signature of the attestor.
 - Profile information could be stored on this system with:
-    - `dataAddress`: `/profile/name` or `/profile/picture`
+    - `dataPath`: `/profile/name` or `/profile/picture`
     - file contents: the data unencrypted, or encrypted to session keys available to the recipients
 - Releases:
-    - `dataAddress`: `/releases/hashOfRelease`
+    - `dataPath`: `/releases/hashOfRelease`
     - file contents: The signature of the key on the hash of the release
 
-Application developers should consider using standards such as JSON-LD to not reinvent the wheel
+If no assocation between to the `dataPath` should be leaked, the `dataPath` itself could be encrypted. The recipient would then just have to decrypt all `dataPath`s (which could be available in an index file).
 
-If no assocation between to the `dataAddress` should be leaked, the `dataAddress` itself could be encrypted. The recipient would then just have to decrypt all `dataAddress`s (which could be available in an index file).
-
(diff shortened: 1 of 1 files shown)


CIPs/0008.md
```diff
@@ -0,0 +1,74 @@
+# CIP 0008: Expand Metadata to general off-chain storage
+
+- Date: 2020-05-05
+- Author: @nambrot
+- Status: WIP
+
+## Overview
+
+In CIP3, the Celo protocol has started with a simple off-chain storage mechanism mainly focused on claims an account can make about their off-chain identity and communicate to other participants in the network. Over time, there have been additional use cases which have similar access patterns that prompted the creation of `Metadata` in the first place. I.e. data that does not require the expensive cost of "hard consensus", instead just needs to be authenticated by a single party.
```
kobigurk, line 9:
```suggestion
In CIP8, the Celo protocol has started with a simple off-chain storage mechanism mainly focused on claims an account can make about their off-chain identity and communicate to other participants in the network. Over time, there have been additional use cases which have similar access patterns that prompted the creation of `Metadata` in the first place. I.e. data that does not require the expensive cost of "hard consensus", instead just needs to be authenticated by a single party.
```

CIPs/0008.md
```diff
@@ -0,0 +1,74 @@
...
+## Goals
+
+- Expand storage and indexing capabilities of this off-chain data
+- Encode the notion of decryption keys
+- Lightweight messaging system
+
+
+## Proposed Solution
+
+We propose to add another claim, called the `STORAGE` claim:
```
kobigurk, line 26: Where is the original claims mechanism described?
If it's not described in a previous CIP or a design document, it might be worth describing it here to answer questions about what it already supports, its integrity and authentication functionalities, etc.

CIPs/0008.md
```diff
@@ -0,0 +1,74 @@
...
+- Profile information could be stored on this system with:
+    - `dataAddress`: `profileName` or `profilePicture`
+    - file contents: the data unencrypted, or encrypted to session keys available to the recipients
+- Releases:
+    - `dataAddress`: `hashOfRelease`
+    - file contents: The signature of the key on the hash of the release
+
+If no assocation between to the `dataAddress` should be leaked, the `dataAddress` itself could be encrypted. The recipient would then just have to decrypt all `dataAddress`s (which could be available in an index file).
+
+All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster.
```
kobigurk, line 54: Should we more concretely specify how the signature happens so there won't be compatibility issues?

CIPs/0008.md
```diff
@@ -0,0 +1,74 @@
...
+        - `sessionKeys`: The session key(s) encrypted with dataEncryptionKey of `otherAccount` on `Accounts.sol`
+        - `attestations`: These could be signed data attestating things such an audit or badges
+- Profile information could be stored on this system with:
+    - `dataAddress`: `profileName` or `profilePicture`
+    - file contents: the data unencrypted, or encrypted to session keys available to the recipients
+- Releases:
+    - `dataAddress`: `hashOfRelease`
+    - file contents: The signature of the key on the hash of the release
+
+If no assocation between to the `dataAddress` should be leaked, the `dataAddress` itself could be encrypted. The recipient would then just have to decrypt all `dataAddress`s (which could be available in an index file).
```
kobigurk, line 52: This might be a nice use-case, similar to what Zcash does. The recipient tries to decrypt each encrypted tx data.

CIPs/0008.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+## Proposed Solution
+
+We propose to add another claim, called the `STORAGE` claim:
+```json
+{
+  "type": "STORAGE",
+  "timestamp": 1588886114907,
+  "address": "https://example.com/celo_storage"
```
kobigurk, line 31: Should we require TLS? It seems like a good idea.

CIPs/0008.md
kobigurk, line 9: Is there reference code to verify the CIP3 signature anywhere?
Currently it says to concatenate and hash, but not sure if it pays attention to whitespace, encoding, etc.

CIPs/0008.md
nambrot: Good point. I was mostly thinking that each payload is along the lines of `{ data: any, signature: string }` where signature is over `JSON.stringify(data)`

CIPs/0008.md
nambrot: Yeah, trying to find a good balance between privacy and performance though, as eventually this becomes quite untenable

CIPs/0008.md
nambrot: My sense is that with signatures on al payloads, it should be somewhat less necessary? Some folks might not have TLS setup on more raw storage setups

CIPs/0008.md
nambrot: Yeah it's not super well specified, but here is the code CK currently uses https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/identity/claims/claim.ts#L89

CIPs/0008.md
```diff
@@ -0,0 +1,76 @@
...
```
diminator, line 31: Typically the address or identifier would be resolvable in some form or fashion.
This can be:
- a service URI
- content-addressable storage
- something like a namespaced identifier
More information can be found here: https://github.com/decentralized-identity/universal-resolver

CIPs/0008.md
```diff
@@ -0,0 +1,76 @@
...
+## Goals
+
+- Expand storage and indexing capabilities of this off-chain data
+- Encode the notion of decryption keys
+- Lightweight messaging system
+
+
+## Proposed Solution
+
+We propose to add another claim, called the `STORAGE` claim:
```
diminator, line 26: What's not explicitly mentioned is where the `STORAGE` claim is stored. Does these actually go on-chain and the actual data is resolved using linked-data (ie. JSON-LD or IPLD)?

CIPs/0008.md
diminator: Additional comment:
should this claim have a proof section containing a signed hash of the claim for non-repudiability?

nambrot: So this would use Celo's existing metadata implementation which includes signatures by the account (or authorized signers)

I guess I'm not sure I understand the comment otherwise. What change are you suggesting?

CIPs/0008.md
nambrot: As part of the existing metadata implementation or possible on the `Accounts.sol` contract directly

CIPs/0008.md
```diff
@@ -0,0 +1,76 @@
...
+
+All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster.
+
+## Alternative Solutions
+
+I considered doing storing all data on IPFS, however the tooling around IPFS as well as IPNS does not seem user-friendly enough to make a substantial number of users use this directly (or centralize anyways). A generic "HTTP-based file system" is proposed as it is the lowest common denominator. It can be served by a simple HTTP server, even popular VCS static page hosting or specialized service providers. The lock-in effect is minimal since data can be easily migrated, and multiple copies could even be hosted with multiple roots. Thus availabillity concerns relative to on-chain storage can be mitigated relatively well.
+
+## Risks
+
+More reliance on off-chain data and misunderstanding such reliance could lead to inconsistency or poor usability. Ultimately this is traded off against cost of storing data on-chain or in centralized services.
```
timmoreton, line 64: A related issue is the ability to do a DoS attack on a single host.  One approach is to leave this to the hosting service; another is to provide mirrors by making the root address a list rather than a singleton (but then leading to possible inconsistencies); perhaps another (but less scalable) is to have an index file that is a signed mapping of paths to content hash, then have the directory structure contain content hashes. This opens the door for anyone to host or cache them in the future.

CIPs/0008.md
```diff
@@ -0,0 +1,76 @@
...
+We propose to add another claim, called the `STORAGE` claim:
+```json
+{
+  "type": "STORAGE",
+  "timestamp": 1588886114907,
+  "address": "https://example.com/celo_storage"
+}
+```
+
+Alternatively, this could be set directly on the `Accounts` smart contract directly, causing off-chain storage to effectively replace the existing metadata functionality. This would have the benefit of not duplicating similar functionality.
```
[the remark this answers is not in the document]
timmoreton: one slightly different approach: the metadata contains a separate entry for each external object eg `{ type: storage, subtype: 'badge', url: .., hash: ... }`. This doesn't scale if you wanted to use it for things like transaction metadata, but it does nicely make the objects unforgeable and also allows each to be hosted separately - e.g as an individual gist rather than a tree.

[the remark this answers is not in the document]
nambrot: >  it does nicely make the objects unforgeable

What do you mean by that?

> also allows each to be hosted separately

What use cases does this benefit? I expect that some of the benefit of this CIP is to tradeoff somewhat increased upfront setup with more marginal returns

CIPs/0008.md
```diff
@@ -0,0 +1,76 @@
...
+- Encode the notion of decryption keys
+- Lightweight messaging system
+
+
+## Proposed Solution
+
+We propose to add another claim, called the `STORAGE` claim:
+```json
+{
+  "type": "STORAGE",
```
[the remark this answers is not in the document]
nambrot: That's what is specified below I believe? I.e. applications on top of. this would have particular schemas they can derive independently (within some general guidelines)

CIPs/0008.md
nambrot: I believe I addressed that in the section above, but I do expect a nice trade-off between these concerns to have multiple storage providers with the data mirrored

CIPs/0008.md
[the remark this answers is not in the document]
timmoreton: oh, I missed the text saying refer to this address as `root`. Why not have the key be `root` in that case?

I do think we should set out how we expect identifiers at the level below that to be determined.

CIPs/0008.md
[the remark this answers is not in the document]
timmoreton: > > it does nicely make the objects unforgeable
> 
> What do you mean by that?

as it stands this scheme includes no unified way of determining if the contents have been changed.

CIPs/0008.md
timmoreton: so is `address` going to be an array then?

CIPs/0008.md
```diff
@@ -0,0 +1,76 @@
...
+    - `dataAddress`: `hashOfRelease`
+    - file contents: The signature of the key on the hash of the release
+
+If no assocation between to the `dataAddress` should be leaked, the `dataAddress` itself could be encrypted. The recipient would then just have to decrypt all `dataAddress`s (which could be available in an index file).
+
+All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster.
+
+## Alternative Solutions
+
+I considered doing storing all data on IPFS, however the tooling around IPFS as well as IPNS does not seem user-friendly enough to make a substantial number of users use this directly (or centralize anyways). A generic "HTTP-based file system" is proposed as it is the lowest common denominator. It can be served by a simple HTTP server, even popular VCS static page hosting or specialized service providers. The lock-in effect is minimal since data can be easily migrated, and multiple copies could even be hosted with multiple roots. Thus availabillity concerns relative to on-chain storage can be mitigated relatively well.
```
nategraf, line 60: Why may this proposal specific to either HTTP or IPFS? Can we design this system with decentralized storage solutions in mind and simply not have them implemented right away?

CIPs/0008.md
```diff
@@ -0,0 +1,76 @@
...
+- Profile information could be stored on this system with:
+    - `dataAddress`: `profileName` or `profilePicture`
+    - file contents: the data unencrypted, or encrypted to session keys available to the recipients
+- Releases:
+    - `dataAddress`: `hashOfRelease`
+    - file contents: The signature of the key on the hash of the release
+
+If no assocation between to the `dataAddress` should be leaked, the `dataAddress` itself could be encrypted. The recipient would then just have to decrypt all `dataAddress`s (which could be available in an index file).
+
+All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster.
```
nategraf, line 56: One alternative to specifying a signature to be included would be include a hash of the storage. If a directory tree structure is to be supported, it could be a hash tree of which the root is stored in the metadata entry.

CIPs/0008.md
```diff
@@ -0,0 +1,76 @@
...
+
+Examples include:
+- Transaction metadata such as comments or encrypted messaging in general
+- Profile information such as name or a profile picture
+- Unidirectional attestations like an audit or badges won in The Great Celo Stake Off
+- Signed statements/claims by a party for relase binaries/docker images/package hashes
+
+## Goals
+
+- Expand storage and indexing capabilities of this off-chain data
```
nategraf, line 19: I don't see any specification for indexing. Is that intended to be included here?

CIPs/0008.md
```diff
@@ -0,0 +1,76 @@
...
+- For transaction metadata, the account can encrypt the metadata and put it in a file with the transaction hash as the metadata. Further CIPs could specify a standard for thee structure of that metadata.
+    - `dataAddress`: `transactionHash`
+    - file contents: Encrypted, structured transaction metadata (comments, or otherwise)
+- Data in relation to other accounts could be communicated to other accounts with:
+    - `dataAddress`: `otherAccountAddress`
+    - file contents: Structured data such as:
+        - `sessionKeys`: The session key(s) encrypted with dataEncryptionKey of `otherAccount` on `Accounts.sol`
+        - `attestations`: These could be signed data attestating things such an audit or badges. An attested account could also store the attestations from the attestor here, as a client can just verify the signature of the attestor.
+- Profile information could be stored on this system with:
+    - `dataAddress`: `profileName` or `profilePicture`
```
nategraf, line 48: Is `profileName` a literal string or a placeholder?

CIPs/0008.md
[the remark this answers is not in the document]
nategraf: Does this mean that multiple storage claims is explicitly unsupported? If so, we should state that.

CIPs/0008.md
nambrot: You are right that there is fundamentally no reason why this CIP would need to specify HTTP vs. IPFS. I did still go with the recommendation of HTTP to avoid having to add running an IPFS stack to what can be reasonably required from participants on the Celo network

CIPs/0008.md
nambrot: What are the advantages of your suggestion?

CIPs/0008.md
nambrot: The indexing would be implicit in the usage of `dataAddresses` which effectively acts as an index into the files stored in this system

CIPs/0008.md
nambrot: It is proposed here in the example as a literal string

CIPs/0008.md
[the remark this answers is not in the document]
nambrot: I'll amend the CIP to support multiple storage roots that can be exclusive in the data they host

nambrot:
> submitted as metadata cannot expire and is valid in perpetuity.

That is a great point that I had not considered sufficiently. The lack of being able to "revoke" signatures in this system does somewhat limit the value of the system. For example attestations on this system would be very hard to revoke as you mention. I think that it's probably an acceptable trade-off (once you need order, you need consensus). As you mention, it should be explicitly stated as such. Possible work-arounds are to provide expiry-timestamps.

> Also, it seems like this CIP proposes to support some of the functionality required by the mobile app. If so, would cLabs then be providing the data root for the mobile app users? It does not need to be part of the CIP, but I am curious.

The CIP was certainly written with those considerations in mind. I.e., the vast majority of end-users like users of the cLabs mobile wallet are unlikely to want to know about this let alone provide their own storage system. It would seem to me that cLabs as the developer of the wallet would provide this to their users. The prompting use case of this CIP actually was transaction metadata that we is currently being stored on-chain via `transferWithComment` which increases gas costs of simple transfers by 20+%.

> we should get better about writing these CIPs as actual specifications

Like many features of the Celo platform, they would greatly benefit from documentation including actual specifications. That being said, I'm not sure that requiring it at this stage is a favorable trade-off as it does increase the burden of change. However, not a strong opinion of mine.

CIPs/0008.md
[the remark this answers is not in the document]
nambrot: Yeah that is a good point, it's related to @nategraf's point that revocation is impossible in the current system, similar to detecting if contents have changed. Either of those applications have a temporal order in common, and I believe once an application requires that, it starts to become unsuitable for this approach?

CIPs/0008.md
nambrot: @timmoreton brought up that it would be nice if the files could be taken as is, and signatures are available under `{dataAddress}.signature`

CIPs/0008.md
nambrot: I'd imagine one could have multiple storage claims

nambrot:
I just realized that another way of adding expiry is to rotate the authorized signer, especially once we make it more granular with CIP10

CIPs/0008.md
```diff
@@ -28,32 +28,35 @@ We propose to add another claim, called the `STORAGE` claim:
...
   "type": "STORAGE",
   "timestamp": 1588886114907,
-  "address": "https://example.com/celo_storage"
+  "address": "https://example.com/celo_storage",
+  "filterDataAddresses": ".*"
 }
 ```
 
-Alternatively, this could be set directly on the `Accounts` smart contract directly, causing off-chain storage to effectively replace the existing metadata functionality. This would have the benefit of not duplicating similar functionality.
+This http address should be referred to as the root. The expectation here is that data can accessed via `root/${dataAddress}`. A root can specify a REGEX with the `filterDataAddresses` key to indicate what kind of data it stores. An account can make multiple `STORAGE` claims, both to increase availability, as well as effectively sharding this off-chain data. This CIP does not specify what to do in case of conflicts.
```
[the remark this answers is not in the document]
nambrot: I tried to avoid overly complicating the specification of what kind of data a root would host. When in doubt, one could always the wild card for everything? I imagine this should cover most use cases to start?

As mentioned, I didn't feel that the CIP should specify what to do when two roots host the different data at the same path. This feels quite application specific to me

CIPs/0008.md
```diff
@@ -28,32 +28,35 @@ We propose to add another claim, called the `STORAGE` claim:
...
 
-Either way, this http address should be referred to as the root. The expectation here is that data can accessed via `root/${dataAddress}`. For private data, this should be encrypted to the recipient. Here are some common ways of modeling applications into this format:
+Here are some common ways of modeling applications into this format:
 
 - For transaction metadata, the account can encrypt the metadata and put it in a file with the transaction hash as the metadata. Further CIPs could specify a standard for thee structure of that metadata.
-    - `dataAddress`: `transactionHash`
+    - `dataAddress`: `/transactions/transactionHash`
```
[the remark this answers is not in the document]
nambrot: Agreed, renamed it to `dataPath`

CIPs/0008.md
```diff
@@ -28,32 +28,35 @@ We propose to add another claim, called the `STORAGE` claim:
...
 - Profile information could be stored on this system with:
-    - `dataAddress`: `profileName` or `profilePicture`
+    - `dataAddress`: `/profile/name` or `/profile/picture`
     - file contents: the data unencrypted, or encrypted to session keys available to the recipients
 - Releases:
-    - `dataAddress`: `hashOfRelease`
+    - `dataAddress`: `/releases/hashOfRelease`
     - file contents: The signature of the key on the hash of the release
 
+Application developers should consider using standards such as JSON-LD to not reinvent the wheel
```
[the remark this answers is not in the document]
nambrot: I've taken it out for now. I was originally thinking that these paths could represent semantic links indeed, but the "host" would have to be some synthetic host like `celo://offchain-${accountAddress}/`. However since that is not core to this CIP I'm gonna leave it as a future improvement

I do think it would be valuable as it aids in making the data more interoperable

CIPs/0008.md
nategraf: Given the root hash for a storage tree at a given point in time, it would allow verification of the storage without needing to sign every piece of data in it. It would also make it impossible for someone to add or remove data from storage, including via MitM attack on the requestor, or someone compromising the storage server without access to the key that signs the metadata. Of course, it would require the root hash to be updated everytime a piece of data is added or removed, so it may be too cumbersome to use.

nategraf:
One concern that I have about moving in this direction is that having users of the Celo protocol access random web servers during normal operation should not be assumed to be safe. Especially with regard to the mobile wallet, if we allow users to host metadata and storage at arbitrary locations that the mobile app then accesses, it could leak sensitive private data, for example they location, wallet address, who they are transacting with etc... A solution to this might be to only allow the mobile app to access a whitelist of URLs, but doing so would hamstring this proposal, and as such should be considered before we go further in this direction.

CIPs/0008.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Proposed Solution
+
+We propose to add another claim, called the `STORAGE` claim:
+```json
+{
+  "type": "STORAGE",
+  "timestamp": 1588886114907,
+  "address": "https://example.com/celo_storage",
+  "filterdataPaths": ".*"
```
nategraf, line 32: This has a nasty sharp edge that will allow for DoS attacks against clients that attempt to evaluate this regex. I would suggest using a simpler glob scheme, or at least adding a warning here https://en.wikipedia.org/wiki/ReDoS

CIPs/0008.md
```diff
@@ -0,0 +1,77 @@
...
+    - `dataPath`: `/accounts/otherAccountAddress`
+    - file contents: Structured data such as:
+        - `sessionKeys`: The session key(s) encrypted with dataEncryptionKey of `otherAccount` on `Accounts.sol`
+        - `attestations`: These could be signed data attestating things such an audit or badges. An attested account could also store the attestations from the attestor here, as a client can just verify the signature of the attestor.
+- Profile information could be stored on this system with:
+    - `dataPath`: `/profile/name` or `/profile/picture`
+    - file contents: the data unencrypted, or encrypted to session keys available to the recipients
+- Releases:
+    - `dataPath`: `/releases/hashOfRelease`
+    - file contents: The signature of the key on the hash of the release
```
nategraf, line 53: I would be wary of including these "examples" here unless you intend them to actually be implemented. Also could you clarify which parts of these example paths are variable and which are literal?

CIPs/0008.md
```diff
@@ -0,0 +1,77 @@
...
+
+If no assocation between to the `dataPath` should be leaked, the `dataPath` itself could be encrypted. The recipient would then just have to decrypt all `dataPath`s (which could be available in an index file).
+
+All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster. Signatures should be available under `${dataPath}.signature`. It is worth noting that this is only protecting against an adversary like the file hoster trying to provide unauthenticated plaintext. It does not protect against the adversary withholding or returning outdated data. If the order, including the existence, of the data is critical, then the developer should consider falling back to on-chain data, as consensus on the current view of the data is required. A possible partial workaround is to add expiry to the schema.
```
nategraf, line 57: In addition to using old data or not returning data, a MitM or the file server may choose to return data for a different path. This could be extremely problematic in cases where, the data from another path is useful in the context of the changed path. For example, the hosting server could return the data for `/accounts/${attackerAddress}/sessionKey` to the query `/accounts/${realAddress}/sessionKey` and the client would be able to verify the signature without knowing the issue.

One solution to this might be to include the path in the signed data

CIPs/0008.md
asaj: +1 to this question

CIPs/0008.md
```diff
@@ -0,0 +1,77 @@
...
+## Goals
+
+- Expand storage and indexing capabilities of this off-chain data
+- Encode the notion of decryption keys
+- Lightweight messaging system
+
+
+## Proposed Solution
+
+We propose to add another claim, called the `STORAGE` claim:
```
asaj, line 26: What are the existing claim types already?

CIPs/0008.md
nambrot: Yeah it seems prohibitive to me to update the root hash for every change, what do you think?

CIPs/0008.md
nambrot: That is an extremely good point

CIPs/0008.md
nambrot: Updated it to use a glob scheme like https://github.com/isaacs/minimatch

CIPs/0008.md
nambrot: I agree that this shouldn't ultimately be in any kind of spec, but received feedback that examples here were useful in aiding understanding. Updated to make it clearer which are variables vs. literals

CIPs/0008.md
nambrot: Added it as a consideration

CIPs/0008.md
nambrot: https://docs.celo.org/celo-codebase/protocol/identity/metadata

- Name Claim - An account can claim a human-readable name. This claim is not verifiable.
- Attestation Service URL Claim - For the lightweight identity layer, validators can make a claim under which their Attestation Service is reachable to provide attestations. This claim is not verifiable.
- Keybase User Claim - Accounts can make claims on Keybase usernames. This claim is verifiable by signing a message with the account and hosting it on the publicly accessible path of the Keybase file system.
- Domain Claim - Accounts can make claims on domain names. This claim is verifiable by signing a message with the account and embedding it in a TXT record.

CIPs/0008.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Overview
+
+In CIP3, the Celo protocol has started with a simple off-chain storage mechanism mainly focused on claims an account can make about their off-chain identity and communicate to other participants in the network. Over time, there have been additional use cases which have similar access patterns that prompted the creation of `Metadata` in the first place. I.e. data that does not require the expensive cost of "hard consensus", instead just needs to be authenticated by a single party.
+
+Examples include:
+- Transaction metadata such as comments or encrypted messaging in general
+- Profile information such as name or a profile picture
+- Unidirectional attestations like an audit or badges won in The Great Celo Stake Off
+- Signed statements/claims by a party for relase binaries/docker images/package hashes
```
codyborn, line 15:
```suggestion
- Signed statements/claims by a party for release binaries/docker images/package hashes
```

CIPs/0008.md
```diff
@@ -0,0 +1,77 @@
...
+    - `dataPath`: `/releases/${hashOfRelease}`
+    - file contents: The signature of the key on the hash of the release
+
+If no assocation between to the `dataPath` should be leaked, the `dataPath` itself could be encrypted. The recipient would then just have to decrypt all `dataPath`s (which could be available in an index file).
+
+All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster. Signatures should be available under `${dataPath}.signature`. It is worth noting that this is only protecting against an adversary like the file hoster trying to provide unauthenticated plaintext. It does not protect against the adversary withholding or returning outdated data. Additionally, this does not protect against the adversary swapping out files. If data is being used for authorization, developers should consider adding the datapath itself to the plaintext to be signed. If the order, including the existence, of the data is critical, then the developer should consider falling back to on-chain data, as consensus on the current view of the data is required. A possible partial workaround is to add expiry to the schema.
```
m-chrzan, line 57: > It does not protect against the adversary withholding or returning outdated data. Additionally, this does not protect against the adversary swapping out files.

One somewhat lightweight solution to this would be to store the hash of the latest signed contents on-chain. Thus only a small constant amount of storage needs to be updated even with large changes/additions to a user's off-chain metadata.

nambrot: I was thinking about it, but it seemed like the cost of effectively providing a merkle tree root that you update on-chain seemed quite heavy and at that point one is replicating the blockchain itself?

nambrot:
Merging this in as WIP since @AlexBHarley and I are going to be working on this. We will update the CIP with any modifications that will come up during development.

nambrot:
Feel free to continue discussion here or on Discord at #celo-improvement-proposals

PR #15, opened 2020-05-14
CIP [0009]: Standalone signing service for Validators

asaj:
WIP

### DIFF
--- a/CIPs/0009.md
+++ b/CIPs/0009.md
@@ -17,7 +17,7 @@ Switching nodes allows validators to upgrade to newer versions of `celo-blockcha
 Validators who wish to switch nodes currently have the following options:
 
 1. Bring down node A, bring up node B with key material copied from node A.
-2. Bring up node B with key material copied from node A, bring down node B.
+2. Bring up node B with key material copied from node A, bring down node A.
 3. Bring up node B with new key material, authorize the new key, and bring down node A after the next election.
 
 Option `1` results in downtime, option `2` risks double signing, and option `3` can only be done once every 24 hours.
@@ -60,6 +60,7 @@ Some options:
 
 - Implement a `clef` RPC to switch the value of the `--rpcaddr` flag.
 - Have `clef` accept RPCs from both nodes, and use `miner.start` and `miner.stop` on the node (or similar RPCs that allow these operations to be scheduled)
+- Implement double signing protection directly in `clef`.
 
 #### Threshold signatures
 Using threshold signatures for block finalization is an approach taken by [Polychain Labs' Cosmos validator](https://blog.polychainlabs.com/tendermint/2020/03/26/threshold-validator-for-tendermint.html) to improve security and availability.
@@ -78,9 +79,12 @@ Multiple `celo-blockchain` nodes with a failover mechanism would protect against
 
 ## Risks
 
-- The new node will not have the same consensus DB, which introduces a small risk of double signing. One option to mitigate this is for `clef` to stop accepting RPCs from node A and wait a block before accepting RPs from node B.
-- Upgrades to `clef` may still require key rotations. This could be mitigated by threshold signatures; so long as `M > N/2` individual `clef` instances could be upgraded one at a time with zero downtime.
+- The new node will not have the same consensus DB, which introduces a small risk of double signing. One option to mitigate this is for `clef` to stop accepting RPCs from node A and wait a block before accepting RPCs from node B.
+- Upgrades to `clef` may still require key rotations. This could be mitigated by threshold signatures; so long as `M > N/2` individual `clef` instances could be upgraded one at a time with zero downtime. Even without threshold signatures, `clef` would likely restart much more quickly than `celo-blockchain`, allowing `clef` to be upgraded without affecting the validator score.
 
 ## Implementation
 
 * To follow.
+
+## Future work
+Currently the ECDSA key used to sign consensus messages can also change the BLS key for a validator. This ECDSA key will need to live on at least one `clef` instance per validator so that the validator can participate in consensus. To improve validator security, the responsibility of signing consensus messages and managing the validator should be split into separate keys, the latter of which can then be kept offline.
--- a/CIPs/0009.md
+++ b/CIPs/0009.md
@@ -42,6 +42,7 @@ To switch nodes, validators will issue a command to `clef` to stop accepting RPC
 
 ### Implementation
 
+#### Wallet interface
 The current `clef` implementation of the `wallet` interface is missing the following functions needed to support running a validator:
 - Decrypt
 - SignHashBLS
@@ -52,13 +53,33 @@ The current `clef` implementation of the `wallet` interface is missing the follo
 
 To implement these functions, new RPCs will need to be added to `clef`.
 
+#### Node mutex
 More investigation will need to be done to figure out exactly how to atomically switch nodes.
 
-This may be as simple as implementing an RPC to switch the value of the `--rpcaddr` flag.
+Some options:
+
+- Implement a `clef` RPC to switch the value of the `--rpcaddr` flag.
+- Have `clef` accept RPCs from both nodes, and use `miner.start` and `miner.stop` on the node (or similar RPCs that allow these operations to be scheduled)
+
+#### threshold signatures
+Using threshold signatures for block finalization is an approach taken by [Polychain Labs on Cosmos](https://blog.polychainlabs.com/tendermint/2020/03/26/threshold-validator-for-tendermint.html) to improve security and availability.
+
+Celo is a natural candidate to adopt a similar feature for BLS block finalization signatures and ECDSA consensus message signatures.
+
+A standalone signing service would simplify the implementation of this feature.
+
+Validators could run `N` instances of `clef`, and one or more instances of `celo-blockchain`.
+
+The `celo-blockchain` `external` wallet could be configured to ping each of the `clef` instances for a signature share, and return once `M` have been collected.
+
+The `M/N` threshold signature would protect against `N-M` `clef` instances being unavailable or compromised.
+
+Multiple `celo-blockchain` nodes with a failover mechanism would protect against a single node being unavailable.
 
 ## Risks
 
 - The new node will not have the same consensus DB, which introduces a small risk of double signing. One option to mitigate this is for `clef` to stop accepting RPCs from node A and wait a block before accepting RPs from node B.
+- Upgrades to `clef` may still require key rotations. This could be mitigated by threshold signatures; so long as `M > N/2` individual `clef` instances could be upgraded one at a time with zero downtime.
 
 ## Implementation
 
--- a/CIPs/0009.md
+++ b/CIPs/0009.md
@@ -1,0 +1,65 @@
+# CIP [0009]: Standalone signing service for Validators
+
+- Date: 2020-05-14
+- Author: @asaj
+- Status: WIP
+
+## Overview
+
+Validator operational excellence is vital to the security and performance of the Celo network.
+
+However, validators currently face significant operational limitations due to key management being bundled into `celo-blockchain`.
+
+In particular, it makes it cumbersome to switch which node is used to participate in consensus.
+
+Switching nodes allows validators to upgrade to newer versions of `celo-blockchain` or to failover to a different node if the primary node fails.
+
+Validators who wish to switch nodes currently have the following options:
+
+1. Bring down node A, bring up node B with key material copied from node A.
+2. Bring up node B with key material copied from node A, bring down node B.
+3. Bring up node B with new key material, authorize the new key, and bring down node A after the next election.
+
+Option `1` results in downtime, option `2` risks double signing, and option `3` can only be done once every 24 hours.
+Furthermore, options `1` and `2` may require unsafe key management practices, which may result in keys being compromised.
+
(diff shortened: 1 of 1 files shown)


CIPs/0009.md
```diff
@@ -0,0 +1,65 @@
...
+Decoupling key management from other validator functionality will allow validators to switch nodes with zero downtime and no risk of double signing.
+
+## Goals
+
+- Zero downtime validator upgrades within an epoch
+- Failover possible with no risk of double signing
+
+## Proposed Solution
+
+The proposed solution is to use `clef`, introduced in [go-ethereum v1.9](https://blog.ethereum.org/2019/07/10/geth-v1-9-0/), as a standalone signer for validators.
```
nambrot, line 35: Great idea

CIPs/0009.md
```diff
@@ -0,0 +1,65 @@
...
+
+`clef` is integrated into `celo-blockchain` as an implementation of the [wallet](https://github.com/celo-org/celo-blockchain/blob/3caca596d1ff23a9faf5339f108c993f38bfa743/accounts/accounts.go#L50) interface in [accounts/external/backend.go](https://github.com/celo-org/celo-blockchain/blob/master/accounts/external/backend.go).
+
+Validators will run a `clef` instance that manages their [validator signer key](https://docs.celo.org/validator-guide/summary/detailed#authorized-validator-signers), and point their node to that instance.
+
+To switch nodes, validators will issue a command to `clef` to stop accepting RPCs from node A and start accepting RPCs from node B.
```
nambrot, line 41: Is that something that `clef` can do natively, or would validators be expected to manage this with their own networking/firewall setup

CIPs/0009.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Implementation
+
+The current `clef` implementation of the `wallet` interface is missing the following functions needed to support running a validator:
+- Decrypt
```
nambrot, line 46: I know this increases scope, it would be great if in the process we could add encrypt and decrypt functionality on the keys, somewhat related to various efforts on Ethereum as well https://github.com/ethereum/EIPs/issues/130

CIPs/0009.md
nambrot: Would this be mandatory or opt-in to use clef externally?

asaj: opt-in

CIPs/0009.md
asaj: I'm suggesting to add it natively to `clef` but open to ideas

CIPs/0009.md
```diff
@@ -0,0 +1,86 @@
...
+
+Celo is a natural candidate to adopt a similar feature for BLS block finalization signatures and ECDSA consensus message signatures.
+
+A standalone signing service would simplify the implementation of this feature.
+
+Validators could run `N` instances of `clef`, and one or more instances of `celo-blockchain`.
+
+The `celo-blockchain` nodes could be configured to ping each of the `clef` instances for a signature share, and return once `M` have been collected.
+
+The `M/N` threshold signature would protect against `N-M` `clef` instances being unavailable or compromised.
```
nvesely, line 75: If the ECDSA key lives on a single machine, then an adversary A only has to compromise that single machine to authorize a new BLS public key of A's choice. To achieve `M/N` threshold security for signing, threshold ECDSA signing would need to be implemented as well.

Note BLS public keys can only be updated once a day, so compromise of the ECDSA key is not immediately equivalent to compromise of the BLS key (although within 24 hours it is).

kobigurk: That's a good point. It would be better to have an ECDSA consensus key that cannot change BLS keys.

CIPs/0009.md
```diff
@@ -0,0 +1,86 @@
...
+- Zero downtime validator upgrades within an epoch
+- Failover possible with no risk of double signing
+
+## Proposed Solution
+
+The proposed solution is to use `clef`, introduced in [go-ethereum v1.9](https://blog.ethereum.org/2019/07/10/geth-v1-9-0/), as a standalone signer for validators.
+
+`clef` is integrated into `celo-blockchain` as an implementation of the [wallet](https://github.com/celo-org/celo-blockchain/blob/3caca596d1ff23a9faf5339f108c993f38bfa743/accounts/accounts.go#L50) interface in [accounts/external/backend.go](https://github.com/celo-org/celo-blockchain/blob/master/accounts/external/backend.go).
+
+Validators will run a `clef` instance that manages their [validator signer key](https://docs.celo.org/validator-guide/summary/detailed#authorized-validator-signers), and point their node to that instance.
```
nategraf, line 39: Clef's security model assumes that the RPC signer link is untrusted and that the user will review and approve every transaction through a UI of some sort. In the validator signer use-case, there will be no user, so presumably Clef would be configured to sign anything it sees and the RPC connection would have to become trusted. Doing so will likely be nontrivial as it goes against the original design assumptions of Clef. 

A quick look at the Clef repo shows that the two options for RPC connection are HTTP and IPC, so either A) support for a secure connection protocol with mutual authentication (e.g. Noise or TLS) will have to added, or B) an external proxy used to proxy the IPC connection over the network. Option B is likely easier, but does add some operational overhead and provides a weaker security boundary.

Once RPC over an encrypted and mutually-authenticated link is supported, switching which validator node is allowed to talk to Clef could be done be removing the public key of the old validator node, and adding the public key of the new one, preferably in an atomic transaction. Upon calling, the old validator should no longer be authorized to make connections and any existing connection should be dropped or at least deauthorized from calling RPCs.

An open question here would be "who is authorized to call initiate a connection handover?" In the Clef model, the UI (`stdin`) Would be the the best candidate, which would essentially require the operator to `ssh` into the machine.

CIPs/0009.md
```diff
@@ -0,0 +1,86 @@
...
+- SignHashBLS
+- SignMessageBLS
+- GenerateProofOfPossession
+- GenerateProofOfPossessionBLS
+- GetPublicKey
+
+To implement these functions, new RPCs will need to be added to `clef`.
+
+#### Node mutex
+More investigation will need to be done to figure out exactly how to atomically switch nodes.
```
nategraf, line 57: An atomic swap should not be assumed prevent a failure to sign. An example would be the following ordering of events when switching validator `V` switches from node `Va` to node `Vb`:

1. Validator `V` is the proposer for block `x`
2. Node `Vb` assembles a block and sends a signing request to `clef`. It is rejected.
3. Operator switches the authorized node from `Va` to `Vb`
4. Node `Va` assembles a block and sends a signing request to `clef`. It is rejected.
5. Validator `V` fails to submit a proposal for block `x` and a round change occurs.

Adding `Vb`, then after a few seconds removing `Va` would be less likely to cause an issue, but obviously causes contention and potential for double signing. A simple solution for this would be to configure Clef to support decoding the consensus messages and only allowing a single block hash to be signed per sequence number and round number by implementing a minimal set of the IBFT consensus rules.

CIPs/0009.md
nvesely: What if the last BLS key has to sign off on the new one? As long as the initial BLS public key, which is still signed by the ECDSA public key, is the same as the one resulting from the DKG, then (m, n)-threshold security should hold for future key rotation updates.

There may be other good reasons to implement ECDSA threshold signatures, but if we can get the same security properties by relying just on our threshold BLS implementation, that keeps our attack surface smaller.

CIPs/0009.md
```diff
@@ -0,0 +1,86 @@
...
+The `celo-blockchain` nodes could be configured to ping each of the `clef` instances for a signature share, and return once `M` have been collected.
+
+The `M/N` threshold signature would protect against `N-M` `clef` instances being unavailable or compromised.
+
+Multiple `celo-blockchain` nodes with a failover mechanism would protect against a single node being unavailable.
+
+## Risks
+
+- The new node will not have the same consensus DB, which introduces a small risk of double signing. One option to mitigate this is for `clef` to stop accepting RPCs from node A and wait a block before accepting RPs from node B.
+- Upgrades to `clef` may still require key rotations. This could be mitigated by threshold signatures; so long as `M > N/2` individual `clef` instances could be upgraded one at a time with zero downtime.
```
nategraf, line 82: One advantage of this approach worth pointing out though is that `clef` would likely restart quite quickly, well under 1 minutes. `celo-blockchain` currently takes over 1 minute to fully reboot and get connected again, which is one of the main issues that is making it hard to operate with 0 downtime.

CIPs/0009.md
nvesely: If you don't have all the nodes verify all the consensus messages themselves, then you sacrifice some of the security benefits of the (m, n)-threshold signatures. So I think each node should in some sense be a full node, and then the one running clef acts as the aggregator, which puts signatures together from partial signatures.

That said, the nodes may need to exchange some messages besides the partial signatures (which are safe to send in the clear), and for those we probably want mutual authentication (and possibly encryption). Although nodes could sign RPC messages with their ECDSA/ BLS threshold keys, for throughput I agree Noise/TLS is a good option. In your option B do you say a "weaker security boundary" because that option involves some sort of IPC or virtual network activity that might be tampered with by malicious but unprivileged processes on the same machine that wouldn't otherwise be able to do so?

CIPs/0009.md
```diff
@@ -0,0 +1,86 @@
...
+However, validators currently face significant operational limitations due to key management being bundled into `celo-blockchain`.
+
+In particular, it makes it cumbersome to switch which node is used to participate in consensus.
+
+Switching nodes allows validators to upgrade to newer versions of `celo-blockchain` or to failover to a different node if the primary node fails.
+
+Validators who wish to switch nodes currently have the following options:
+
+1. Bring down node A, bring up node B with key material copied from node A.
+2. Bring up node B with key material copied from node A, bring down node B.
```
mcortesi, line 20: typo: bring down A

CIPs/0009.md
```diff
@@ -0,0 +1,86 @@
+# CIP [0009]: Standalone signing service for Validators
+
+- Date: 2020-05-14
+- Author: @asaj
+- Status: WIP
+
+## Overview
```
mcortesi, line 7: Love the idea....

Not sure if `clef` is the answer... but something like `clef`. I would explore the needs of this `remote signer`, i believe clef feature set tries to address a similiar but different problem. And that building a custom remote signer with ingrained feature to do the mutex might be a better impl strategy.

At least, i would do a small research about that as an option.

CIPs/0009.md
nategraf: I don't mean to suggest that nodes should off-load consensus message processing. Instead, I am suggesting that one way to prevent double signing is to implement some IBFT checks in the Clef sign such that it will refuses to sign two blocks at the same height and round number, to prevent risk of double signing.

To clarify, I am discussing here the communication between the validator nodes and the Clef signer node, which would need to be over an encrypted link. I am not discussing the future case of validator nodes or signer nodes talking between each other to implement a threshold scheme.

On the "weaker security boundary", yes that's exactly what I mean 👍

CIPs/0009.md
```diff
@@ -0,0 +1,86 @@
...
+
+The `celo-blockchain` nodes could be configured to ping each of the `clef` instances for a signature share, and return once `M` have been collected.
+
+The `M/N` threshold signature would protect against `N-M` `clef` instances being unavailable or compromised.
+
+Multiple `celo-blockchain` nodes with a failover mechanism would protect against a single node being unavailable.
+
+## Risks
+
+- The new node will not have the same consensus DB, which introduces a small risk of double signing. One option to mitigate this is for `clef` to stop accepting RPCs from node A and wait a block before accepting RPs from node B.
```
hqueue, line 81: nit: RPs -->  RPCs

CIPs/0009.md
asaj: Sorry, I'm not clear why extending `clef` isn't the best option here

CIPs/0009.md
asaj: Agreed that clef/node communication will need to be secure.

CIPs/0009.md
asaj: That seems orthogonal to me

CIPs/0009.md
asaj: > A simple solution for this would be to configure Clef to support decoding the consensus messages and only allowing a single block hash to be signed per sequence number and round number by implementing a minimal set of the IBFT consensus rules.

I like this

CIPs/0009.md
asaj: I think the best option here is to have an ECDSA key that is *only* allowed to sign consensus messages, as @kobigurk suggested

asaj: Added in a *future work* section below

asaj: Or to move to only use BLS for consensus messages..

CIPs/0009.md
```diff
@@ -0,0 +1,90 @@
+# CIP [0009]: Standalone signing service for Validators
+
+- Date: 2020-05-14
+- Author: @asaj
```
YazzyYaz, line 4: Hi Asa @asaj thanks for submitting this proposal. Can you update your PR to reflect the new CIP template that's been activated as part of the new CIP process. Link to it here: https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

YazzyYaz (changes requested):

PR #17, opened 2020-05-15
CIP10: General Account Signer Authorization

nambrot:
WIP

### DIFF
--- a/CIPs/0010.md
+++ b/CIPs/0010.md
@@ -1,0 +1,56 @@
+# CIP 0010: General Account Signer Authorization
+
+- Date: 2020-05-13
+- Author: @nambrot
+- Status: WIP
+
+## Overview
+
+The `Accounts.sol` smart contract houses a general account abstraction that currently services the following purposes:
+- Acts as the canonical account for the Proof-of-stake system (aka LockedGold account)
+- keeps track of authorized signer keys in the PoS system (validator, voter and attestation signers)
+- stores basic profile information such as a name and mapping to `dataEncryptionKey` and `walletAddress`
+- entry into an account's off-chain metadata
+
+Over the past months, it has become noticable that especially the authorization of signer keys by the "account key" is a very useful method to authenticate the "account owner" without having to unncessarily bring the "account key" online which is the most priveledged, and thus most important to secure. Example of such cases have been:
+
+- The signing of the off-chain metadata should ideally not happen with the account key. During the development of the `RelaeseGold` smart contract it has been found that smart contracts as accounts cannot sign metadata. As a workaround, verifying metadata has been adjusted to allow any authorized signer to sign the payload, which violates the intention behind having well scoped authrized signers in the first place.
+- As part of PGPNP, accounts have to authenticate with the PGPNP service for rate-limiting purposes. To do that, the "account key" has to sign the request which also violates the intention of keeping keys that can move funds as cold as possible.
+- For other applications such as Profile Picture/Names or more down-the-road applications such as messaging, specific encryption keys are going to be useful. Additionally, the `dataEncryptionKey` seems out of place, despite it being effectively also an authorized key (if not signer).
+
+As applications get developed on Celo, it becomes evident that this mechanism of authorizing signers is useful beyond the PoS system and this CIP proposes standardization to facilitate this good practice.
+
+## Goals
+
+- Reduce the chance of fund loss on Celo
+- Increase the ergonomics of key management on Celo
+
+## Proposed Solution
+
+Expand the `Signers` struct in `Accounts.sol` with a generic mapping and generic version of `authorizeXSigner`, `removeXSigner`, `getXSigner` and `signerToAccount` that take as additional input `string namespace`.
+
+- multiple account keys? (effectively making lockeedgold somewhat of a custody solution)
+- authorization key
+- makes key management easier by allowing tools implement once interface
+
+## Alternative Solutions
+
+Alternately, applications could always derive their own authorization scheme/smart contracts. However, it would not benefit of the tooling that could be possible to faciliate key management.
+
+Instead of extending the `Signres` struct, a completely different smart contract `AuthorizedSigners` could be created. Functionality would be practically duplicated, but separated from the critical PoS system while gaining the benefit from a standardized authorizaton scheme.
+
+## Risks
+
+Upgrading core smart contracts + RG contracts.
+
+## Useful Links
+
+* Optional section
+* Links to related CIPs or other documents
+
+## Implementation
+
+* To follow.
+* Or a link to a PR.
+
+Add any implementation considerations here. For all proposals going through the governance process, this section should reference the code implementing the proposed change. It’s recommended to get community feedback before writing any code.


nategraf:
I like this proposal a lot! I would love to see a name-spacing scheme included as part of this proposal. Giving more flexibility to users as to how many keys they end up managing would be a great improvement to usability and security!

CIPs/0010.md
```diff
@@ -0,0 +1,56 @@
...
+## Goals
+
+- Reduce the chance of fund loss on Celo
+- Increase the ergonomics of key management on Celo
+
+## Proposed Solution
+
+Expand the `Signers` struct in `Accounts.sol` with a generic mapping and generic version of `authorizeXSigner`, `removeXSigner`, `getXSigner` and `signerToAccount` that take as additional input `string namespace`.
+
+- multiple account keys? (effectively making lockeedgold somewhat of a custody solution)
```
asaj, line 32: What's going on with these bullets?

CIPs/0010.md
```diff
@@ -0,0 +1,56 @@
...
+- authorization key
+- makes key management easier by allowing tools implement once interface
+
+## Alternative Solutions
+
+Alternately, applications could always derive their own authorization scheme/smart contracts. However, it would not benefit of the tooling that could be possible to faciliate key management.
+
+Instead of extending the `Signres` struct, a completely different smart contract `AuthorizedSigners` could be created. Functionality would be practically duplicated, but separated from the critical PoS system while gaining the benefit from a standardized authorizaton scheme.
```
asaj, line 40: typo

CIPs/0010.md
```diff
@@ -0,0 +1,56 @@
...
+- For other applications such as Profile Picture/Names or more down-the-road applications such as messaging, specific encryption keys are going to be useful. Additionally, the `dataEncryptionKey` seems out of place, despite it being effectively also an authorized key (if not signer).
+
+As applications get developed on Celo, it becomes evident that this mechanism of authorizing signers is useful beyond the PoS system and this CIP proposes standardization to facilitate this good practice.
+
+## Goals
+
+- Reduce the chance of fund loss on Celo
+- Increase the ergonomics of key management on Celo
+
+## Proposed Solution
```
asaj, line 28: How will this be made backwards compatible?

CIPs/0010.md
```diff
@@ -0,0 +1,56 @@
...
+
+- The signing of the off-chain metadata should ideally not happen with the account key. During the development of the `RelaeseGold` smart contract it has been found that smart contracts as accounts cannot sign metadata. As a workaround, verifying metadata has been adjusted to allow any authorized signer to sign the payload, which violates the intention behind having well scoped authrized signers in the first place.
+- As part of PGPNP, accounts have to authenticate with the PGPNP service for rate-limiting purposes. To do that, the "account key" has to sign the request which also violates the intention of keeping keys that can move funds as cold as possible.
+- For other applications such as Profile Picture/Names or more down-the-road applications such as messaging, specific encryption keys are going to be useful. Additionally, the `dataEncryptionKey` seems out of place, despite it being effectively also an authorized key (if not signer).
```
asaj, line 19: Agree it's out of place but note that it's a public key and not an address. How does this CIP address that?

asaj:
> I like this proposal a lot! I would love to see a name-spacing scheme included as part of this proposal. Giving more flexibility to users as to how many keys they end up managing would be a great improvement to usability and security!

@nategraf I'm not sure I see how it gives that much more flexibility to users, at least as-is. Ultimately, the requirement for users will be based on how these keys are consumed, e.g. by other smart contracts. If anything, this feels like it will provide more flexibility for protocol developers, as it would allow for adding keys specific to new features

CIPs/0010.md
nambrot: As noted, this CIP is a WIP. The bullet points should signify possible additions to the proposal that could be considered. If these are unclear, i'm happy to expand

CIPs/0010.md
nambrot: This is not explicitly specified here, but the address of the `dataEncryptionKey` can be just another signer of the account?

CIPs/0010.md
nambrot: The current `validating/voting/attestation` signers would effectively get special treatment, and would not be affected, thus they would be backwards compatible?

nambrot:
I think @nategraf meant that once protocol developers implemented the spirit of this CIP, it would allow for users to be more flexible in their key management. #23 for example could be implemented with this and thus yield more flexibility for users eventually

kobigurk:
> I think @nategraf meant that once protocol developers implemented the spirit of this CIP, it would allow for users to be more flexible in their key management. #20 for example could be implemented with this and thus yield more flexibility for users eventually

#20 would definitely benefit from this. I brought up #20 as a simpler change, though if this is implemented, I believe it should satisfy all the requirements of #20.

kphed:
Moving my messages from the Discord server here, since it seems like an appropriate place to get a discussion going:

> Is anyone currently looking into alternatives methods of authorizing signers?
>
> Currently, a signer needs to generate a proof-of-possession before it can be authorized by a registered Celo account (i.e. "locked gold account") - it also cannot be re-authorized later.
>
> Based on the above, here are a few ideas that I had - would appreciate any thoughts, as well as other ideas:
> - Offer another method of verifying mutual consent when authorizing signer accounts (this would be useful in the scenario where a signer account is a smart contract)
> - Allow signer accounts to be reusable. this can be easily done by deleting the old signer account key from the `authorizedBy` mapping after authorizing a new signer account

I'd love to hear your thoughts and feedback. Also, if there isn't any opposition to the above, I can implement the above and submit a PR next week.

nategraf:
@kphed I really like the idea of allowing contracts to be signers by adding the option to prove control of an address via a transaction. I don't see any issue immediately and think it would be a great extension to the system.

On the topic of allowing signers to be reused, we currently assume that a given address is the signer for at most one Account both at any one time, and across time. An example of this is in Slashing, where a signer, which may have been deauthorized between the slash-able event and and the slashing attempt, is used to identify the validator and by extension the accused group. If the key is deleted upon being deauthorized, this lookup would not be guaranteed to work and may even result in slashing the wrong group if that key is reauthorized to a new validator (which should not happen under and normal scenario, but is an example of a correctness violation). There may also be other cases in the code that make the same historical uniqueness assumption, so we would need to go through the code base and provide solutions to any examples that come up before this could be implemented.

nambrot:
@kphed It would be great if we can amend this CIP to include generally authorizing smart contracts as signers as well!

kphed:
> @kphed I really like the idea of allowing contracts to be signers by adding the option to prove control of an address via a transaction. I don't see any issue immediately and think it would be a great extension to the system.
> 
> On the topic of allowing signers to be reused, we currently assume that a given address is the signer for at most one Account both at any one time, and across time. An example of this is in Slashing, where a signer, which may have been deauthorized between the slash-able event and and the slashing attempt, is used to identify the validator and by extension the accused group. If the key is deleted upon being deauthorized, this lookup would not be guaranteed to work and may even result in slashing the wrong group if that key is reauthorized to a new validator (which should not happen under and normal scenario, but is an example of a correctness violation). There may also be other cases in the code that make the same historical uniqueness assumption, so we would need to go through the code base and provide solutions to any examples that come up before this could be implemented.

@nategraf Thanks a lot for sharing your thoughts and clarifying the reasoning around single-use signer accounts (it clears things up a lot for me).

> @kphed It would be great if we can amend this CIP to include generally authorizing smart contracts as signers as well!

@nambrot thanks Nam, I'll hop on that now!

kphed:
Opened a [PR](https://github.com/celo-org/celo-proposals/pull/39) - if there aren't any glaring issues that you both can see, I'll dig into the implementation today and start coding tomorrow. The main challenge I can see is that there aren't any battle-tested RLP encoding library for Solidity - would appreciate thoughts there. I'll be checking out the existing libraries available (they aren't really active) and digging into the docs detailing RLP tomorrow and adding whatever we need locally.

Also, the alternative solution detailed in the proposal sounds really cool - I had a couple of ideas and will share them tomorrow (want to think them through and make sure they're not bogus first). Excited to be working on this with you all.

nambrot:
Use cases here that popped  up that would be relevant:
- ODIS authorization
- Metadata/CIP8 signer

CIPs/0010.md
```diff
@@ -0,0 +1,56 @@
+# CIP 0010: General Account Signer Authorization
+
+- Date: 2020-05-13
+- Author: @nambrot
+- Status: WIP
```
YazzyYaz, line 5: Hi @nambrot as in the other PRs, please ensure the CIP is following the new template for the newly activated CIP process here: https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

Please ensure it's in Draft status in order for me to also merge it in.

YazzyYaz (changes requested):

CIPs/0010.md
```diff
@@ -0,0 +1,84 @@
+---
+cip: 10
+title: General Account Signer Authorization
+author: Nam Chu Hoai (@nambrot), Khoa Phan (@kphed), Eko Mirhard (@eko-mirhard)
+discussions-to: https://github.com/celo-org/celo-proposals/pull/17
```
YazzyYaz, line 5: Can you create an issue on this repo for the discussions-to link? that way it doesn't get lost after we merge the PR

CIPs/0010.md
```diff
@@ -0,0 +1,84 @@
...
+- For other applications such as CIP8 or other down-the-road applications, specific keys are going to be useful. Additionally, the `dataEncryptionKey` seems out of place without useful segmentation ability, despite it being effectively also an authorized key (if not signer).
+
+Allowing the segmentation of domains for a Celo account (i.e. an account on `Accounts.sol`) gives Celo stakeholders more flexibility in their custody setup. It allows sensitive domains (like moving balances) to be only accessible by keys that are "cold", i.e. harder to compromise, while avoiding making less sensitive domains more user friendly. Expanding such segmentation aids the user experience, both of interacting with the Celo Core Contracts themselves (via authorized smart contracts), as well as dapps that currently would need to copy this authorization scheme to receive similar benefits for their users. While this CIP does not specify further applications, per-dapp-authorizations can become practical.
```
oneeman, line 39: A couple of things:
1. "while avoiding making less sensitive [...]" should be "while making less sensitive [...]"
2. Where it says "currently would need to copy this authorization scheme", it's not fully clear to me which scheme it's referring to.  Is it to the scheme already implemented through `Account`?  Perhaps a bit more background and explanation on this point (what they would have to do and why) would be helpful for people like me who don't know the ins and outs as well.  For example, I'm not clear on whether they would have to copy it to define new domains or also for making use of existing domains in their dapps.

nambrot: Thanks @oneeman for the read

For `currently would need to copy this authorization scheme`, I tried to say that without generalizing this scheme for other applications, dapps would have to copy `Accounts` to get the same benefit for their application.

CIPs/0010.md
```diff
@@ -0,0 +1,84 @@
...
+- Add a generalized `authorizeSignerWithSignature(address signer, string calldata role, uint8 v, bytes32 r, bytes32 s)` (where v, r, s is an EIP712 signature) for proof-of-possession-based authorization (TODO: EIP712 domain specification)
+
+Both of the methods above would populate the relevant `SignerAuthorization` in the `signerAuthorizations` struct after hashing the `role`, as well as the `authorizedBy` struct. These methods should continue the invariant that a signer is only authorized by a single account ever. However, we can relax the invariant that a signer can only be authorized for at most 1 role.
+
+- Add a generalized `getAuthorizedSigner(address account, string calldata role)` for returning completed authorized signers.
+
+To explicitly retain backwards compatability for existing consumers, the general write methods would special case the roles of, `celo.org/authorizationSchemas/core/vote` and `celo.org/authorizationSchemas/core/attestation`, and also populate the existing `accounts` struct. Similarly, `getAuthorizedSigner` would read from `accounts` in this case.
```
oneeman, line 60: "compatability" -> "compatibility"

CIPs/0010.md
```diff
@@ -0,0 +1,84 @@
...
+
+As mentioned above, this CIP aims to make backwards compatible changes.
+
+## Test Cases
+
+Celo Core Contract unit tests will maintain the invariants currently expected by consumers of `Accounts.sol`.
+
+## Security Considerations
+
+`Accounts.sol` is a core piece of the Celo platform and as such as wide-ranging impact on the functionality of it. `Accounts.sol` ultimately determines they keys that will be participating in the consensus protocol, eligible voters in the Proof-of-Stake and identity protocol. As mentioned above, the requirements to `Accounts.sol` are expected to remain unchanged, but failure to do so could have significant consequences upon the security of the Celo platform.
```
oneeman, line 84: "they keys" -> "the keys"

CIPs/0010.md
```diff
@@ -0,0 +1,84 @@
...
+
+The proposed changes to `Accounts.sol`:
+
+- Add a `signerAuthorizations` struct:
+    ```
+    struct SignerAuthorization {
+      address signer
+      bool authorized
+    }
+    mapping(address => mapping(bytes32 => SignerAuthoriation)) public signerAuthorizations
```
oneeman, line 51: "SignerAuthoriation" -> "SignerAuthorization"

CIPs/0010.md
```diff
@@ -0,0 +1,84 @@
...
+    ```
+- Add generalized `authorizeSigner(address signer, string calldata role)` and `completeSignerAuthorization(address account, string calldata role)` for smart contract authorizations (`authorizeSigner` creates a pending authorization that `completeSignerAuthorization` completes)
```
oneeman, line 53: I take it that the `authorized` field would be used for this, with it set to `false` until the signer calls `completeSignerAuthorization()`.  If that's so, perhaps it would be better called `completed` rather than `authorized`, aligning the field name with the name of the method that sets it to true.  Granted, there is also `authorizeSignerWithSignature()`, which doesn't mention completion, but still I would have a slight preference for `completed`.

And a question: say I authorize signer X (and the authorization is completed) for some role, and I then authorize signer Y for the same role and never complete that authorization.  The second authorization would overwrite the first, and leave that role without any authorized signer, right?  If that is so, this could be used as a hacky way to revoke an authorization without setting a new one, which raises the question whether we envision a separate function to revoke authorizations.

nambrot: Ohh bringing up revocation is a great one! I think revocation would be for sure great to support, and should be done explicitly IMO.

You scenario of an overwritten, but not completed signer is also a great point. IMO, we should allow for signers to be "updated atomically", even if completion of the authorization does not. What do you think?

oneeman: I'm not sure what you're suggesting.  The update being atomic is already the case if you use `authorizeSignerWithSignature()`, while if you use `authorizeSigner()` I'm not sure what it is that should be atomic or in what sense.

The only thing I could think of is maybe you mean for the old signer to be valid until the new signer authorization has been completed.  But I would hesitate to add any implementation or storage complexity for something like that, given that `authorizeSignerWithSignature()` is also an option for those who want 'zero-downtime' in signer authorization.

AlexBHarley: Another thought here is expanding to allow multiple signers per `role`.

To me that seems definitely what we'd want for general roles, but perhaps `Attestation` , `Vote` and `Validator` roles should be special cased to match the current implementation (ie. just one signer).

AlexBHarley:
FYI I've started exploring the changes needed for this in a branch https://github.com/celo-org/celo-monorepo/pull/5913

brad-larson:
@AlexBHarley - there is some existing work that was being done by @kphed on this branch: https://github.com/celo-org/celo-monorepo/compare/master...conclave-dev:new-signer-auth?expand=1, in case that is helpful

CIPs/0010.md
```diff
@@ -0,0 +1,84 @@
...
+- Add a generalized `authorizeSignerWithSignature(address signer, string calldata role, uint8 v, bytes32 r, bytes32 s)` (where v, r, s is an EIP712 signature) for proof-of-possession-based authorization (TODO: EIP712 domain specification)
+
+Both of the methods above would populate the relevant `SignerAuthorization` in the `signerAuthorizations` struct after hashing the `role`, as well as the `authorizedBy` struct. These methods should continue the invariant that a signer is only authorized by a single account ever. However, we can relax the invariant that a signer can only be authorized for at most 1 role.
+
+- Add a generalized `getAuthorizedSigner(address account, string calldata role)` for returning completed authorized signers.
```
nategraf, line 58: Currently, if a signer is not authorized the default is to [return the account address](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/common/Accounts.sol#L399-L408). What is the default for this updated system? Is there a way to have hierarchical keys such that if a dapp queries for `example.com/application/vote` and the key is not set but `example.com/application` is, it will return the more general key?

This could also be done by the caller by progressively querying up the path as long, provided that this function returns an indication that the key is not set. Providing native support for this would potentially make the lives of users easier, as there are already complaints about having to manage too many keys, so creating an easy way to establish a single "default signing key" and progressively additional keys as the user desires would be very helpful.

CIPs/0010.md
nategraf, line 60: Is `authorizationSchemas` included in the path to avoid collision in the hashing domain with other applications?  If so, we might want to drop that and use a constant personalization string prepended to the path when hashing. (Using a personalization string here is a likely a good idea anyway to eliminate a sharp edge for dApp developers who want to use this system.)

CIPs/0010.md
```diff
@@ -0,0 +1,84 @@
...
+    struct SignerAuthorization {
+      address signer
+      bool authorized
+    }
+    mapping(address => mapping(bytes32 => SignerAuthoriation)) public signerAuthorizations
+    ```
+- Add generalized `authorizeSigner(address signer, string calldata role)` and `completeSignerAuthorization(address account, string calldata role)` for smart contract authorizations (`authorizeSigner` creates a pending authorization that `completeSignerAuthorization` completes)
+- Add a generalized `authorizeSignerWithSignature(address signer, string calldata role, uint8 v, bytes32 r, bytes32 s)` (where v, r, s is an EIP712 signature) for proof-of-possession-based authorization (TODO: EIP712 domain specification)
+
+Both of the methods above would populate the relevant `SignerAuthorization` in the `signerAuthorizations` struct after hashing the `role`, as well as the `authorizedBy` struct. These methods should continue the invariant that a signer is only authorized by a single account ever. However, we can relax the invariant that a signer can only be authorized for at most 1 role.
```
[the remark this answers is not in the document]
nambrot: We haven't and I'd recommend to do so indeed. Though I think we should be fine (famous last words)

CIPs/0010.md
```diff
@@ -0,0 +1,84 @@
...
+
+- The signing of the off-chain data should ideally not happen with the account key. During the development of the `ReleaseGold` smart contract it has been found that smart contracts as accounts cannot sign metadata. As a workaround, verifying metadata has been adjusted to allow any authorized signer to sign the payload, which violates the intention behind having well scoped authorized signers in the first place.
+- As part of ODIS, accounts have to authenticate with the ODIS signers for rate-limiting purposes. To do that, the "account key" or the "authorized DEK" has to sign the request which also violates the intention of keeping keys that can move funds as cold as possible.
+- For other applications such as CIP8 or other down-the-road applications, specific keys are going to be useful. Additionally, the `dataEncryptionKey` seems out of place without useful segmentation ability, despite it being effectively also an authorized key (if not signer).
```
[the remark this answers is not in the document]
nambrot: I tried to not expand the scope too much but that is a fair point. A separate addrss => publicKey mapping later could do the job?

CIPs/0010.md
nambrot: I think that is really cool and sounds like something that can be done "entirely in user-land"? (though I certainly agree that codifying this here is more useful. Again, to keep the scope focused, do you think we can do this in a separate CIP?

CIPs/0010.md
nambrot: May you elaborate on this, I'm not sure I'm 100% following

CIPs/0010.md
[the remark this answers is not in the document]
nategraf: Is the assumption here that if any conflicts arise, they will be fixed in the calling contract (e.g. `Attestations`). If not, we may want to make sure we do that work now so we can make any required changes to this design if an issue is observed.

CIPs/0010.md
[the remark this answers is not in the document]
nategraf: True, I think this is a faire tradeoff.  If we want we could still start using this for the DEK, or other applications that require the public key by setting the address in this mapping and using it as a "public key commitment" for this mapping.

CIPs/0010.md
nategraf: We can do that separately. It doesn't nessasrly need to be part of any CIP, and could instead just be a convention. The key question here is: What is returned when no signer is specified for a particular path?

CIPs/0010.md
nategraf: Rephrasing: Why is the string `authorizationSchemas` included in the path? It seems clear from the context that this is the usage.

On the personalization string suggestion, it's actually not necessary.  Looking at this a second time I see now that these hashes don't need to be globally unique, as is the case with signatures, so you can ignore that.

CIPs/0010.md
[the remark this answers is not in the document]
nambrot: My current assumption is that there is no actual conflict, but as you said, we should actually look through

CIPs/0010.md
nambrot: Conventions to me are like protocols, and should be encoded as CIPs where possible.

I believe the existing return when no signer is specified  is to return the account

CIPs/0010.md
nambrot: Gotcha. it's in the string with the possible hope that some kind of structured schema can live at the actual HTTP address eventually. I assume it's at no cost.

CIPs/0010.md
nategraf: Alright. I think we should change that then. If the account is returned when no signer is specified, it will not be possible differentiate between when the signer is not specified. Instead I would recommend we return either the zero address and let the caller decide what to use as the default. (Since they already know the account address, it will be trivial for them to use it as the default, but this leaves the option for the caller to use another key path as a default instead.)

CIPs/0010.md
nategraf: Should we mention that explicitly here? I imagine more developers will start to use to use this scheme as specified here, and will be choosing their own paths, so it seems important to mention any suggested format for those paths (e.g. that they should point to a unique HTTP URL that they control) so a future application will have a chance to use them in a consistent way.

CIPs/0010.md
nambrot: That would break backwards compatibility though, right?

nategraf: I don't think so. This is a new API, so there shouldn't be any requirement to preserve the same exact behavior as the old one. If needed, the caller can default to the account address anyway.

nambrot: Good point for the new functions

YazzyYaz (approved):

YazzyYaz:
Going to merge. I realize there are a lot of discussions on the PR, and I would urge to move them to the Github discussions ticket since they can be captured better. It's best practice to minimize discussions on the actual technical implementations directly on the PR and reserve draft reviews only for whether the proposal is technically sound and related to Celo plus if it follows the template and grammar is good.

PR #18, opened 2020-05-16
Removed a legacy TODO from proposalID=2

martinvol:


CGPs/0001.md
```diff
@@ -27,7 +27,7 @@ This represents a value that will return 6% annual growth for voter rewards (as
 
 ## Verification
 
-1. Confirm proposal steps: run `celocli governance:view --proposalID 2` // TODO add proposal ID once submitted proposal
```
aslawson, line 30: nice catch!  Can you update in 0002.md as well?

PR #19, opened 2020-05-19
CIP11: Off-chain attestations and claims

nambrot:
WIP

### DIFF
--- a/CIPs/0011.md
+++ b/CIPs/0011.md
@@ -12,7 +12,7 @@ Fundamentally, attestations generally however are contextual to the attestor and
 
 Another form of an attestation is produce an account's signature and then have the attestor (such as a social media network) host it in a public way. That effectively acts as an attestation by attestor in a particular namespace (such as a profile). On Celo, an account can then make a claim to that attestation and any user can verify that attestation by checking the attestor without needing protocol-level consensus. Celo's `metadata` feature currently allows accounts to make claims over Keybase accounts or domain names. Metadata was originally intended for more narrow use cases.
 
-This CIP tries to standardize how accounts can make claims and attestations by migrating from it's original location in `metadata` to generalized off-chain storage as proposed in CIP8.
+This CIP tries to standardize how accounts can make claims and attestations by migrating from its original location in `metadata` to generalized off-chain storage as proposed in CIP8.
 
 ## Goals
 
--- a/CIPs/0011.md
+++ b/CIPs/0011.md
@@ -28,6 +28,8 @@ On top of CIP8, Attestations for an account should be made under `/accounts/${ot
   "type": "${attestationType}",
   "attestor": "${accountAddress}",
   "attestee": "${otherAccountAddress}",
+  "timestamp": "${timestamp}", // ISO8601
+  "expires_at"?: "${expires_at}", //ISO8601, optional
 }
 ```
 
@@ -60,11 +62,20 @@ For releases, attestations can be made with under `/${package}/releases/${versio
   "type": "release_signature",
   "package": "${package}",
   "version": "${version}",
-  "hash": "${hash}"
+  "hash": "${hash}",
+  "url": "${url}",
+  "timestamp": "${timestamp}" // ISO8601
 }
 ```
 
-Each attestation requires the signature under `/${dataAddress}.signature`. Attestations for an account can be indexed `/accounts/${otherAccountAddress}/attestations` if necessary.
+Each attestation requires the signature under `/${dataPath}.signature`. Attestations for an account can be indexed `/accounts/${otherAccountAddress}/attestations` if necessary:
+
+```json
+[
+  "${dataPathOfAttestation1}",
+  "${dataPathOfAttestation2}"
+]
+```
 
 Additionally, we propose to establish metadata claims under `/claims/${claimType}-${identifier}` with them indexed via `/claims`
 
@@ -89,7 +100,7 @@ Additionally, we propose to establish metadata claims under `/claims/${claimType
 }
 ```
 
-Each claim requires the signature under `/${dataAddress}.signature`
+Each claim requires the signature under `/${dataPath}.signature`
 
 ## Alternative Solutions
 
--- a/CIPs/0011.md
+++ b/CIPs/0011.md
@@ -1,0 +1,112 @@
+# CIP 11: Off-chain claims and attestations
+
+- Date: 2020-05-20
+- Author: @nambrot
+- Status: WIP
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


sharppanda:
I really like the idea of separating individual claims. Now it is very cumbersome to work with metadata. When you need to change, for example, a metadata URL, you need to recreate the whole metadata from scratch and add all other claims. 

I am also looking forward to Master Validator and other attestations, which will assure users about validator's professionalism.

CIPs/0011.md
```diff
@@ -0,0 +1,112 @@
...
+
+Another form of an attestation is produce an account's signature and then have the attestor (such as a social media network) host it in a public way. That effectively acts as an attestation by attestor in a particular namespace (such as a profile). On Celo, an account can then make a claim to that attestation and any user can verify that attestation by checking the attestor without needing protocol-level consensus. Celo's `metadata` feature currently allows accounts to make claims over Keybase accounts or domain names. Metadata was originally intended for more narrow use cases.
+
+This CIP tries to standardize how accounts can make claims and attestations by migrating from it's original location in `metadata` to generalized off-chain storage as proposed in CIP8.
```
timmoreton, line 15:
```suggestion
This CIP tries to standardize how accounts can make claims and attestations by migrating from its original location in `metadata` to generalized off-chain storage as proposed in CIP8.
```

CIPs/0011.md
```diff
@@ -0,0 +1,112 @@
...
+  "attestor": "${accountAddress}",
+  "attestee": "${otherAccountAddress}",
+}
+```
+
+We define the following attestation types:
+
+```json
+{
+  "type": "attestation_TGCSO_attestation_maven"
```
timmoreton, line 38: capitalization? case-sensitive?

CIPs/0011.md
```diff
@@ -0,0 +1,112 @@
...
+```json
+{
+  "type": "attestation_TGCSO_founder_validator",
+  "score": ""
+}
+```
+
+```json
+{
+  "type": "attestation_TGCSO_master_validator",
```
timmoreton, line 51: also: `attestation_mainnet_first_proposer_validator`, `attestation_mainnet_first_transaction_validator`

CIPs/0011.md
```diff
@@ -0,0 +1,112 @@
...
+```json
+{
+  "type": "attestation_TGCSO_attestation_maven"
+}
+```
+
+```json
+{
+  "type": "attestation_TGCSO_founder_validator",
+  "score": ""
```
timmoreton, line 45: what is `score` here?

CIPs/0011.md
```diff
@@ -0,0 +1,112 @@
...
+
+```json
+{
+  "type": "${attestationType}",
+  "attestor": "${accountAddress}",
+  "attestee": "${otherAccountAddress}",
+}
+```
+
+We define the following attestation types:
```
timmoreton, line 34: these fields extend those in the dict above?

CIPs/0011.md
```diff
@@ -0,0 +1,112 @@
...
+}
+```
+
+```json
+{
+  "type": "attestation_TGCSO_master_validator",
+  "score": ""
+}
+```
+
```
timmoreton, line 55: so for attestations, the attestation is hosted under the metadata of the attestee, and the signature is of the attestor - right?   It is interesting that it could equally be hosted under the metadata of the attestor, e.g the foundation - worth specifying how these inter-operate? 

For releases, the signature is of the party hosting the metadata?

CIPs/0011.md
```diff
@@ -0,0 +1,112 @@
...
+For releases, attestations can be made with under `/${package}/releases/${version}/attestation`:
+
+```json
+{
+  "type": "release_signature",
+  "package": "${package}",
+  "version": "${version}",
+  "hash": "${hash}"
```
timmoreton, line 63: `url`, `date` (and define format) ?

CIPs/0011.md
```diff
@@ -0,0 +1,112 @@
...
+## Proposed Solution
+
+On top of CIP8, Attestations for an account should be made under `/accounts/${otherAccountAddress}/attestations/${attestationType}`.
+
+```json
+{
+  "type": "${attestationType}",
+  "attestor": "${accountAddress}",
+  "attestee": "${otherAccountAddress}",
+}
```
timmoreton, line 31: `date` ? `expiry`? (and define format)

CIPs/0011.md
```diff
@@ -0,0 +1,112 @@
...
+```json
+{
+  "type": "release_signature",
+  "package": "${package}",
+  "version": "${version}",
+  "hash": "${hash}"
+}
+```
+
+Each attestation requires the signature under `/${dataAddress}.signature`. Attestations for an account can be indexed `/accounts/${otherAccountAddress}/attestations` if necessary.
```
timmoreton, line 67: what's the index format?

CIPs/0011.md
nambrot: Since that data already lives on-chain, it feels superflous here? Also I'm not sure that one getting lucky as the proposer should really be a worthwhile attestation?

CIPs/0011.md
nambrot: Even though I'm the naming czar I actualy do not know how abbreviations should be handled in snake case, is the correct version `attestation_TGCSO_attestation_maven` ?

CIPs/0011.md
nambrot: So I couldn't find a definition of the badges, but I thought the founder validator was the general stake off ranking?

CIPs/0011.md
nambrot: Added

CIPs/0011.md
nambrot: Yeah, I basically just wanted to express the types and any additional keys they would require

CIPs/0011.md
nambrot: So I was originally thinking that these could really be hosted at both locations indeed. With the attestee, one wouldn't have to also lookup the attestor's data. However, since that made revoking somewhat harder, in this CIP I suggest to just keep with the attestor for now. Applications could certainly decide to also trust them being stored on the attestee, but this CIP wouldn't specify that

CIPs/0011.md
```diff
@@ -0,0 +1,123 @@
+# CIP 11: Off-chain claims and attestations
+
+- Date: 2020-05-20
+- Author: @nambrot
```
YazzyYaz, line 4: Hi @nambrot as the other PRs, please ensure this CIP follows the new CIP template in the newly activated CIP process as found here https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

YazzyYaz (changes requested):

PR #27, opened 2020-06-11
Validator Hot Standby

kevjue:
WIP

### DIFF
--- a/CIPs/0015.md
+++ b/CIPs/0015.md
@@ -1,0 +1,59 @@
+# CIP [0015]: Hot standby for validators
+
+- Date: 2020-06-10
+- Author: @githubhandle
+- Status: WIP
+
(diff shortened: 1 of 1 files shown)


CIPs/0015.md
```diff
@@ -0,0 +1,59 @@
+# CIP [0015]: Hot standby for validators
+
+- Date: 2020-06-10
+- Author: @githubhandle
```
YazzyYaz, line 4: hi @kevjue  thank you for submitting this CIP, I would like you to update it to follow the newly activated CIP process which has it's newly updated template here: https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

YazzyYaz (changes requested):

PR #31, opened 2020-06-21
[GCP_0007P] Proposal for Enabling Celo Dollar transfers and activating the stability protocol

aslawson:


CGPs/0006.md
```diff
@@ -32,7 +32,7 @@ To not put too much selling pressure on CELO, the rebalancing will happen gradua
 
 ## Verification
 
-1. Confirm proposal steps: run `celocli governance:view --proposalID x` // TODO add proposal ID once submitted proposal
+1. Confirm proposal steps: run `celocli governance:view --proposalID 6`
```
rcroessmann, line 35:
```suggestion
1. Confirm proposal steps: run `celocli governance:view --proposalID 7`
```
This would be proposal ID 7 IIUC

CGPs/0007P.md
```diff
@@ -0,0 +1,61 @@
+# CGP [0007P]: Enabling Celo Dollar transfers and activating the stability protocol
+
+- Date: 2020-06-19
```
rcroessmann, line 3:
```suggestion
- Date: 2020-06-24
```

CGPs/0006.md
aslawson: This is a bit out of place, sorry for the confusion.  I'm just editing the previous proposal.

rcroessmann: Oh sure that makes sense, thanks for clarifying.

PR #39, opened 2020-07-23
Propose contract signers

kphed:
## Changes
- Fix minor grammatical & spelling errors
- Propose smart contract signers as an alternative solution in the interim

### DIFF
--- a/CIPs/0010.md
+++ b/CIPs/0010.md
@@ -39,6 +39,27 @@ Alternately, applications could always derive their own authorization scheme/sma
 
 Instead of extending the `Signers` struct, a completely different smart contract `AuthorizedSigners` could be created. Functionality would be practically duplicated, but separated from the critical PoS system while gaining the benefit from a standardized authorization scheme.
 
+Alternative Solution: Enable smart contracts to be used as signers (as an addition to existing implementation)
+
+In the current implementation, it is not possible to authorize smart contract addresses as signers, since they do not have private keys and are unable to generate the proofs required. To account for that limitation, a "smart contract signer authorization" method can be added to the `Accounts` contract with the following characteristics:
+
+Note: Each signer role will have its own variant of the solution outlined below (e.g. `authorizeAttestationSignerContract`, `authorizeVoteSignerContract`, etc. - similar to the existing implementation) - until the plan for the generalized signer authorization method is finalized.
+
+- Restricts smart contracts that are authorizable as signers to those created by `msg.sender` (for now, to limit potential damage from unforeseen malicious behavior)
+- Accepts two parameters:
+    - `contract (address)`
+    - `creationTxNonce (uint256)`
+- Performs the following preliminary checks on `msg.sender` and `contract`*
+    - `msg.sender` is a valid account
+    - `contract` is not a valid account and has never been authorized as a signer
+- Verifies that `msg.sender` is the smart contract’s creator:
+    - RLP encodes an array containing the `msg.sender` and `creationTxNonce` and hashes the result using the `keccak256` function
+    - Checks for equality between the hash and `contract`
+        - If the two values are not equal, revert the transaction
+- Assign `contract` to X role for `msg.sender` (see note above)
+
+*This logic has already been implemented in the `authorize` method (lines [562-566](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/common/Accounts.sol#L562)) but is not accessible in our case (i.e. `authorize` requires signature parameters). One solution, to avoid code duplication, is to extract that chunk of code into a reusable method that can be shared by various signer authorization methods.
+
 ## Risks
 
 Upgrading core smart contracts + RG contracts.
--- a/CIPs/0010.md
+++ b/CIPs/0010.md
@@ -1,7 +1,7 @@
 # CIP 0010: General Account Signer Authorization
 
 - Date: 2020-05-13
-- Author: @nambrot
+- Author: @nambrot @kphed @eko-mirhard
 - Status: WIP
 
 ## Overview
@@ -22,22 +22,64 @@ As applications are developed on Celo, it becomes evident that this mechanism of
 
 ## Goals
 
-- Reduce the chance of fund loss on Celo
+- Reduce the chance of funds loss on Celo
 - Increase the ergonomics of key management on Celo
+- Reduce signer authorization overhead (i.e. remove development efforts related to PoP generation, and provide a friendlier experience for users)
+- Enable smart contracts to be authorized as signers
 
 ## Proposed Solution
 
-Expand the `Signers` struct in `Accounts.sol` with a generic mapping and generic version of `authorizeXSigner`, `removeXSigner`, `getXSigner` and `signerToAccount` that take as additional input `string namespace`.
+A new signer authorization implementation that makes use of state-transitioning method calls (as opposed to signature generation) as a means of confirming mutual consent between a Celo account and a signer address. In addition to being a lightweight* alternative for signer authorization, the proposed solution will also enable the authorization of smart contracts as signers, which are unable to generate the possession proofs that are currently required (due to their lack of private keys).
 
-- Multiple account keys? (effectively making locked gold somewhat of a custody solution)
-- Authorization key
-- Makes key management easier by allowing tools implement once interface
+*Signer authorization in the proposed method will consist of only two transactions.
 
-Another solution is a signer authorization process that makes use of state-transitioning method calls (as opposed to signature generation) as a means of proving that a Celo account has consent from an address to use it as a signer. This solution would also enable the authorization of smart contracts as signers since they cannot generate the signatures required by the current implementation (as they do not have their own private keys) but can make cross-contract calls.
+The proposed solution consists of two phases (each executed via method calls): initiation and completion.
+- The "initiation" phase is carried out by a Celo account and creates a partial (i.e. pending) authorization by associating the account with the address of a signer and its type.
+- The "completion" phase is carried out by a signer by “approving” an associated partial authorization and finalizing its assignment as a signer for an account.
 
-The authorization process is comprised of two phases: initiation and completion.
-- The "initiation" phase is executed by a Celo account and results in the address of the prospective signer and their signer type being stored in preparation for authorization.
-- The "completion" phase is executed by a signer that is in the process of being authorized (i.e. "partially authorized") and, if successful, results in their assignment as an account's signer.
+### Changes
+
+Streamline the signer-related logic within the `Accounts` contract and extend its utility to cover smart contract signer authorization with the following actions:
+- Consolidate the following methods with their generic equivalents: (e.g. `XSignerToAccount` - where X is the signer type - replaced with `signerToAccountByType`**)
+     - `authorizeXSigner`
+     - `removeXSigner`
+     - `getXSigner`
+     - `removeXSigner`
+     - `XSignerToAccount`
+     - `hasAuthorizedXSigner`
+- Implement a means for authorizing signers via method calls, as opposed to signatures, to enable the usage of smart contracts as signers.
+- Update the account signer store to accommodate partial signer authorizations (utilized by the new signer authorization implementation as means of proving mutual consent).
+
+**The `signerToAccountByType` method will be created, as opposed to updating the existing `signerToAccount` method, in order to minimize the number of changes to other contracts.
+
+Update the following contracts by replacing the existing methods (listed below each contract) with their appropriate equivalent from the proposed solution:
+- ReleaseGold
+    - authorizeVoteSigner
+    - authorizeValidatorSigner
+    - authorizeValidatorSignerWithPublicKey
+    - authorizeValidatorSignerWithKeys
+    - authorizeAttestationSigner
+    - getVoteSigner
+    - getValidatorSigner
+- Attestations
+    - attestationSignerToAccount
+    - hasAuthorizedAttestationSigner
+- Validators
+    - validatorSignerToAccount
+    - getValidatorSigner
+- Election
+    - voteSignerToAccount
+- Governance
+    - voteSignerToAccount
+
+### Execution
+
+To minimize the chances of uncaught bugs or mistakes, the proposed solution will be fully implemented over a series of steps:
+1. Implement the new signer authorization procedure as a complement to the existing one. Changes will only be made to the Accounts contract.
+2. Update the signer-related methods called by other contracts in a manner that ensures a high level of operability confidence (e.g. updating, testing, and deploying contracts piecemeal, as opposed to all at once).
+3. Remove the deprecated (now unused) signer authorization logic from the `Accounts` contract.
+
+Proposed
 
 ## Alternative Solutions
 
--- a/CIPs/0010.md
+++ b/CIPs/0010.md
@@ -33,17 +33,11 @@ Expand the `Signers` struct in `Accounts.sol` with a generic mapping and generic
 - Authorization key
 - Makes key management easier by allowing tools implement once interface
 
-## Alternative Solutions
+Another solution would be to enable the usage of smart contracts as signers.
 
-Alternately, applications could always derive their own authorization scheme/smart contracts. However, it would not benefit from the tooling that could be possible for facilitating key management.
+In the current implementation, it is not possible to authorize smart contract addresses as signers, since they do not have private keys and are unable to generate the proofs required. To address that limitation, a "smart contract signer authorization" method can be added to the `Accounts` contract with the following characteristics:
 
-Instead of extending the `Signers` struct, a completely different smart contract `AuthorizedSigners` could be created. Functionality would be practically duplicated, but separated from the critical PoS system while gaining the benefit from a standardized authorization scheme.
-
-Alternative Solution: Enable smart contracts to be used as signers (as an addition to existing implementation)
-
-In the current implementation, it is not possible to authorize smart contract addresses as signers, since they do not have private keys and are unable to generate the proofs required. To account for that limitation, a "smart contract signer authorization" method can be added to the `Accounts` contract with the following characteristics:
-
-Note: Each signer role will have its own variant of the solution outlined below (e.g. `authorizeAttestationSignerContract`, `authorizeVoteSignerContract`, etc. - similar to the existing implementation) - until the plan for the generalized signer authorization method is finalized.
+Note: Each signer role will have its own variant of the solution outlined below (e.g. `authorizeAttestationSignerContract`, `authorizeVoteSignerContract`, etc. - similar to the existing implementation).
 
 - Restricts smart contracts that are authorizable as signers to those created by `msg.sender` (for now, to limit potential damage from unforeseen malicious behavior)
 - Accepts two parameters:
@@ -59,6 +53,12 @@ Note: Each signer role will have its own variant of the solution outlined below 
 - Assign `contract` to X role for `msg.sender` (see note above)
 
 *This logic has already been implemented in the `authorize` method (lines [562-566](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/common/Accounts.sol#L562)) but is not accessible in our case (i.e. `authorize` requires signature parameters). One solution, to avoid code duplication, is to extract that chunk of code into a reusable method that can be shared by various signer authorization methods.
+
+## Alternative Solutions
+
+Alternately, applications could always derive their own authorization scheme/smart contracts. However, it would not benefit from the tooling that could be possible for facilitating key management.
+
+Instead of extending the `Signers` struct, a completely different smart contract `AuthorizedSigners` could be created. Functionality would be practically duplicated, but separated from the critical PoS system while gaining the benefit from a standardized authorization scheme.
 
 ## Risks
 
--- a/CIPs/0010.md
+++ b/CIPs/0010.md
@@ -33,26 +33,11 @@ Expand the `Signers` struct in `Accounts.sol` with a generic mapping and generic
 - Authorization key
 - Makes key management easier by allowing tools implement once interface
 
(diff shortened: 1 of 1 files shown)


kphed:
Please see this commit specifically: https://github.com/celo-org/celo-proposals/pull/39/commits/e15ad94a81034c46c0a7e8ab59f4f109600f0565.

Tried being helpful with the fixes, but those changes may be distracting.

CIPs/0010.md
```diff
@@ -29,15 +29,36 @@ As applications get developed on Celo, it becomes evident that this mechanism of
...
 
 ## Alternative Solutions
 
-Alternately, applications could always derive their own authorization scheme/smart contracts. However, it would not benefit of the tooling that could be possible to faciliate key management.
+Alternately, applications could always derive their own authorization scheme/smart contracts. However, it would not benefit from the tooling that could be possible for facilitating key management.
 
-Instead of extending the `Signers` struct, a completely different smart contract `AuthorizedSigners` could be created. Functionality would be practically duplicated, but separated from the critical PoS system while gaining the benefit from a standardized authorizaton scheme.
+Instead of extending the `Signers` struct, a completely different smart contract `AuthorizedSigners` could be created. Functionality would be practically duplicated, but separated from the critical PoS system while gaining the benefit from a standardized authorization scheme.
+
+Alternative Solution: Enable smart contracts to be used as signers (as an addition to existing implementation)
```
nambrot, line 42: I probably wouldn't say that this is an alternative solution, IMO it should be part of the Proposed solution. Alternative solutions are other solutions that have been explored but are not recommended for various reasons

CIPs/0010.md
```diff
@@ -29,15 +29,36 @@ As applications get developed on Celo, it becomes evident that this mechanism of
...
+
+In the current implementation, it is not possible to authorize smart contract addresses as signers, since they do not have private keys and are unable to generate the proofs required. To account for that limitation, a "smart contract signer authorization" method can be added to the `Accounts` contract with the following characteristics:
+
+Note: Each signer role will have its own variant of the solution outlined below (e.g. `authorizeAttestationSignerContract`, `authorizeVoteSignerContract`, etc. - similar to the existing implementation) - until the plan for the generalized signer authorization method is finalized.
+
+- Restricts smart contracts that are authorizable as signers to those created by `msg.sender` (for now, to limit potential damage from unforeseen malicious behavior)
```
nambrot, line 48: I assume here that `msg.sender` is the account? What do you do if the account is a `ReleaseGold` smart contract (or any other smart contract)?

kphed: That's correct. A smart contract would receive the same treatment as a "normal" account: if it is a registered/Locked Gold Celo account then we would proceed with verifying whether it had created the signer smart contract.

CIPs/0010.md
kphed: I see, thanks for clearing that up Nam, I'll move it now.

kphed:
Finishing the first iteration tomorrow (need to clean up and optimize, and create actual signer methods), but sharing what I have so far (logic is in lace for verifying that an account created the contract that it is authorizing as a signer): https://github.com/celo-org/celo-monorepo/compare/master...conclave-dev:add-contract-signers?expand=1.

There are 3 utility methods (RLP encoding and computing unpadded bytes) that can be moved into a separate helper contract, but I was hoping to hear your thoughts first, to limit the number of contracts that may need to be deployed.

Also, Nam had told me that the updated contracts won't be deployed for ~1-2 months, but if there's a path to getting there sooner, it would greatly help our project - which aims to provide great value to the Celo community - a ton.

I'd greatly appreciate any thoughts or suggestions you both have @nambrot @nategraf.

CIPs/0010.md
nambrot: Are we sure that this aids security? IMO, this will require the account key to deploy a contract which seems very difficult to most users to do

kphed: Thanks for the thought-provoking questions @nambrot .

> Are we sure that this aids security?

Great question. After thinking about it further, since it's a common practice for contracts to create contracts and assign their ownership to others, verifying creatorship alone is probably not enough. Back to the drawing board!

> IMO, this will require the account key to deploy a contract which seems very difficult to most users to do 

Since most of the contract-deploying tx can be prepopulated (besides the account nonce) by the developer, it would be virtually zero effort for users - all that would be needed is their signature.

kphed:
Adding that the proposed solution was designed in such a way as to ensure that only a minimal amount of effort would be required to integrate and maintain the functionality for smart contracts (goal was to avoid developers and their users needing to update + upgrade contracts if there were inconsequential changes to the signer auth method - e.g. may be the case for a different solution where a state variable, used for verifying contract ownership, has its name changed).

asaj:
I think this is a great idea! IMO the account and/or the authorized signer should be able to be smart contracts. One thing that we should preserve is that *both* parties need to approve, i.e. the account needs to authorize the signer, and the signer needs to accept.

One option would look like the following:
```
enum SignerAuthorizationStatus { NONE, PROPOSED, ACCEPTED }
enum SignerType { NONE, VOTE, VALIDATOR, ATTESTATION }
struct SignerAuthorization {
  SignerAuthorizationStatus status;
  SignerType type;
}
mapping(address => mapping(address => SignerAuthorization)) partialAuthorizations;
// If status is ACCEPTED, authorizes the signer, otherwise sets status to PROPOSED.
function authorizeSigner(address signer, SignerType type);
// If status is PROPOSED, authorizes the signer, otherwise sets status to ACCEPTED.
function acceptAuthorization(address account, SignerType type);
```

@kphed what do you think?

kphed:
> ner needs to accept.

Thanks for sharing your thoughts as well as the proposed solution @asaj . Crazy timing, since my co-founder and I are literally brainstorming a similar idea - right now 😂  - yours is more concretely-formed, so we will go off of it as our base. I'll update the proposal after our call with what you shared + other additions we might have (if any).

^The interesting thing is that it could replace the current method for authorizing signers entirely.

kphed:
> > ner needs to accept.
> 
> Thanks for sharing your thoughts as well as the proposed solution @asaj . Crazy timing, since my co-founder and I are literally brainstorming a similar idea - right now 😂 - yours is more concretely-formed, so we will go off of it as our base. I'll update the proposal after our call with what you shared + other additions we might have (if any).
> 
> ^The interesting thing is that it could replace the current method for authorizing signers entirely.

Something we're thinking about is how to handle redundant attempts at authorizing a proposed signer.

asaj:
> ^The interesting thing is that it could replace the current method for authorizing signers entirely.

The nice thing about the current method is that it can be done in a single tx, but I agree we could consider deprecating it.

> Something we're thinking about is how to handle redundant attempts at authorizing a proposed signer.

What do you mean by redundant? Two calls to `authorizeSigner` with the same arguments? Seems to me like the second call should just revert

kphed:
```
// If status is ACCEPTED, authorizes the signer, otherwise sets status to PROPOSED.
function authorizeSigner(address signer, SignerType type);
// If status is PROPOSED, authorizes the signer, otherwise sets status to ACCEPTED.
function acceptAuthorization(address account, SignerType type);
```

Listing the version we were considering below, which should provide context for the question I had asked above. The collision would not occur with how you're storing proposals.

- A registered Celo account calls `proposeSignerAuth`, which adds the signer address to a `proposedSigners` mapping
    - Key = signer address
    - Value = struct containing fields for account address and role
- The signer calls `acceptSignerAuth`, which checks the `proposedSigners` mapping and, if it exists, adds it as the signer for the account under the role

Going to finish our call, update the proposal, and tag you guys right after.

kphed:
```
enum SignerType { VOTE, VALIDATOR, ATTESTATION }
mapping(SignerType => string) signerTypes;
mapping(address => mapping(SignerType => address)) partialAuthorizations;
// Called by Celo accounts
function initiateSignerAuthorization(address signer, SignerType type);
// Called by signers
function completeSignerAuthorization(address account, SignerType type);
```

Changes
- Removed `enum SignerAuthorizationStatus...` since the signer should be considered as "accepted" if they're in [authorizedBy](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/common/Accounts.sol#L49)
- Removed `struct SignerAuthorization...` since we no longer need to store the authorization status
- Added `mapping(SignerType => string) signerTypes;` which will map `SignerType`s integers to their string equivalent (for setting the signer - detailed below)
- Map `SignerType` to `address` to: 
    - Avoid state bloat from multiple partially-authorized signers
    - Prevent the scenario where a compromised, partially-authorized "honeypot" signer address is later fully-authorized for an account when they least suspect it
- Changed method names to make it clearer that it's a two-step process:
    - `authorizeSigner` changed to `initiateSignerAuthorization`
    - `acceptAuthorization` changed to `completeSignerAuthorization`

Functionality
- `initiateSignerAuthorization(address signer, SignerType type)`
    - Validates the accounts
        - Checks that `msg.sender` is a registered Celo account
        - Checks that `signer` is not in the `authorizedBy` mapping (authorized signers) and is not a registered Celo account
    - Accesses the value at `partialAuthorizations[msg.sender]` and updates it by mapping `type` to `signer`
        - By not checking whether a value at key `signer` exists, we are allowing `type` to be changed for a signer that is pending authorization (tx would have reverted if they were already authorized)
- `completeSignerAuthorization(address account, SignerType type)`
    - Validates the accounts
        - Checks that `account` is a registered Celo account
        - Checks that `msg.sender` is not in the `authorizedBy` mapping (authorized signers) and is not a registered Celo account
    - Checks whether the value at `partialAuthorizations[account][type]` is equal to `msg.sender`
        - If equal
            - Set `accounts[account].signers[signerTypes[type]]` to `msg.sender` (i.e. setting it as the account's signer for the type)
            - Set `authorizedBy[msg.sender]` to `account`
        - Else revert
    - Delete the value at `partialAuthorizations[account][signerTypes[type]]`

Adding this comment since I won't be including the code snippet in the proposal text doc (updating it shortly with write-up). What do you think of the implementation details? @asaj @nambrot

nambrot:
If we will already add code and storage to support 2-step authorizations, can we do it in a way that makes progress torwards the original goal of CIP10 to generalize signers beyond voting/validating/attesting?

kphed:
> If we will already add code and storage to support 2-step authorizations, can we do it in a way that makes progress torwards the original goal of CIP10 to generalize signers beyond voting/validating/attesting?

Great point - thinking about it, we're 90% of the way there with what we've discussed so far in this PR. I'll consolidate the new logic and the existing logic now and update the proposal accordingly.

@nambrot @asaj  is it possible to roll out the new implementation without replacing the existing one (for now), to minimize the number of contract upgrades that would be needed? Since the ability to authorize contracts as signers sooner would be really amazing. Since the final version of this proposal will cause updates in other protocol contracts (e.g. change `voteSignerToAccount` to `signerToAccount` in Election.sol), it would take longer to go live.

nambrot:
Yeah I don't think we'd want to touch the existing public interfaces, and leave them as is as they are very important. I think it's ok to do the generalized public interfaces later

kphed:
> Yeah I don't think we'd want to touch the existing public interfaces, and leave them as is as they are very important. I think it's ok to do the generalized public interfaces later

Sounds good Nam, thanks for confirming. I'll update the proposal with "implementation phases" and see what you think.

kphed:
Good to go - would love to hear thoughts that you all might have before coding @nambrot @asaj !

kphed:
> Good to go - would love to hear thoughts that you all might have before coding @nambrot @asaj !

Actually, already started implementing what we have all discussed so far, which should address what you mentioned this morning @nambrot (should be done tomorrow).

kphed:
Sorry about the delay, last week was crazier than expected.

[Here](https://github.com/celo-org/celo-monorepo/compare/master...conclave-dev:new-signer-auth?expand=1) is the new signer authorization implementation 🚀  (I'll be refactoring code, writing tests, and submitting the PR on Monday). It's backward compatible but does not have the generic methods yet, as they would require changes in other contracts.

Presuming all requested changes are addressed, would it be possible for the new implementation to go live on mainnet within 1-2 months? It would be a boon for Celo developers, like ourselves, as it would increase the safety of our users considerably (an understatement). I'd sincerely appreciate your candid thoughts on the matter @nambrot @asaj .

nambrot:
If I understand correctly, you have chosen to create a separate `signerAuthorizations` struct in addition to the `authorizedBy` mapping that you hope to deprecate. To support the generalization, I would like to see it move beyond the `enum`. I.e. what do you think about the "SignerRole" key  to be a string?

kphed:
> If I understand correctly, you have chosen to create a separate `signerAuthorizations` struct in addition to the `authorizedBy` mapping that you hope to deprecate. To support the generalization, I would like to see it move beyond the `enum`. I.e. what do you think about the "SignerRole" key to be a string?

That's right, Nam. Here is what I have in mind for `signerAuthorizations`:
- Remove `authorizedBy` (replaced by `signerAuthorizations`)
- Remove `authorizeXSigner` methods (new implementation handles authorization)
- Replace remaining signer methods with generic versions. Here's an example of `signerToAccount` (replaces `XSignerToAccount`):
```
function signerToAccount(address signer, SignerRole role)
        external
        view
        returns (address)
    {
        SignerAuthorization memory signerAuth = signerAuthorizations[signer];
        require(signerAuth.authorized, "Signer has not been authorized");
        require(
            signerAuth.role == role,
            "Signer was not authorized for the specified role"
        );

        return signerAuth.account;
    }
```
- Remove `Signers` struct
- Change `Account.signers` to `mapping(uint => address)` (uint-casted `SignerRole` - enum can't be mapping key type). Here's an example of `getSigner` (replaces `getXSigner` methods):
```
function getSigner(address account, SignerRole role)
        public
        view
        returns (address)
    {
        require(isAccount(account), "Unknown account");
        return accounts[account].accountSigners[uint256(role)];
    }
```

> To support the generalization, I would like to see it move beyond the enum. I.e. what do you think about the "SignerRole" key to be a string?

To clarify, when you say "generalization" do you mean that arbitrary signer names should be supported, Nam? I'm asking since "signer callable" methods such as `vote` verify signer permissions by specifying their role. If that's unlikely to change in the medium-term, then an `enum` may be the most succinct option, since struct fields have to be explicitly accessed with dot notation. In the `getSigner` function above, if we changed the `SignerRole role` param to `string role`, we may have to do something like this:
```
function getSigner(address account, string memory role)
        public
        view
        returns (address)
    {
        require(isAccount(account), "Unknown account");
        Signers memory signers = accounts[account].signers;

        // Cannot compare strings for equality and must use their hashes
        if (keccak256(abi.encodePacked((role))) == keccak256(abi.encodePacked(("vote")))) {
            require(signers.vote != address(0), "No account signer for role");
            return signers.vote;
        }

        if (keccak256(abi.encodePacked((role))) == keccak256(abi.encodePacked(("validator")))) {
            require(signers.validator != address(0), "No account signer for role");
            return signers.validator;
        }

        if (keccak256(abi.encodePacked((role))) == keccak256(abi.encodePacked(("attestation")))) {
            require(signers.attestation != address(0), "No account signer for role");
            return signers.attestation;
        }
    }
```

Here's an idea for achieving generalization: swap out role-based permissioning for a method-based one. For example, an account may authorize a signer to only call the `vote` method. The `vote` method would later look up whether the signer has permission to call it.

Achieving true composability for signer keys would be really cool and I'm committed to working with you on it but, since it would take time, I think it would be a great idea to enable smart contract signers first as it would likely provide immediate value.

kphed:
> ... I'll be refactoring code, writing tests, and submitting the PR on Monday...

Holding off on submitting the PR until after we decide on a direction. For now, going to refocus on our project.

nambrot:
What I am proposing is the following:

Add a general mapping like your `signerAuthorizations` + (plus a way to track of pending authorizations):
`mapping(address => (mapping(bytes32) => address))`

The first address is the account, the bytes32 is the hash of the role and the last address is the signer

Existing external API remains the same:
- `authorizeXSigner`
- `getXSigner`

Implementation wise, it would continue to write to the old `signers` struct, but also into the `signerAuthorizations`.

This allows us to add generalized authorization functions (now or later) a la `authorizeGeneralSigner(address signer, bytes32 role)`. The general function would ensure that the role is not matching any of our special roles (vote, validate, attestation), but otherwise just adds it to `signerAuthorizations`. 

What do you think?

kphed:
Thanks for elaborating, Nam. I'm all for laying out the groundwork for achieving the proposal's goal. My thoughts below regarding implementation and why/how the role-specific methods can be replaced by their generic counterparts:

> Add a general mapping like your signerAuthorizations + (plus a way to track of pending authorizations):
mapping(address => (mapping(bytes32) => address))
> 
> The first address is the account, the bytes32 is the hash of the role and the last address is the signer

Can you share the rationale behind the first address being the authorizing account? Unless there's another purpose for `authorizedBy`, `signerAuthorizations` would eventually replace it for signer-to-account lookups (mentioned [here](https://github.com/celo-org/celo-proposals/pull/39#issuecomment-668053041) for planned changes). 

Incomplete/uninitiated authorizations are currently tracked using the `SignerAuthorization` struct's `authorized` field (default value is `false` and is changed to `true` only after a signer successfully calls `completeSignerAuthorization`). 

I'll replace `enum` with `bytes32` to enable arbitrarily-named signer roles. For the sake of development UX (for new devs), I'm leaning towards `role` as a function param can be type `string` but stored as type `bytes32`.

> Implementation wise, it would continue to write to the old signers struct, but also into the signerAuthorizations.
>
> This allows us to add generalized authorization functions (now or later) a la authorizeGeneralSigner(address signer, bytes32 role). The general function would ensure that the role is not matching any of our special roles (vote, validate, attestation), but otherwise just adds it to signerAuthorizations.

Is there a reason for why the generic methods shouldn't replace the role-specific ones? Special role names can be preserved by hardcoding the role name as an argument when calling generic methods (example using `vote` below). However, this would require swapping out the `Signers` struct with `mapping(bytes32 => address)` as the type for `signers` (since struct fields can only be accessed by explicitly specifying the field name w/ dot notation). 

For example, currently, `vote` calls `voteSignerToAccount` to implicitly authenticate the caller by retrieving the account for the "vote" signer role: `address account = getAccounts().voteSignerToAccount(msg.sender);` (line [247](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/governance/Election.sol#L247)).

Here's how it would look with the generic method: `address account = getAccounts().genericSignerToAccount(msg.sender, "vote");`

`genericSignerToAccount`
```
function genericSignerToAccount(address signer, string calldata role)
        external
        view
        returns (address)
    {
        SignerAuthorization memory signerAuth = signerAuthorizations[signer];
        if (signerAuth.account != address(0)) {
            require(
                accounts[signerAuth.account].accountSigners[keccak256(
                    abi.encodePacked(role)
                )] == signer,
                "not active authorized vote signer"
            );
            return signerAuth.account;
        } else {
            // require(isAccount(signer), "not an account");
            return signer;
        }
    }
```

What're your thoughts?

kphed:
^Update: Important to note that certain changes mentioned below would be made over time. Hopefully, this paints a more vivid picture of what the final version could look like.

First set of changes would be an updated version of [this](https://github.com/celo-org/celo-monorepo/compare/master...conclave-dev:new-signer-auth?expand=1) with changes from our discussion.

nambrot:
> Can you share the rationale behind the first address being the authorizing account? Unless there's another purpose for authorizedBy, signerAuthorizations would eventually replace it for signer-to-account lookups (mentioned here for planned changes).

You'll always need a mapping both ways, right? Not a particularly strong opinion how they are actually named, only preference towards changing less if possible. It seems to me that `authorizedBy` has to remain. It looks like I had a slight confusion in naming, I treated `signerAuthorizations` as your `accountSigners`.

> I'll replace enum with bytes32 to enable arbitrarily-named signer roles. For the sake of development UX (for new devs), I'm leaning towards role as a function param can be type string but stored as type bytes32.

You can take a look at the `Registry.sol`, we provide both methods, i.e. string + bytes32 (latter being cheaper)

> Is there a reason why the generic methods shouldn't replace the role-specific ones?

You are right that eventually we would just move to the generic implementations everywhere, but until then we should maintain the specific implementations to avoid having to change other contracts

brad-larson:
Just adding my two cents here, but for me it would be preferable to allow a single multisig contract to be an authorized signer for multiple accounts. 

Having to deploy a unique multisig contract for each account, but having the same set of owners for each of those multisig contracts doesn't really increase my security, it just increases the complexity I have to manage. 

I understand that this is likely a minority use case, and that this might introduce additional complexity, but just wanted to register my thoughts/feature request.

Thanks for all the hard work on this!

asaj:
@brad-larson yeah I agree that this would be a nice feature. Unfortunately, right now it's a bit tricky because the account address isn't passed in as an argument to most function calls. Instead, the account address is deduced from the signer address, which would be impossible with a one-to-many mapping.

I think your suggestion fits in with the general pattern of "fewer restrictions on signers", which @kphed's PR is advancing, but I think it's probably a bit further down the line. There are other things that can be done in the meantime, such as allowing signers to be re-used, and perhaps allowing for a more general authorization framework (rather than just voter, validator, and attestations).

CIPs/0010.md
```diff
@@ -1,43 +1,91 @@
 # CIP 0010: General Account Signer Authorization
 
 - Date: 2020-05-13
-- Author: @nambrot
+- Author: @nambrot @kphed @eko-mirhard
```
YazzyYaz, line 4: hi @nambrot @kphed  thank you for submitting this CIP, I would like you to update it to follow the newly activated CIP process which has it's newly updated template here: https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

YazzyYaz (changes requested):

nambrot:
I made some modifications to this proposed CIP change and pushed it to https://github.com/celo-org/celo-proposals/pull/17, so I would propose for the discussion to continue there

PR #40, opened 2020-07-27
[CIP 0016] Universal celo protocol URI scheme for QRs/deep links

yorhodes:
see `./CIPs/0016.md`


CIPs/0016.md
```diff
@@ -0,0 +1,89 @@
...
+
+`prefix` is used for differentiating between user and contract `target`s and actions. `pay` and `request` are for user targets and intended for use with phone numbers or addresses and the `token` endpoint. `call` & `send` are for contract targets and intended for use with the registry/addresses and the `function` endpoint. 
+
+`token` is provided as an endpoint to more concisely encode celo protocol-aware ERC20 transfers that would otherwise be represented via a `function` endpoint.
+
+`registry_id` is provided to simplify and shorten celo protocol-aware contract interactions (to be eventually translated by the application to a registry-specified 40 byte address). 
+
+`function_args` have an optional `TYPE` because, in some cases, the application may have knowledge of the ABIs deeming the solidity types unnecessary. These can most likely be omitted when used with a celo registry contract via a `registry_id` target.
+
+`localAmount` is intended for use with `localCurrencyCode` and specifies a value in a locale-specific denomination (to be eventually translated by the application to a celo-specific denomination). Resolving `value` with `localAmount` is discretionary.
```
asaj, line 56: Curious about this decision, can you elaborate on why `localAmount`/`localCurrencyCode` were chosen rather than translating to a `value` client side and encoding that in the QR code instead?

CIPs/0016.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Risks
+
+- Confusion between extensive set of endpoint parameters.
+- Ease of use for phone number targets.
+- Deprecated registry IDs or phone number verifications.
+
+## Useful Links
+
+Adapted from [ERC 67](https://github.com/ethereum/EIPs/issues/67), [EIP 681](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-681.md0), and [EIP 831](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-831.md) to celo protocol needs and users
```
asaj, line 80: nit: md0->md, capitalize Celo here and everywhere

CIPs/0016.md
```diff
@@ -0,0 +1,89 @@
...
+
+###  Rationale
+
+Various QR schemes are used across implementations such as Valora, ContractKit, and likely others.
+
+Expanding the `target` spec to be inclusive of phone numbers and contract registry identifiers makes statically generated QR codes less likely to expire than if an address (which can be changed or compromised) is used. 
+
+For low-end mobile devices, QR codes are primarily bottlenecked by pixel density and image quality (Reed-Solomon error-correction). This proposal optimizes for these parameters and enables short URIs which can be deep-linked to application functionality.  
+
+- Valora example: `celo:pay-442071838750/cUSD` (26 chars)
```
nambrot, line 66: Nit, but if this is a phone number, shouldn't it be preceeded by a `+`?

CIPs/0016.md
```diff
@@ -0,0 +1,89 @@
...
+
+`function_args` have an optional `TYPE` because, in some cases, the application may have knowledge of the ABIs deeming the solidity types unnecessary. These can most likely be omitted when used with a celo registry contract via a `registry_id` target.
+
+`localAmount` is intended for use with `localCurrencyCode` and specifies a value in a locale-specific denomination (to be eventually translated by the application to a celo-specific denomination). Resolving `value` with `localAmount` is discretionary.
+
+###  Rationale
+
+Various QR schemes are used across implementations such as Valora, ContractKit, and likely others.
+
+Expanding the `target` spec to be inclusive of phone numbers and contract registry identifiers makes statically generated QR codes less likely to expire than if an address (which can be changed or compromised) is used.
```
nambrot, line 62: This seems dangerous. What's the risk of a phone number not being mapped uniquely or the user not having enough quota to request a pepper vs. the risk of an address being expired?

CIPs/0016.md
```diff
@@ -0,0 +1,89 @@
...
+    endpoint                = token / function
+    token                   = "cUSD" / "CELO" / ...
+    function                = function_name "(" [ function_args ] ")"
+    function_name           = STRING
+    function_args           = function_arg *( "," function_arg )
+    function_arg            = ( NUMBER / STRING ) [ ":" TYPE ]
+
+    parameters              = parameter *( "&" parameter )
+    parameter               = key "=" value
+    key                     = "value" / "gas" / "gasPrice" / "feeCurrency" / "gatewayFee" / "gatewayFeeRecipient" / "localAmount" / "localCurrencyCode" / ...
```
nambrot, line 39: It would be nice to have an exhaustive list of keys that are valid and reject all keys that do not match

CIPs/0016.md
yorhodes: The constructor may not have access to the relevant exchange rate APIs and/or a merchant may want to denominate an item's price in a local denomination without being subject to exchange rate fluctuations.

CIPs/0016.md
```diff
@@ -0,0 +1,89 @@
...
+    value                   = NUMBER / STRING / celo_address
+
+### Semantics
+
+`TYPE` is a standard ABI type name, as defined in [Ethereum Contract ABI specification](https://solidity.readthedocs.io/en/develop/abi-spec.html). `STRING` is a URL-encoded unicode string of arbitrary length, where delimiters and the percentage symbol (`%`) are mandatorily hex-encoded with a `%` prefix.
+
+`E164_PHONE_NUMBER` is a standard E164 universal phone number, as defined by the International Telecommunication Union [E164 Number Specification](https://www.itu.int/rec/T-REC-E.164).
+
+`prefix` is used for differentiating between user and contract `target`s and actions. `pay` and `request` are for user targets and intended for use with phone numbers or addresses and the `token` endpoint. `call` & `send` are for contract targets and intended for use with the registry/addresses and the `function` endpoint.
```
cmcewen, line 48: So just so I understand this - `call` and `send` represent view calls vs sending a transaction on a contract? Also, I think there's a little bit of weirdness around `pay` and `request` - pay directly maps to sending a transaction whereas request is an application level construct that is stored off chain. I wonder if we should differentiate that somehow

CIPs/0016.md
```diff
@@ -0,0 +1,89 @@
...
+
+`prefix` is used for differentiating between user and contract `target`s and actions. `pay` and `request` are for user targets and intended for use with phone numbers or addresses and the `token` endpoint. `call` & `send` are for contract targets and intended for use with the registry/addresses and the `function` endpoint. 
+
+`token` is provided as an endpoint to more concisely encode Celo protocol-aware ERC20 transfers that would otherwise be represented via a `function` endpoint.
+
+`registry_id` is provided to simplify and shorten Celo protocol-aware contract interactions (to be eventually translated by the application to a registry-specified 40 byte address). 
+
+`function_args` have an optional `TYPE` because, in some cases, the application may have knowledge of the ABIs deeming the solidity types unnecessary. These can most likely be omitted when used with a Celo registry contract via a `registry_id` target.
+
+`localAmount` is intended for use with `localCurrencyCode` and specifies a value in a locale-specific denomination (to be eventually translated by the application to a Celo-specific denomination). Resolving `value` with `localAmount` is discretionary.
```
cmcewen, line 56: I get the rationale here but this feels like it could be brittle. Don't have any better suggestions though 😄

CIPs/0016.md
cmcewen: I think the other thing here is that this requires internet access to resolve, which hampers any sort of offline mode a little bit. If you're in a basement with your friend for dinner and want to send a payment that gets queued for later, this makes it difficult. Or sending via mesh networking, etc

CIPs/0016.md
```diff
@@ -0,0 +1,89 @@
...
+    key                     = "value" / "gas" / "gasPrice" / "feeCurrency" / "gatewayFee" / "gatewayFeeRecipient" / "localAmount" / "localCurrencyCode" / ...
+    value                   = NUMBER / STRING / celo_address
+
+### Semantics
+
+`TYPE` is a standard ABI type name, as defined in [Ethereum Contract ABI specification](https://solidity.readthedocs.io/en/develop/abi-spec.html). `STRING` is a URL-encoded unicode string of arbitrary length, where delimiters and the percentage symbol (`%`) are mandatorily hex-encoded with a `%` prefix.
+
+`E164_PHONE_NUMBER` is a standard E164 universal phone number, as defined by the International Telecommunication Union in [E164 Number Specification](https://www.itu.int/rec/T-REC-E.164).
+
+`prefix` is used for differentiating between user and contract `target`s and actions. `pay` and `request` are for user targets and intended for use with phone numbers or addresses and the `token` endpoint. `call` & `send` are for contract targets and intended for use with the registry/addresses and the `function` endpoint.
```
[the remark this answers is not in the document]
detach8: Just an observer's comment, not sure if this makes sense:

Is the use of a dash `-` separator for `prefix` deliberate? It seems to deviate from a typical URI. Would sticking to a slash `/` make more sense, e.g. `celo:prefix/target[/endpoint]`

This would make `prefix` mandatory, and could possibly result in more specific spec for each type of `prefix`.

CIPs/0016.md
yorhodes: yeah it's apparent that the prefix system doesn't make much sense as it's mixing a few different categories
I figured these would be nice to haves for deep-linking to specific application state

CIPs/0016.md
yorhodes: A transaction needs to be resolved online either way so I don't exactly see how an additional online step makes this prohibitive
"queued for later" could just include the lookup no?
I do recognize that including phone numbers here comes with a unique set of problems but imo this is one of the unique value props of Celo and app devs should be able to tap into it
Maybe it shouldn't be as "first-class" as it is here

CIPs/0016.md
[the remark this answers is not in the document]
nambrot: It similar to @carterqw2 's feedback, and keeping a more classic URI, especially one that is similar to existing deeplinks would be a good idea

CIPs/0016.md
```diff
@@ -0,0 +1,84 @@
...
+
+`cUSD` and `CELO` are provided as a concise encoding of native ERC20 transfers that would otherwise be represented via a contract `registry_id`, `function_name`, and recipient `celo_address`. These optionally take 1 `function_arg` which is an amount in the token denomination.
+
+`function_args` have an optional `TYPE` because, in some cases, the application may have knowledge of the ABIs deeming the solidity types unnecessary. These can most likely be omitted when used with a Celo registry contract via a `registry_id` target.
+
+`localValue` is intended for use with `localCurrencyCode` and specifies a value in a locale-specific denomination (to be eventually translated by the application to a Celo-specific denomination). Resolving `value` with `localValue` is discretionary.
+
+All `key` values should be acommodated in abbreviations as the first character from each camelcase section. As an example, `value`, `localCurrency` and `gatewayFeeRecipient` could be provided as `v`, `lc` and `gfr`.  
+
```
thelostone-mc, line 54: Hey @yorhodes ! 
From the gitcoin team here, just had a question regarding the format 

If I wanted to send :
**from**: 123
**to**: 456
**amount**: 10 cUSD

The qr string would look like : `celo://456/cUSD?v=10`
Would I be right in that ?

Also I see `from` as a parameter. I assume this is a from celo address but not super clear on how would happen if someone who isn't the owner scans the code

PR #41, opened 2020-07-27
[CIP 0017] Integrate plumo proofs into the lightest sync client

lucasege:
WIP

### DIFF
--- a/CIPs/0017.md
+++ b/CIPs/0017.md
@@ -35,13 +35,13 @@ GetPlumoProofMsg - Request plumo proof(s) from a peer
 
 The `les` protocol must then also be extended via at least two messages:
 ```
-GetEpoch - Sends the current highest epoch known to the sender. Server responds with the minimal set of proofs & headers to give the light client the fastest sync time.
-Epochs - Receives a mix of header(s) and proof(s) from the server.
+GetEpochsMsg - Sends the current highest epoch known to the sender. Server responds with the minimal set of proofs & headers to give the light client the fastest sync time.
+EpochsMsg - Receives a mix of header(s) and proof(s) from the server.
 ```
 
-The `les` client must then be modified to send a `GetEpoch` message during `lightest` sync. This will likely replace the function `getEpochOrNormalHeaders` within `eth/downloader`.
+The `les` client must then be modified to send a `GetEpochsMsg` message during `lightest` sync. This will likely replace the function `getEpochOrNormalHeaders` within `eth/downloader`.
 
-Upon receiving the `Epoch` response message, the light client must then verify headers and proofs provided. Proofs will be verified via a call into `celo-bls-go`'s `verify` fn.
+Upon receiving the `EpochsMsg` response message, the light client must then verify headers and proofs provided. Proofs will be verified via a call into `celo-bls-go`'s `verify` fn.
 
 The verify function requires information on an epoch's validator set, so this will likely require some access changes (private -> public) and exposing more data structures to allow the validator set to be queried and modified upon proof verification.
 
--- a/CIPs/0017.md
+++ b/CIPs/0017.md
@@ -37,9 +37,15 @@ Specify one or more new plumo proofs which have appeared on the network.
 This is intended to be honored with a proceeding GetPlumoProofsMsg to request the proof.
 
 GetPlumoProofsMsg (0x1a)
-[[firstEpoch, lastEpoch, versionNumber], ...]
-Sent from node A to node B with the set of proofs that node A currently has, requires node B to return a PlumoProofsMsg.
-Node B's reply must contain any proofs that node B has that node A did not send.
+[complement: Boolean, [[firstEpoch, lastEpoch, versionNumber], ...]]
+Sent from node A to node B: if `complement` is true, then this message is sent with the set of proofs
+that node A currently *has* and requests those it is missing.
+Node B's reply must contain any proofs that node B has that were not included in the message.
+If `complement` is false, then this message is sent with the set of proofs that node A currently
+*does not have* and requests these from this peer.
+Node B's reply must contain the proofs requested, if node B has them. Node A should only request proofs
+that node B has indicated via `NewPlumoProof` that it has.
+
 
 PlumoProofMsg (0x1b)
 [plumoProof_0, plumoProof_1, ...]
@@ -60,14 +66,11 @@ Upon receiving the `CertifiedEpochsValidatorSet` response message, the light cli
 The verify function requires information on an epoch's validator set, so this will likely require some access changes (private -> public) and exposing more data structures to allow the validator set to be queried and modified upon proof verification.
 
 ### Upgradeability
-In order to support an upgradeable snark circuit, we utilize a `versionNumber` parameter with every proof. This corresponds to a version of the snark circuit and thus a corresponding verification key. These version numbers will be stored on chain, within the `blockchain_parameters` smart contract, with the corresponding hash of the verification key. The clients are expected to store all applicable verification keys.
+In order to support an upgradeable snark circuit, we utilize a `versionNumber` parameter with every proof. This corresponds to a version of the snark circuit and thus a corresponding verification key. The clients are expected to be shipped with all available verification keys, and can be upgraded to receive added verification keys.
 
 For any uses of the snark's `verify` fn in the above protocols, a client is expected to:
-* Grab the verification key corresponding to the proof's `versionNumber`
-* Hash this key and compare the result to `blockchain_parameters` published VK hash.
-* If equal, the client may proceed to verify the proof.
-
-This allows the snark circuit to be upgraded via incrementing future proofs' `versionNumber` and publishing a new mapping of `versionNumber: VK_hash`, as well as distributing the new VK.
+* Fetch the verification key corresponding to the proof's `versionNumber` from its storage
+* The client then uses that verification key to call `verify` on the proof.
 
 ### Bad proofs
 Bad proofs should be prevented from propagated throughout the network via the `verify` check that each node performs. In the case that any participant receives a proof that does not pass verification for its inputs, that participant is expected to drop the sender. 
--- a/CIPs/0017.md
+++ b/CIPs/0017.md
@@ -21,6 +21,7 @@ More information on Plumo can be found in the [reference paper](https://docs.zkp
 
 ## Proposed Solution
 
+### Full node changes (RPC and `eth` protocol)
 We propose to add two new RPCs: 
 * `AddProof(proof []byte, firstEpoch uint, lastEpoch uint, versionNumber uint)`
 * `Proofs() ([][]byte)`.
@@ -55,17 +56,78 @@ Reply to GetPlumoProofMsg. The items in the list are plumo proof structs, {proof
 This may validly contain no plumo proofs if the GetPlumoProofMsg included all proofs this node is aware of.
 ```
 
-This will likely be accompanied by a new `proofFetcher` object to be in charge of fetching these proofs from peers while providing protections against DoS attacks. This feels like overkill, but seems necessary to prevent attacks and maintain state.
+This will be accompanied by a new `proofFetcher` object to be in charge of fetching these proofs from peers while providing protections against DoS attacks. This feels like overkill, but seems necessary to prevent attacks and maintain state.
 
-The `les` protocol must then also be extended via at least two messages:
+### `les` protocol changes
+
+Relevant proposed structs:
 ```
-GetCertifiedEpochsValidatorSet - Sends the current highest epoch known to the server. Server responds with the minimal set of proofs & headers to give the light client the fastest sync time.
-CertifiedEpochsValidatorSet - Receives a mix of header(s) and proof(s) from the server.
+// Same as in the `eth` protocol, minimal proof metadata
+ProofMetadata struct {
+  FirstEpoch uint
+  LastEpoch uint
+  VersionNumber uint
+}
+
+// Metadata for each epoch to construct `EpochBlock` for the snark's verify fn
+// May be unnecessary if we can make assumptions on `MaxNonSigners`
+LightEpochBlock struct {
+  Index         uint // Epoch Index 
+  MaxNonSigners uint // MinQuorumSize()
+}
+
+// All of the data needed by the light client to verify and utilize a proof.
+// This is also crafted to allow for batch verifications using `FirstHashToField`
+// My rough calculations show this struct is ~439 bytes for a non-batched proof with no val set differences.
+LightPlumoProof struct {
+  Proof             []byte // Serialized proof. Latest estimate from @psi = 383 bytes
+  FirstEpoch        LightEpochBlock // First epoch this proof is over
+  LastEpoch         LightEpochBlock // Last epoch this proof is over
+  VersionNumber     uint // Proof version number, indicates the corresponding verification key to use
+  LastHashToField  []byte  // Hash to field of Last epoch's `snark.EpochBlock` struct. To be used for batch verification.
+                           // If there is no batch verification, this will be empty. May need an additional boolean to signify this.
+                           // If batch verification should be used, newValidators and deletedValidators will be empty.
+  NewValidators     []blsCrypto.SerializedPublicKey // ValSetDiff(FirstEpoch, LastEpoch), added validators
+  DeletedValidators *big.Int // ValSetDiff(FirstEpoch, LastEpoch): bitmap of deleted validators from the `firstEpoch` valset.
+}
 ```
 
-The `les` client must then be modified to send a `GetCertifiedEpochsValidatorSet` message during `lightest` sync. This will likely replace the function `getEpochOrNormalHeaders` within `eth/downloader`.
+The light clients will need to store verification keys corresponding to the existing versions of the snark circuit, as elaborated in #upgradeability. 
 
-Upon receiving the `CertifiedEpochsValidatorSet` response message, the light client must then verify headers and proofs provided. Proofs will be verified via a call into `celo-bls-go`'s `verify` fn.
+There are two proposed extensions for the `les` protocol:
+
+Option 1:
+This option is more resilient to incomplete proof propagation for full nodes, but requires an extra round trip of messages to achieve this. The idea is for the client to query the server to see what proofs are available, and then request the optimal proofs. If the client wants to minimize network traffic and assume proofs are well distributed, they can also bypass the `GetAvailableProofs` step and just request that the server sends the best proofs it has for the client's current epoch (using the `handleLogic` flag on `GetEpochHeadersOrProofs`).
+```
+GetAvailableProofs - Asks the server what proofs it has. Server responds with `[]proofMetadata` of all available proofs.
+GetEpochHeadersOrProofs:
+  Client sends (highestEpoch uint, handleLogic boolean, requestedProofs []proofMetadata)
+    `highestEpoch` is the current highestEpoch the client knows of.
+    `handleLogic`: If true, indicates the server should send back the best set of proofs/headers it has
+                   If false, the server should look at `requestedProofs` and send back proofs requested, with headers to fill gaps as appropriate.
+    `requestedProofs`: Set of proofs' metadata requested by the client
+  Server responds with `EpocHeadersOrProofs` dependent on `handleLogic`.
+  Both responses are subject to message size limits limiting the number of proofs/headers.
+EpochHeadersOrProofs: Contains the set of proofs & headers to help the client, as requested in `GetEpochHeadersOrProofs`.
+```
+
+Option 2: This option benefits from assumptions on the pattern of proof generation (likely at least one for every ~120 epochs) and by assuming proofs are well distributed (all full nodes have all proofs). The idea here is to pass a bitmap to the `GetEpochHeadersOrProofs` msg that indicates if the client knows the validator set for the past `SIZE_OF_BITMAP` (~240?) epochs. This allows the full node to optimize the set of proofs sent back, while minimizing computation for the client. The main win here would be in a case like this: the client uses a proof to sync to epoch 120, then syncs further to epoch 121, so it knows the validator set of 120 and 121. Then the client goes down for > 120 epochs, and when it comes back the server can figure out to send a proof from 120-240 that we assume is generated, rather than sending 119 epoch headers to verify.
+
+The msgs are:
+```
+GetEpochHeadersOrProofs:
+  Client sends (highestEpoch uint, knownValSets *big.Int)
+    `knownValSets` here would be a bitmap of the last ~240 or so epochs that the client knows the val set for
+  Server computes the optimal path of proofs & headers that the client can use to sync fastest and send them back using `EpochHeadersOrProofs`.
+EpochHeadersOrProofs:
+  Server sends this back from `GetEpochHeadersOrProofs` with the optimal set of proofs and headers this client can utilize.
+```  
+
+It may be preferred to combine both options, and send a bitmap with `handleLogic == false` in option 1.
+
+Regardless of the option chosen, the `les` client must then be modified to request proofs & headers during `lightest` sync. This will likely replace the function `getEpochOrNormalHeaders` within `eth/downloader`. This will also involve a new function to update the underlying validator set and blockchain data, which will be built to work with checkpoints as well.
+
+Upon receiving proofs and headers in a response message, the light client will verify headers the same way as before, and will verify proofs via a call into `celo-bls-go`'s `verify` fn.
 
 The verify function requires information on an epoch's validator set, so this will likely require some access changes (private -> public) and exposing more data structures to allow the validator set to be queried and modified upon proof verification.
 
--- a/CIPs/0017.md
+++ b/CIPs/0017.md
@@ -81,7 +81,7 @@ ProofInputs struct {
 // My rough calculations show this struct is ~439 bytes for a non-batched proof with no val set differences.
 LightPlumoProof struct {
   Proof             []byte // Serialized proof. Latest estimate from @psi = 383 bytes
-  FirstEpoch        LightEpochBlock // First epoch this proof is over
+  FirstEpoch        uint // First epoch this proof is over - maxNonSigners should be available to client
   LastEpoch         LightEpochBlock // Last epoch this proof is over
   VersionNumber     uint // Proof version number, indicates the corresponding verification key to use
   LastHashToField  []byte  // Hash to field of Last epoch's `snark.EpochBlock` struct. To be used for batch verification.
@@ -97,9 +97,10 @@ The light clients will need to store verification keys corresponding to the exis
 There are two proposed extensions for the `les` protocol:
 
 Option 1:
-This option is more resilient to incomplete proof propagation for full nodes, but requires an extra round trip of messages to achieve this. The idea is for the client to query the server to see what proofs are available, and then request the optimal proofs. If the client wants to minimize network traffic and assume proofs are well distributed, they can also bypass the `GetAvailableProofs` step and just request that the server sends the best proofs it has for the client's current epoch (using the `handleLogic` flag on `GetEpochHeadersOrProofs`).
+
+This option involves one extra round trip in order to get peers' available proofs to choose the "best peer". It's seemingly possible to include this extra round trip in the handshake between peers to optimize performance. If the client wants to minimize network traffic and assume proofs are well distributed, they can also bypass the `GetProofInventory` step and just request that the server sends the best proofs it has for the client's current epoch (using the `handleLogic` flag on `GetEpochHeadersOrProofs`).
 ```
-GetAvailableProofs - Asks the server what proofs it has. Server responds with `[]proofMetadata` of all available proofs.
+GetProofInventory - Asks the server what proofs it has. Server responds with `[]proofMetadata` of all available proofs.
 GetEpochHeadersOrProofs:
   Client sends (highestEpoch uint, handleLogic boolean, requestedProofs []proofMetadata)
     `highestEpoch` is the current highestEpoch the client knows of.
@@ -111,19 +112,34 @@ GetEpochHeadersOrProofs:
 EpochHeadersOrProofs: Contains the set of proofs & headers to help the client, as requested in `GetEpochHeadersOrProofs`.
 ```
 
-Option 2: This option benefits from assumptions on the pattern of proof generation (likely at least one for every ~120 epochs) and by assuming proofs are well distributed (all full nodes have all proofs). The idea here is to pass a bitmap to the `GetEpochHeadersOrProofs` msg that indicates if the client knows the validator set for the past `SIZE_OF_BITMAP` (~240?) epochs. This allows the full node to optimize the set of proofs sent back, while minimizing computation for the client. The main win here would be in a case like this: the client uses a proof to sync to epoch 120, then syncs further to epoch 121, so it knows the validator set of 120 and 121. Then the client goes down for > 120 epochs, and when it comes back the server can figure out to send a proof from 120-240 that we assume is generated, rather than sending 119 epoch headers to verify.
+The client will call `GetProofInventory`, either during the sync or pipelined within the handshake. This information will then be stored for each peer, and before each request the client will assess the "best peer", where "best" here would be defined as the minimum number of proofs & headers needed to sync from the client's epoch to `currEpoch`. This calculation could work as follows: 
 
-The msgs are:
+`numEpochs`: Number of epochs we need to sync, the difference between the current epoch of the canonical chain (`currEpoch`) and the client's highest epoch `clientEpoch` = `currEpoch - clientEpoch`
+
+`optimalPath`: For each peer, compute the optimal path of proofs available based on the client's available VKs and known val sets.
+
+`proofRange` : For each proof, the difference between `LastEpoch` and `FirstEpoch`'s indices.
+
+`peerScore`: This is the total number of proofs and headers needed to sync by this peer. For each peer, subtract the range of each proof in `optimalPath` from `numEpochs`, the result is the number of necessary remaining headers. Add `optimalPath`.length to this for the peer's score. 
+
+Whichever peer has the lowest score indicates the shortest path (best set of proofs) for our client. This could also be short-circuited if the optimal score is provided. Once the best peer is chosen, the client requests the set of proofs and headers from this peer that it has assessed as optimal.
+
+Option 2:
 ```
 GetEpochHeadersOrProofs:
-  Client sends (highestEpoch uint, knownValSets *big.Int)
-    `knownValSets` here would be a bitmap of the last ~240 or so epochs that the client knows the val set for
+  Client sends (highestEpoch uint, knownValSets *big.Int, maxResponses uint)
+    `knownValSets` is a bitmap of the last ~240 or so epochs that the client knows the val set for
+    `maxResponses` indicates the maximum number of (proofs + headers) the client is willing to accept
   Server computes the optimal path of proofs & headers that the client can use to sync fastest and send them back using `EpochHeadersOrProofs`.
 EpochHeadersOrProofs:
   Server sends this back from `GetEpochHeadersOrProofs` with the optimal set of proofs and headers this client can utilize.
-```  
+``` 
+Expected usage:
 
-It may be preferred to combine both options, and send a bitmap with `handleLogic == false` in option 1.
+On initial sync (from genesis): Send one message `GetEpochheadersOrProofs` msg, `knownValSets` is `0`, set `maxResponses` to `numEpochs / 120 + numEpochs % 120 + BUFFER` where `BUFFER` could be lowered towards 0 as we are more confident on proof propagation every 120 epochs. If the request cannot be fulfilled in `maxResponses` the peer sends back `[]`, so the client tries the next peer. If every peer is unable to service this request, the `maxResponses`can be increased or even diregarded.
+
+When syncing after "downtime", the `knownValSets` will now be populated. Potentially this could be limited to only send known validator sets for epochs on "proof boundaries" (epochs % 120 == 0). The server will use this bitmap to respond with the optimal path based on client's known validator sets, while still abiding by the maximum requested.
+
 
 Regardless of the option chosen, the `les` client must then be modified to request proofs & headers during `lightest` sync. This will likely replace the function `getEpochOrNormalHeaders` within `eth/downloader`. This will also involve a new function to update the underlying validator set and blockchain data, which will be built to work with checkpoints as well.
 
--- a/CIPs/0017.md
+++ b/CIPs/0017.md
@@ -53,6 +53,8 @@ Reply to GetPlumoProofMsg. The items in the list are plumo proof structs, {proof
 This may validly contain no plumo proofs if the GetPlumoProofMsg included all proofs this node is aware of.
 ```
 
+This will likely be accompanied by a new `proofFetcher` object to be in charge of fetching these proofs from peers while providing protections against DoS attacks. This feels like overkill, but seems necessary to prevent attacks and maintain state.
+
 The `les` protocol must then also be extended via at least two messages:
 ```
 GetCertifiedEpochsValidatorSet - Sends the current highest epoch known to the sender. Server responds with the minimal set of proofs & headers to give the light client the fastest sync time.
--- a/CIPs/0017.md
+++ b/CIPs/0017.md
@@ -70,7 +70,6 @@ ProofMetadata struct {
 }
 
 // Metadata for each epoch to construct `EpochBlock` for the snark's verify fn
-// May be unnecessary if we can make assumptions on `MaxNonSigners`
 ProofInputs struct {
   Index         uint // Epoch Index 
   MaxNonSigners uint // MinQuorumSize()
@@ -78,7 +77,8 @@ ProofInputs struct {
 
 // All of the data needed by the light client to verify and utilize a proof.
 // This is also crafted to allow for batch verifications using `FirstHashToField`
-// My rough calculations show this struct is ~439 bytes for a non-batched proof with no val set differences.
+// My rough calculations show this struct is ~431 bytes for a non-batched proof with no val set differences.
+// This size increases by 96 bytes for each validator added, so the size is roughly: 431 + (96 * n) where n is the num new validators
 LightPlumoProof struct {
   Proof             []byte // Serialized proof. Latest estimate from @psi = 383 bytes
   FirstEpoch        uint // First epoch this proof is over - maxNonSigners should be available to client
@@ -91,28 +91,26 @@ LightPlumoProof struct {
   DeletedValidators *big.Int // ValSetDiff(FirstEpoch, LastEpoch): bitmap of deleted validators from the `firstEpoch` valset.
 }
 ```
+TODO: The val set diff structure still needs work to develop a succinct representation of a diff that maintains order.
 
 The light clients will need to store verification keys corresponding to the existing versions of the snark circuit, as elaborated in #upgradeability. 
 
-There are two proposed extensions for the `les` protocol:
+The proposed extensions for the `les` protocol:
 
-Option 1:
-
-This option involves one extra round trip in order to get peers' available proofs to choose the "best peer". It's seemingly possible to include this extra round trip in the handshake between peers to optimize performance. If the client wants to minimize network traffic and assume proofs are well distributed, they can also bypass the `GetProofInventory` step and just request that the server sends the best proofs it has for the client's current epoch (using the `handleLogic` flag on `GetEpochHeadersOrProofs`).
+This design involves one extra round trip in order to get peers' available proofs to choose the "best peer". It's possible to include this extra round trip in the handshake between peers to optimize performance, but this may add unnecessary complication. If the client wants to minimize network traffic and assume proofs are well distributed, they can also bypass the `GetProofInventory` step and just request that the server sends the best proofs it has for the client's current epoch (using the `handleLogic` flag on `GetEpochHeadersOrProofs`).
 ```
 GetProofInventory - Asks the server what proofs it has. Server responds with `[]proofMetadata` of all available proofs.
-GetEpochHeadersOrProofs:
-  Client sends (highestEpoch uint, handleLogic boolean, requestedProofs []proofMetadata)
-    `highestEpoch` is the current highestEpoch the client knows of.
-    `handleLogic`: If true, indicates the server should send back the best set of proofs/headers it has
-                   If false, the server should look at `requestedProofs` and send back proofs requested, with headers to fill gaps as appropriate.
+GetPlumoProofsMsg:
+  Client sends (requestedProofs []proofMetadata)
     `requestedProofs`: Set of proofs' metadata requested by the client
-  Server responds with `EpocHeadersOrProofs` dependent on `handleLogic`.
-  Both responses are subject to message size limits limiting the number of proofs/headers.
-EpochHeadersOrProofs: Contains the set of proofs & headers to help the client, as requested in `GetEpochHeadersOrProofs`.
+  Server responds with `PlumoProofsMsg` containing the requested proofs, or an empty response if the server does have hold them.
+PlumoProofMsg:
+  Server sends (requestedProofs []LightPlumoProof)
+    `requestedProofs`: Set of proofs with all necessary data for the client to verify these proofs and utilize them to sync.
+  Client verifies and consumes these proofs in conjunction with requested headers (explained in sync model below) to sync.
 ```
 
-The client will call `GetProofInventory`, either during the sync or pipelined within the handshake. This information will then be stored for each peer, and before each request the client will assess the "best peer", where "best" here would be defined as the minimum number of proofs & headers needed to sync from the client's epoch to `currEpoch`. This calculation could work as follows: 
+The intended usage would be for the client to call `GetProofInventory`, either during the sync or pipelined within the handshake. This information will then be stored for each peer, and before each request the client will assess the "best peer", where "best" here would be defined as the minimum number of proofs & headers needed to sync from the client's epoch to `currEpoch`. This calculation could work as follows: 
 
 `numEpochs`: Number of epochs we need to sync, the difference between the current epoch of the canonical chain (`currEpoch`) and the client's highest epoch `clientEpoch` = `currEpoch - clientEpoch`
 
@@ -124,28 +122,9 @@ The client will call `GetProofInventory`, either during the sync or pipelined wi
 
 Whichever peer has the lowest score indicates the shortest path (best set of proofs) for our client. This could also be short-circuited if the optimal score is provided. Once the best peer is chosen, the client requests the set of proofs and headers from this peer that it has assessed as optimal.
 
-Option 2:
-```
-GetEpochHeadersOrProofs:
-  Client sends (highestEpoch uint, knownValSets *big.Int, maxResponses uint)
-    `knownValSets` is a bitmap of the last ~240 or so epochs that the client knows the val set for
-    `maxResponses` indicates the maximum number of (proofs + headers) the client is willing to accept
-  Server computes the optimal path of proofs & headers that the client can use to sync fastest and send them back using `EpochHeadersOrProofs`.
-EpochHeadersOrProofs:
-  Server sends this back from `GetEpochHeadersOrProofs` with the optimal set of proofs and headers this client can utilize.
-``` 
-Expected usage:
-
-On initial sync (from genesis): Send one message `GetEpochheadersOrProofs` msg, `knownValSets` is `0`, set `maxResponses` to `numEpochs / 120 + numEpochs % 120 + BUFFER` where `BUFFER` could be lowered towards 0 as we are more confident on proof propagation every 120 epochs. If the request cannot be fulfilled in `maxResponses` the peer sends back `[]`, so the client tries the next peer. If every peer is unable to service this request, the `maxResponses`can be increased or even diregarded.
-
-When syncing after "downtime", the `knownValSets` will now be populated. Potentially this could be limited to only send known validator sets for epochs on "proof boundaries" (epochs % 120 == 0). The server will use this bitmap to respond with the optimal path based on client's known validator sets, while still abiding by the maximum requested.
-
-
-Regardless of the option chosen, the `les` client must then be modified to request proofs & headers during `lightest` sync. This will likely replace the function `getEpochOrNormalHeaders` within `eth/downloader`. This will also involve a new function to update the underlying validator set and blockchain data, which will be built to work with checkpoints as well.
+The `les` client must then be modified to use these msgs to request proofs & headers during `lightest` sync. This will likely replace the function `getEpochOrNormalHeaders` within `eth/downloader`. This will also involve a new function to update the underlying validator set and blockchain data, which will be built to work with checkpoints as well.
 
 Upon receiving proofs and headers in a response message, the light client will verify headers the same way as before, and will verify proofs via a call into `celo-bls-go`'s `verify` fn.
-
-The verify function requires information on an epoch's validator set, so this will likely require some access changes (private -> public) and exposing more data structures to allow the validator set to be queried and modified upon proof verification.
 
 ### Upgradeability
 In order to support an upgradeable snark circuit, we utilize a `versionNumber` parameter with every proof. This corresponds to a version of the snark circuit and thus a corresponding verification key. The clients are expected to be shipped with all available verification keys, and can be upgraded to receive added verification keys.
@@ -159,7 +138,29 @@ Bad proofs should be prevented from propagated throughout the network via the `v
 
 ## Alternative Solutions
 
-Alternatively, `GetPlumoProofMsg` could utilize a bloom filter instead, allowing for a constant sized message. However, bloom filters allow for false positives: a false positive here would indicate to a peer that the sender has a proof when the sender does not. This could prevent certain proofs from ever getting to some nodes. For example, say node A is on the network and has received a certain amount of proofs via traditional propagation. Node A then goes down for a short period, and within that period proof P is added and propagated. Node A then comes back online, and sends a `GetPlumoProofsMsg` to request any proofs it may have missed, and sends along the corresponding bloom filter for the proofs it holds. If proof P happens to be a false positive for the set of proofs node A already has, then all of node A's peers will simply return nothing. Thus node A will never receive proof P. This can then propagate similarly through the network either through a similar path as node A, or for any nodes where all of their peers have already "lost" proof P.
+An alternative `les` protocol:
+```
+GetEpochHeadersOrProofs:
+  Client sends (highestEpoch uint, knownValSets [][]uint, knownVersion []uint, verificationTimeLimit uint)
+    `knownValSets` is an array of [start, end] pairs indicating ranges of validator sets the client knows
+    `knownVersions` is an array of `versionNumber`s that this client has the VK for
+    `verificationTimeLimit` indicates the maximum number of (profos + headers) the client is willing to accept.
+    This would be calculated as the verification time for a proof * the amount of expected proofs + the verification time of a traditional header * the amount of expected headers + some buffer.
+  Server computes the optimal path of proofs & headers that the client can use to sync fastest and send them back using `EpochHeadersOrProofs`.
+EpochHeadersOrProofs:
+  Server sends this back from `GetEpochHeadersOrProofs` with the optimal set of proofs and headers this client can utilize.
+``` 
+Expected usage:
+
+On initial sync (from genesis): Send one message `GetEpochheadersOrProofs` msg, `knownValSets` is `[]`, set `verificationTimeLimit` to `(numEpochs / 120 * PROOF_VERIFICATION_TIME) + (numEpochs % 120 * HEADER_VERIFICATION_TIME) + BUFFER` where `BUFFER` could be lowered towards 0 as we are more confident on proof propagation every 120 epochs. If the request cannot be fulfilled within `verificationTimeLimit` the peer sends back `[]`, so the client tries the next peer. If every peer is unable to service this request, the `verificationTimeLimit`can be increased or even disregarded.
+
+When syncing after "downtime", the `knownValSets` will now be populated. This could probably be limited to only send known validator sets for epochs on "proof boundaries" (epochs % 120 == 0). The server will use this list to respond with the optimal path based on client's known validator sets, while still abiding by the maximum requested.
+
+
+
+An alternative `eth` protocol:
+
+`GetPlumoProofMsg` could utilize a bloom filter instead, allowing for a constant sized message. However, bloom filters allow for false positives: a false positive here would indicate to a peer that the sender has a proof when the sender does not. This could prevent certain proofs from ever getting to some nodes. For example, say node A is on the network and has received a certain amount of proofs via traditional propagation. Node A then goes down for a short period, and within that period proof P is added and propagated. Node A then comes back online, and sends a `GetPlumoProofsMsg` to request any proofs it may have missed, and sends along the corresponding bloom filter for the proofs it holds. If proof P happens to be a false positive for the set of proofs node A already has, then all of node A's peers will simply return nothing. Thus node A will never receive proof P. This can then propagate similarly through the network either through a similar path as node A, or for any nodes where all of their peers have already "lost" proof P.
 
 `GetPlumoProofMsg` could also request *all* plumo proofs from the receiver, leaving the responsibility of filtering seen proofs with the caller. This is a simpler implementation, but is less scalable as we increase the amount of stored proofs. 
 
--- a/CIPs/0017.md
+++ b/CIPs/0017.md
@@ -20,9 +20,10 @@ The prover/verifier for plumo proofs have already been implemented, and the expo
 ## Proposed Solution
 
 We propose to add two new RPCs: 
-`AddProof(proof []byte, proofId uint)`
-and
-`Proofs() ([][]byte)`. When an eventual "proving service" generates a new proof it will publish this proof to a full node that it is running, which will then propogate the proof throughout the network.
+* `AddProof(proof []byte, proofId uint)`
+* `Proofs() ([][]byte)`.
+
+When an eventual "proving service" generates a new proof it will publish this proof to a full node that it is running, which will then propagate the proof throughout the network.
 
 This `AddProof` function will store the proof in a new database `plumoproofdata`, keyed by a unique proof ID.
 
@@ -55,9 +56,9 @@ Specific aspects of the `les` changes may have tradeoffs that I've yet to encoun
 
 ## Risks
 
-Altering the `eth` protocol introduces some risk of introducing bugs.
+Altering the `eth` protocol introduces some risk of introducing bugs, which could inhibit a nodes ability to connect to the network.
 
-Introducing bugs to mobile client syncing.
+Introducing bugs to mobile client syncing. Possible effects of a faulty implementation include not being able to sync to the network, and introducing a security vulnerability which would allow the node to sync to an attack controlled chain.
 
 ## Useful Links
 - [Celo snark go interface](https://github.com/celo-org/celo-bls-go)
--- a/CIPs/0017.md
+++ b/CIPs/0017.md
@@ -1,5 +1,5 @@
 
-# CIP [0016]: Integrate Plumo Proofs into the lightest sync protocol
+# CIP [0017]: Integrate Plumo Proofs into the lightest sync protocol
 
 - Date: 2020-07-27
 - Author: @lucasege
--- a/CIPs/0017.md
+++ b/CIPs/0017.md
@@ -7,15 +7,17 @@
 
 ## Overview
 
-In order to continue supporting the dream of a seamless mobile sync experience, the lightest sync protocol must be extended to support PLUMO Zk-SNARKs which can quickly verify transitions from epoch X to epoch X + Y, where Y could be on the order of 10s-100s of epochs.
+In order to continue supporting the development of a seamless mobile sync experience, the lightest sync protocol should be extended to support Plumo Zk-SNARKs which can quickly verify transitions from epoch X to epoch X + Y, where Y could be on the order of 10s-100s of epochs.
 
-The prover/verifier for plumo proofs have already been implemented, and the exposed go interface is available [here](https://github.com/celo-org/celo-bls-go)
+The prover/verifier for Plumo proofs have already been implemented, and the exposed go interface is available [here](https://github.com/celo-org/celo-bls-go)
+
+More information on Plumo can be found in the [reference paper](https://docs.zkproof.org/pages/standards/accepted-workshop3/proposal-plumo_celolightclient.pdf)
 
 ## Goals
 
 - Support adding/storing proofs on full nodes
-- Gossip these proofs to propogate them through the network
-- Extend the `les` protocol such that the server will provide these proofs in place of headers (when appropriate), and the client will consume/verify proofs accordingly.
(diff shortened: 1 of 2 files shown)


CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+## Goals
+
+- Support adding/storing proofs on full nodes
+- Gossip these proofs to propogate them through the network
+- Extend the `les` protocol such that the server will provide these proofs in place of headers (when appropriate), and the client will consume/verify proofs accordingly.
+
+## Proposed Solution
+
+We propose to add two new RPCs: 
+`AddProof(proof []byte, proofId uint)`
```
tkporter, line 23: Are the proofs signed, or can anyone submit a proof and it gets gossiped through the network if the proof is verified?

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+To propogate these proofs throughout the network, we also propose to extend the `eth` protocol with two new message types: 
+```
+PlumoProofMsg - Sends plumo proof(s) either in response to a `GetPlumoProofMsg` or after a successful RPC call to `AddProof`
+GetPlumoProofMsg - Request plumo proof(s) from a peer
+```
+`GetPlumoProofMsg` will be implemented by having the requester send a set of proof IDs that they hold, and ask for any they are missing. This may be optimized using a bloom filter as well.
+
+The `les` protocol must then also be extended via at least two messages:
+```
+GetEpoch - Sends the current highest epoch known to the sender. Server responds with the minimal set of proofs & headers to give the light client the fastest sync time.
```
tkporter, line 38: Slightly confused by the wording here and line 39, so the `GetEpoch` is a message sent by a client and then the server responds with `Epochs`?

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+`AddProof(proof []byte, proofId uint)`
+and
+`Proofs() ([][]byte)`. When an eventual "proving service" generates a new proof it will publish this proof to a full node that it is running, which will then propogate the proof throughout the network.
+
+This `AddProof` function will store the proof in a new database `plumoproofdata`, keyed by a unique proof ID.
+
+To propogate these proofs throughout the network, we also propose to extend the `eth` protocol with two new message types: 
+```
+PlumoProofMsg - Sends plumo proof(s) either in response to a `GetPlumoProofMsg` or after a successful RPC call to `AddProof`
+GetPlumoProofMsg - Request plumo proof(s) from a peer
```
tkporter, line 32: Under what condition would a `GetPlumoProofMsg` be sent to all its peers?
Would be good to have a little more detail here on how exactly the gossiping will work

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+- Support adding/storing proofs on full nodes
+- Gossip these proofs to propogate them through the network
+- Extend the `les` protocol such that the server will provide these proofs in place of headers (when appropriate), and the client will consume/verify proofs accordingly.
+
+## Proposed Solution
+
+We propose to add two new RPCs: 
+`AddProof(proof []byte, proofId uint)`
+and
+`Proofs() ([][]byte)`. When an eventual "proving service" generates a new proof it will publish this proof to a full node that it is running, which will then propogate the proof throughout the network.
```
tkporter, line 25: How many bytes do we expect the proofs to be?

CIPs/0017.md
lucasege: I think they wouldn't need to be signed, any proof that is verified should be able to be gossiped, allowing us to decentralize the proof creation. C labs will host an initial "proving service" to do this, but community members should be able to contribute.

This is where I would see issues with proof IDs coming in to play

CIPs/0017.md
lucasege: Depends on many factors, but the estimates I've seen for what we may use range from ~5kB - ~12kB. 
The dependent factors include (probably not exhaustive): max number of epochs the circuit is designed for and the type of proof protocol (marlin vs. groth16)

CIPs/0017.md
lucasege: Yes, I imagine this working very similarly to existing `les` messages, such as `GetBlockHeadersMsg` and `BlockHeadersMsg` where I believe the client sends `GetBlockHeadersMsg` with some corresponding parameters (block number/hash) to a full node, and the full node sends back a `BlockHeadersMsg` with the same parameters for the light client to consume.
I'd imagine `GetEpochsMsg` should work similarly, where the client sends this msg with the parameter of the client's highest known epoch, and the full node responds with `EpochsMsg` with the same parameters and the corresponding set of headers/proofs to progress the client's sync

CIPs/0017.md
lucasege: The goal of this msg `GetPlumoProofMsg` is to allow new syncing nodes to fetch previous proofs, and to allow nodes that have gone down to fetch proofs they may have missed.

I'm not sure exactly where this should be handled, but I'm guessing within `downloader`, when a node is syncing, it would send this `get` message.

In regard to sending to all its peers, I imagined sending one message to each peer you know of, and then just taking the union set of all proofs from those responses as your set (in case any peer is similarly out of sync). Maybe this is overkill, if for example peers only respond with `PlumoProofMsg`s if they are in sync already.

Does this address what you were getting at? I'm not very familiar with the already implemented gossiping protocols so appreciate feedback and advice

CIPs/0017.md
kevjue: Ya.  Agreed about potential issues with proof IDs.  Seems like it's a DOS attack vector.  E.g. a malicious party could submit many duplicated proof but with different proofIDs to fill up all the full nodes' disks.

kevjue, line 23: Will this RPC call verify that the plumo proof is valid?  Or does that even need to be done? (I don't know enough about the plumo/SNARK proofs to know if that's even a valid question).

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+
+This `AddProof` function will store the proof in a new database `plumoproofdata`, keyed by a unique proof ID.
+
+To propogate these proofs throughout the network, we also propose to extend the `eth` protocol with two new message types: 
+```
+PlumoProofMsg - Sends plumo proof(s) either in response to a `GetPlumoProofMsg` or after a successful RPC call to `AddProof`
```
kevjue, line 31: Should there be a new message type that will broadcast the new proof's metadata.  E.g. the proof's ID/hash and epoch range?  Analogous to `NewBlockHashes` and `NewPooledTransactionHashes` (https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblockhashes-0x01).

kevjue, line 31: Seems like there could be plumo proofs that have overlapping epoch ranges.  In that case, are all the full nodes expected to store all of the proofs, or will it prune redundant proofs to minimize disk space?

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+GetPlumoProofMsg - Request plumo proof(s) from a peer
+```
+`GetPlumoProofMsg` will be implemented by having the requester send a set of proof IDs that they hold, and ask for any they are missing. This may be optimized using a bloom filter as well.
+
+The `les` protocol must then also be extended via at least two messages:
+```
+GetEpochsMsg - Sends the current highest epoch known to the sender. Server responds with the minimal set of proofs & headers to give the light client the fastest sync time.
```
kevjue, line 38: nit:  Consider renaming to `GetCertifiedEpochsValidatorSet`, or something along those lines.

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+EpochsMsg - Receives a mix of header(s) and proof(s) from the server.
```
nategraf, line 39: I would propose that that instead of this extension we add messages that will allow the client to be inform of what proofs the server can provide filtered by block range, and to download those proofs. That would give the client flexibility to decide what mix of proofs and headers it wants to use for syncing.

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+This `AddProof` function will store the proof in a new database `plumoproofdata`, keyed by a unique proof ID.
+
+To propogate these proofs throughout the network, we also propose to extend the `eth` protocol with two new message types: 
+```
+PlumoProofMsg - Sends plumo proof(s) either in response to a `GetPlumoProofMsg` or after a successful RPC call to `AddProof`
+GetPlumoProofMsg - Request plumo proof(s) from a peer
+```
+`GetPlumoProofMsg` will be implemented by having the requester send a set of proof IDs that they hold, and ask for any they are missing. This may be optimized using a bloom filter as well.
+
+The `les` protocol must then also be extended via at least two messages:
```
nategraf, line 36: Implementation note: `les/4` is pending on `master` to include gateway fee changes that are not fully complete. We will want to provide MVP support for those (which may even be a stub) or figure out another way to handle this if we want to have these modification as part of `les/4` (Another option is to move gateway fee messages to `les/5` as well if there is no time to get it fully ready)

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+```
+GetEpochsMsg - Sends the current highest epoch known to the sender. Server responds with the minimal set of proofs & headers to give the light client the fastest sync time.
+EpochsMsg - Receives a mix of header(s) and proof(s) from the server.
+```
+
+The `les` client must then be modified to send a `GetEpochsMsg` message during `lightest` sync. This will likely replace the function `getEpochOrNormalHeaders` within `eth/downloader`.
+
+Upon receiving the `EpochsMsg` response message, the light client must then verify headers and proofs provided. Proofs will be verified via a call into `celo-bls-go`'s `verify` fn.
+
+The verify function requires information on an epoch's validator set, so this will likely require some access changes (private -> public) and exposing more data structures to allow the validator set to be queried and modified upon proof verification.
```
nategraf, line 46: Could you elaborate on this a bit? I don't quite understand from this description.

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+
+`GetPlumoProofMsg` could also request *all* plumo proofs from the receiver, leaving the responsibility of filtering seen proofs with the caller. This is a simpler implementation, but is less scalable as we increase the amount of stored proofs. 
+
+On the type of ID to use for proofs: an incrementing counter seems most "user-friendly" because of the ability to sense gaps, but could introduce some synchronization issues in the future.
+
+Specific aspects of the `les` changes may have tradeoffs that I've yet to encounter, so any input and feedback on things to consider there would be helpful.
+
+## Risks
+
+Altering the `eth` protocol introduces some risk of introducing bugs.
```
nategraf, line 58: Just some added clarification on the risk.
```suggestion
Altering the `eth` protocol introduces some risk of introducing bugs, which could inhibit a nodes ability to connect to the network.
```

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+
+Introducing bugs to mobile client syncing.
```
nategraf, line 60:
```suggestion
Introducing bugs to mobile client syncing. Possible effects of a faulty implementation include not being able to sync to the network, and introducing a security vulnerability which would allow the node to sync to an attack controlled chain.
```

CIPs/0017.md
lucasege: Yup good points - I plan to key by epoch range like you mentioned @kevjue - specifically `types.PlumoProofEpochs{ firstEpoch, lastEpoch }`

I had originally wanted to follow this path, but thought we might want incremental IDs. But these are good points, and the idea of using a bloom filter makes the incremental IDs unnecessary in my mind

lucasege: Yes in my mind this could and should be done. This would help prevent potential for DDoS via someone spamming bad proofs to one node which then propagate to the whole network. I'll add this to the CIP after I finish synthesizing the rest of the feedback

CIPs/0017.md
lucasege: This is certainly a concern - Currently I was planning to just store all proofs for simplicity, although pruning redundant proofs wouldn't add much burden. Proofs aren't very large (estimated between 5-12kB) so I don't think this is necessary, but please let me know what you think.

CIPs/0017.md
lucasege: Hm good point, thanks for pointing this out. I'm inclined to go the second route (plumo = les/4, gateway fees to les/5), let me know if you agree or if this changes anything else @nategraf

CIPs/0017.md
lucasege: hm - I think this is something worth discussing synchronously. I'm personally partial to the original plan, because it seems to reduce any necessary back/forth between client/server.
I don't see any uses for the client being able to chose the mix of proofs/headers, is there a benefit you see?
The motivation behind the original approach was that the light client doesn't care _how_ it syncs, only that it syncs in the fastest way. It seems logical imo for the server (with all relevant info, no need to download more) to interpret this and return the data necessary to make this happen

nategraf:
Some high level questions I'd like for us to shake out here are:
* What is the data structure of a proof?
   * What fields do we need to verify a proof?
   * What fields do we need to index proofs for retrieval?
   * Is a proof unique when under a particular identifier (e.g. start and end epoch)?
* What does a typical proof distribution flow look like among full nodes? (Including examples may help us shake out edge cases and give a better mental picture)
   * How big are the proofs? What is the size of a request for the proof IDs versus just returning the proofs themselves?
* What does a typical proof distribution flow look like among light clients?
* What do we do with bad proofs?
   * If a light client receives a bad proof from the server, what does the client do?
   * If a full node receives a bad proof from a peer, what does it do?
   * Is there and chance that a maliciously constructed proof could cause issues (e.g. DoS, indexing bad proofs, etc)?
* How many proofs might we expect to handle?
   * Can we establish an upper bound on the number of proofs?
   * Do we need to consider pruning duplicate proofs?
* How can we enable ourselves to extend this protocol in the future?
   * E.g. Can make the message flexible to include things like bloom filters as needed in the future?
   * Can we support improvements to the Plumo circuit? (e.g. proof/circuit/CRS versioning)

CIPs/0017.md
lucasege: This is an area I don't have great context on, I was tipped off that this will likely be an issue by asa.

My understanding is that the underlying `lightchain` data is likely not readily accessible for the light client when syncing in `downloader` so there will need to be changes to expose this information and may require new methods to update the validator set

lucasege:
Responded in-line, great questions - got me thinking about a lot more intricate details.
> What is the data structure of a proof?

A proof in `celo-blockchain` will just be the serialized proof, so a string of bytes. This will also likely be accompanied by the relevant epochs and the validator set diffs.
> What fields do we need to verify a proof?

* Verification key: []byte  (global - ideally I’d like to store this in a SC to allow for governable upgrades, but it appears to be quite long, getting an estimate)
* Proof : []byte
* firstEpoch/lastEpoch : { epochIndex, maxNonSigners, validatorSet }
   * maxNonSigners is the difference between the validator set size and the minimum quorum size
* This is the go fn to call into the snark verifier for reference https://github.com/celo-org/celo-bls-go/blob/master/snark/snark.go

> What fields do we need to index proofs for retrieval?

Currently I’m thinking just {firstEpoch, lastEpoch} to retrieve, but you also need the above params to verify.
> Is a proof unique when under a particular identifier (e.g. start and end epoch)?

This is not guaranteed, no
> What does a typical proof distribution flow look like among full nodes? (Including examples may help us shake out edge cases and give a better mental picture)      

* This is currently in flux, but here is my most up to date plan based on some changes incoming to the CIP:
  * Proving service adds proof to its full node via RPC, node verifies the proof and accepts if it verifies
  * The proof is then added to this node’s `proofDb`
  * This node then broadcasts this proof to all its peers
     * This needs more thought and should be modeled almost exactly how blocks are propagated.
  * The verification/storage/broadcast process then repeats on each peer

> How big are the proofs? What is the size of a request for the proof IDs versus just returning the proofs themselves?

* Proofs are a couple of kBs: for Groth16 estimated between 364 bytes in the best case (no validator set differences) and around 10kB (100 validator set differences)
* The requests are still being fleshed out right now, but a proof request in my current idea - where you would send a message that includes all the `PlumoProofEpochs` spans that you have, and receive a response of all that you’re missing - the request would be the size of the base message + 16 bytes for each proof you have.
* This seems better than receiving multiple kBs per proof for all proofs

> What does a typical proof distribution flow look like among light clients?

TBD, currently no plan for distribution among light clients, only from les server to its client
> What do we do with bad proofs?

TBD
> If a light client receives a bad proof from the server, what does the client do?

Proof is dropped, unsure how we can signal the server, but we should do that too.
> If a full node receives a bad proof from a peer, what does it do?

TBD, but likely signals the peer or drops the peer.
> Is there and chance that a maliciously constructed proof could cause issues (e.g. DoS, indexing bad proofs, etc)?

Assuming we don’t introduce any bugs, no. All proofs should be verified before inserting and before any further broadcasting
> How many proofs might we expect to handle? Can we establish an upper bound on the number of proofs?

Not really, but if we adopt @kevjue 's suggestion that we prune redundant proofs then we could theoretically upper bound the proofs as: `totalEpochs / epochsPerProof`, where `epochsPerProof` has been estimated ~100, and of course `totalEpochs` increases monotonically
> Do we need to consider pruning duplicate proofs? 

I think so, if we can implement it quickly. It isn’t a requirement, but is a worthy optimization and will prevent some nasty logic when trying to find a proof (verifying that you didn’t miss a faster proof path)
> How can we enable ourselves to extend this protocol in the future? E.g. Can make the message flexible to include things like bloom filters as needed in the future?

* TBD, will discuss on our sync.
* We should discuss the bloom filter specifically since I'm not sure we can actually utilize one here since the key space is constantly increasing.
> Can we support improvements to the Plumo circuit? (e.g. proof/circuit/CRS versioning)

* Yes, this is also something we should discuss. My plan to do this would be to publish the hash of the current VK to `blockchain_parameters` and verify on that, although I’m not sure how regularly we can/should do that.
* We could also extend this logic to do something like storing a dictionary i.e.` { 0-1000: VK_0, 1000: VK_1 }` such that the original proofs are still valid and useable. This introduces more complication on the verification side, though

CIPs/0017.md
lucasege: Very good point, I'll add this to the CIP, wasn't aware of how exactly this mechanism worked

CIPs/0017.md
```diff
@@ -0,0 +1,96 @@
...
+To propagate these proofs throughout the network, we also propose to extend the `eth` protocol with three new message types: 
+```
+NewPlumoProof (0x019)
+[[firstEpoch, lastEpoch, versionNumber], ...]
+Specify one or more new plumo proofs which have appeared on the network.
+This is intended to be honored with a proceeding GetPlumoProofsMsg to request the proof.
+
+GetPlumoProofsMsg (0x1a)
+[[firstEpoch, lastEpoch, versionNumber], ...]
+Sent from node A to node B with the set of proofs that node A currently has, requires node B to return a PlumoProofsMsg.
```
nategraf, line 41: In steady state, sending the `GetPlumoProodsMsg` with the list of proofs the node has will be much larger than sending a light of the proofs the node wants. Especially for use by long-running nodes who send `GetPlumoProofs` in response to a `NewPlumoProof` message indicating the existence of a proof the node does not have, it will be more efficient to send a message containing only the identifier for the proof the node wants. Adding a flag, or other indicator in the request to allow for requesting specific proof(s) would be beneficial.

Another case where receiving every proof the requestor does not have may not be ideal is if the requester is only interested in a certain subset of proofs (e.g. proofs with the latest version number, or proofs after a certain epoch)

CIPs/0017.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+The `les` client must then be modified to send a `GetCertifiedEpochsValidatorSet` message during `lightest` sync. This will likely replace the function `getEpochOrNormalHeaders` within `eth/downloader`.
+
+Upon receiving the `CertifiedEpochsValidatorSet` response message, the light client must then verify headers and proofs provided. Proofs will be verified via a call into `celo-bls-go`'s `verify` fn.
+
+The verify function requires information on an epoch's validator set, so this will likely require some access changes (private -> public) and exposing more data structures to allow the validator set to be queried and modified upon proof verification.
+
+### Upgradeability
+In order to support an upgradeable snark circuit, we utilize a `versionNumber` parameter with every proof. This corresponds to a version of the snark circuit and thus a corresponding verification key. These version numbers will be stored on chain, within the `blockchain_parameters` smart contract, with the corresponding hash of the verification key. The clients are expected to store all applicable verification keys.
```
nategraf, line 63: Doesn't storing the version numbers on-chain create a chicken-and-egg problem in that the user of a plumo proof is not synced to the chain yet? It may be necessary to ship verification keys in the client code instead.

CIPs/0017.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Alternative Solutions
+
+Alternatively, `GetPlumoProofMsg` could utilize a bloom filter instead, allowing for a constant sized message. However, bloom filters allow for false positives: a false positive here would indicate to a peer that the sender has a proof when the sender does not. This could prevent certain proofs from ever getting to some nodes. For example, say node A is on the network and has received a certain amount of proofs via traditional propagation. Node A then goes down for a short period, and within that period proof P is added and propagated. Node A then comes back online, and sends a `GetPlumoProofsMsg` to request any proofs it may have missed, and sends along the corresponding bloom filter for the proofs it holds. If proof P happens to be a false positive for the set of proofs node A already has, then all of node A's peers will simply return nothing. Thus node A will never receive proof P. This can then propagate similarly through the network either through a similar path as node A, or for any nodes where all of their peers have already "lost" proof P.
```
nategraf, line 77: 👍 One place where bloom filters could be handy is if the proof serving peer provides a filter to help answer the client peer's question of "does this peer have a proof with identifier X". That is not how the protocol is specified here though, so I agree that it doesn't seem relevant here.

CIPs/0017.md
lucasege: Yes certainly, I was wondering if I should split this message into two messages, one to request specifics and one to request any you may be missing.

I like the idea of including this in one message since the logic is of course very similar. 

I'll adapt the spec to follow this idea ^ thanks

CIPs/0017.md
lucasege: This is a great point - I was thinking we could ship the VKs in the client code, but then have the clients verify each key against the stored reference hash on chain, but of course if you aren't synced you won't see that hash.

I agree, I thought through some plans to allow for this, but if we want to use chain state to verify keys then I believe you need to set boundaries for verification keys, i.e. you can only use version n to prove transitions on epochs greater than X (since you publish the hash to the chain in epoch X, can't access it before). I think this is worse than just shipping all VKs with the client like you mentioned, since then we can upgrade our circuit and use it for any epochs, allowing us to replace a potentially inferior proof scheme.

CIPs/0017.md
```diff
@@ -0,0 +1,69 @@
...
+## Overview
+
+In order to continue supporting the dream of a seamless mobile sync experience, the lightest sync protocol must be extended to support PLUMO Zk-SNARKs which can quickly verify transitions from epoch X to epoch X + Y, where Y could be on the order of 10s-100s of epochs.
+
+The prover/verifier for plumo proofs have already been implemented, and the exposed go interface is available [here](https://github.com/celo-org/celo-bls-go)
+
+## Goals
+
+- Support adding/storing proofs on full nodes
+- Gossip these proofs to propogate them through the network
```
psivesely, line 17: propagate*

CIPs/0017.md
psivesely: We've decided on Groth16 proofs. Marlin and Groth16 both have fully succinct proofs, so the proof size is independent of the number of epochs the circuit is designed for. Groth16 proofs over BW6-761 will be 764B. You will also need to send the client half the instance. The instance includes the public keys at epochs A and B, where B is some number of epochs later. The client will already have validator keys for epoch A and needs the keys at epoch B. 100 public keys over BLS12-377 are 18.9KB.

CIPs/0017.md
psivesely: Validators should have a fixed list of epoch number pairs (A_1, B_1), (A_2,B_2), ... they accept proofs for. The distance B_i - A_i will be in some fixed set we must determine before carrying out the setup "ceremony." I don't think it makes sense to store (0,32) and (1,33). I think just storing the minimal number of proofs to get from the genesis epoch to the current (or as close as the different fixed lengths let you) is ideal. There's a new circuit @kobigurk has been working on that allows for dummy epoch data and maybe that actually allows for proving an arbitrary number of epochs up to some limit. Hopefully he can chime in here. That would be really nice, as nodes would only have to ceil(t/~120) proofs, where t is the current epoch number.

As for duplicate proofs, nodes should probably just store the first valid proof they receive for a given instance (epoch number pair).

psivesely:
> Responded in-line, great questions - got me thinking about a lot more intricate details.
> 
> > What is the data structure of a proof?
> 
> A proof in `celo-blockchain` will just be the serialized proof, so a string of bytes. This will also likely be accompanied by the relevant epochs and the validator set diffs.
> 
> > What fields do we need to verify a proof?
> 
>     * Verification key: []byte  (global - ideally I’d like to store this in a SC to allow for governable upgrades, but it appears to be quite long, getting an estimate)

The verification key is quite short, especially because the actual instance that's verified is a hash of the initial and final epochs. The prover key is the one that's very long.

>     * Proof : []byte
> 
>     * firstEpoch/lastEpoch : { epochIndex, maxNonSigners, validatorSet }
>       
>       * maxNonSigners is the difference between the validator set size and the minimum quorum size

You don't need this to verify the proof. If the proof verifies it implies this number is < 1/3 of validators.

>     * This is the go fn to call into the snark verifier for reference https://github.com/celo-org/celo-bls-go/blob/master/snark/snark.go
> 
> 
> > What fields do we need to index proofs for retrieval?
> 
> Currently I’m thinking just {firstEpoch, lastEpoch} to retrieve, but you also need the above params to verify.
> 
> > Is a proof unique when under a particular identifier (e.g. start and end epoch)?
> 
> This is not guaranteed, no

Zero-knowledge proofs and interactive proofs always use randomness, but we're using a non-zk version of the non-interactive Groth16 proof system, which results in unique proofs given the same initial and final epoch (short of the existence of a fork that merges back at some later point in the epoch message chain).

> > What does a typical proof distribution flow look like among full nodes? (Including examples may help us shake out edge cases and give a better mental picture)
> 
>     * This is currently in flux, but here is my most up to date plan based on some changes incoming to the CIP:
>       
>       * Proving service adds proof to its full node via RPC, node verifies the proof and accepts if it verifies
>       * The proof is then added to this node’s `proofDb`
>       * This node then broadcasts this proof to all its peers
>         
>         * This needs more thought and should be modeled almost exactly how blocks are propagated.
>       * The verification/storage/broadcast process then repeats on each peer
> 
> 
> > How big are the proofs? What is the size of a request for the proof IDs versus just returning the proofs themselves?
> 
>     * Proofs are a couple of kBs: for Groth16 estimated between 364 bytes in the best case (no validator set differences) and around 10kB (100 validator set differences)

Proofs for Groth16 are always the same size over a given curve. Our proofs are over BW6-761, so they will be 383B. Keys are in group 2 of BLS12-377, so they are 95B each. I think it just makes sense to send the 100 keys instead of the diff because a maximum saving 9.5kB is not worth adding the logic of the diff.

>     * The requests are still being fleshed out right now, but a proof request in my current idea - where you would send a message that includes all the `PlumoProofEpochs` spans that you have, and receive a response of all that you’re missing - the request would be the size of the base message + 16 bytes for each proof you have.
> 
>     * This seems better than receiving multiple kBs per proof for all proofs
> 
> 
> > What does a typical proof distribution flow look like among light clients?
> 
> TBD, currently no plan for distribution among light clients, only from les server to its client
> 
> > What do we do with bad proofs?
> 
> TBD
> 
> > If a light client receives a bad proof from the server, what does the client do?
> 
> Proof is dropped, unsure how we can signal the server, but we should do that too.
> 
> > If a full node receives a bad proof from a peer, what does it do?
> 
> TBD, but likely signals the peer or drops the peer.
> 
> > Is there and chance that a maliciously constructed proof could cause issues (e.g. DoS, indexing bad proofs, etc)?
> 
> Assuming we don’t introduce any bugs, no. All proofs should be verified before inserting and before any further broadcasting
> 
> > How many proofs might we expect to handle? Can we establish an upper bound on the number of proofs?
> 
> Not really, but if we adopt @kevjue 's suggestion that we prune redundant proofs then we could theoretically upper bound the proofs as: `totalEpochs / epochsPerProof`, where `epochsPerProof` has been estimated ~100, and of course `totalEpochs` increases monotonically
> 
> > Do we need to consider pruning duplicate proofs?
> 
> I think so, if we can implement it quickly. It isn’t a requirement, but is a worthy optimization and will prevent some nasty logic when trying to find a proof (verifying that you didn’t miss a faster proof path)
> 
> > How can we enable ourselves to extend this protocol in the future? E.g. Can make the message flexible to include things like bloom filters as needed in the future?
> 
>     * TBD, will discuss on our sync.
> 
>     * We should discuss the bloom filter specifically since I'm not sure we can actually utilize one here since the key space is constantly increasing.
> 
> 
> > Can we support improvements to the Plumo circuit? (e.g. proof/circuit/CRS versioning)
> 
>     * Yes, this is also something we should discuss. My plan to do this would be to publish the hash of the current VK to `blockchain_parameters` and verify on that, although I’m not sure how regularly we can/should do that.
> 
>     * We could also extend this logic to do something like storing a dictionary i.e.` { 0-1000: VK_0, 1000: VK_1 }` such that the original proofs are still valid and useable. This introduces more complication on the verification side, though

@kobigurk Can you explain if it's possible with the new circuit that allows "dummy epoch messages" to create a proof covering any number of epochs up to some limit. I think nodes should only need to store about 3 proofs per year passed assuming this is the case that limit is 120 as suggested elsewhere. Again if this is the case then during epoch 256 a node would just be storing proofs for (0, 120), (120, 240), (240, 256). There is the case to consider where a client goes offline for a long time right at 180 and comes back at 256. They shouldn't have to download and verify 60 headers before they receive the next SNARK, so storing more incremental proofs will help in this scenario.

psivesely:
One thing missing from discussion here is batch verification. Batch verification refers to using a modified version of a verification algorithm to verify multiple SNARKs/ signatures/ etc. together faster than they could be independently verified using the standard verification algorithm. If we ensure the proofs we store line up (e.g., (0, 120), (120, 240), (240, 256)), then in our case there is another optimization we can use: following the example above, a client can receive 3 proofs pi_1,pi_2,pi_3 for the 3 time periods above, but they only need to receive the 9.5kB of final validator keys at epoch 256, and a hash-to-field (of size 2 field elements over BW6-761 = 96B) for both the validator keys at 120 and 240. So since they're small, the hash-to-field of the initial epoch of each proof should be stored as well so they don't need to be recomputed.

psivesely:
Just wanted to correct something: the *honest* non-zk-variant groth16 prover will produce the same proof for an instance x, but anyone can randomize any valid proof for x to create another valid proof. E.g., if (A, C, B) in G1^2 x G2 then so is (rA, C, r^(-1)B) for any r in Fp.

CIPs/0017.md
```diff
@@ -0,0 +1,101 @@
...
+
+We propose to add two new RPCs: 
+* `AddProof(proof []byte, firstEpoch uint, lastEpoch uint, versionNumber uint)`
+* `Proofs() ([][]byte)`.
+
+`versionNumber` here refers to the snark circuit version, for more info refer to #upgradeability below.
+
+When an eventual "proving service" generates a new proof it will publish this proof to a full node that it is running, which will then propagate the proof throughout the network.
+
+This `AddProof` function will then utilize the `verificationKey` corresponding to the `versionNumber` and the corresponding validator sets for these epochs to verify the proof. If successful, it will then store the proof in a new database `plumoproofdata`, keyed by it's combination of first and last epoch.
```
pranaymohan, line 30: curious why we need to associate the validator set with the verification key -- would the identifier / version number related to the circuit not be sufficient?

CIPs/0017.md
```diff
@@ -0,0 +1,101 @@
...
+
+`GetPlumoProofMsg` could also request *all* plumo proofs from the receiver, leaving the responsibility of filtering seen proofs with the caller. This is a simpler implementation, but is less scalable as we increase the amount of stored proofs.
```
pranaymohan, line 84: qq -- as a naive reader, I'd be curious, do all proofs have to be stored in perpetuity? Eg. Proof A maps from epoch 0 to epoch 10, proof B from epoch 11-20, incrementing up to epoch 1000. Do we store all 100 proofs?

Wondering why a prover wouldn't aggregate this into one proof from epoch 0-1000?

CIPs/0017.md
```diff
@@ -0,0 +1,101 @@
...
+
+`GetPlumoProofMsg` could also request *all* plumo proofs from the receiver, leaving the responsibility of filtering seen proofs with the caller. This is a simpler implementation, but is less scalable as we increase the amount of stored proofs. 
+
+Specific aspects of the `les` changes may have tradeoffs that I've yet to encounter, so any input and feedback on things to consider there would be helpful.
+
+## Risks
+
+Altering the `eth` protocol introduces some risk of introducing bugs, which could inhibit a nodes ability to connect to the network.
+
+Introducing bugs to mobile client syncing. Possible effects of a faulty implementation include not being able to sync to the network, and introducing a security vulnerability which would allow the node to sync to an attack controlled chain.
```
pranaymohan, line 92: Please expand on this and include how it might be prevented?

Also, is there any risk related to producing snarks? Eg with the mpc setup, or if all provers are malicious

psivesely:
> @kobigurk Can you explain if it's possible with the new circuit that allows "dummy epoch messages" to create a proof covering any number of epochs up to some limit. I think nodes should only need to store about 3 proofs per year passed assuming this is the case that limit is 120 as suggested elsewhere. Again if this is the case then during epoch 256 a node would just be storing proofs for (0, 120), (120, 240), (240, 256).
>
> There is the case to consider where a client goes offline for a long time right at 180 and comes back at 256. They shouldn't have to download and verify 60 headers before they receive the next SNARK, so storing more incremental proofs will help in this scenario.

Replying to my own message, I just reviewed the code for the new circuit and it is indeed the case you can now make a proof from A to B for all B in {A+1,A+2,...,A+120}. So ideally the proofs a node would have saved at time 256 would be (1, 120), (120, 240), (240, 256). The node would also store the hash to field for `(index, validator_keys, max_non_signers)` triple at times 120 and 240. (To clarify something from earlier, I was wrong about max_non_signers, it's part of the instance now rather than the witness to account for epochs where the committee is "underfilled.") A bootstrapping client at time t=256 receives the 3 proofs, the hash-to-field of `(index, validator_keys, max_non_signers)` for t={120,240}, and the `(index, validator_keys, max_non_signers)` for t=256. The client hash-to-fields this final triple themselves (and has the hash-to-field of t=1 hardcoded as part of the genesis data), and then plugs the resulting 4 field elements and 3 proofs into the resulting Groth16 batch verification equation. (I may have misspoke earlier, the hash-to-field of each instance is 2 field elements, but the instance is two triples and each one is hashed individually resulting in 1 field element).

As per the case of a client going offline for a while, I had the idea to simple "restart" them from the genesis block so they could quickly sync. The downside of this is that if a long range attack happens against blocks they've already validated, now they will be affected by it, whereas before they were safe. The upside is the logic around what proofs to send, pruning, etc. stays extremely simple: at time t you always send the same ceil(t/120) proofs, floor(t/120) partial instance hashes, and the current epoch's `(index, validator_keys, max_non_signers)`. There will be some lag, maybe 30min between when epochs roll over, and when a new proof can be created for that day, and for that time you can just send the proof chain and associated data up to t-1 and then the full epoch message for t. As soon as the proof for t rolls in and you validate it, drop the proof for t-1 and its associated data unless t-1 % 0 = 0, in which case save it along with the hash-to-field of `(index, validator_keys, max_non_signers)` for that day. I just talked with @kobigurk and apparently he and some other people were discussing and leaning towards this solution as well for clients who go offline for too long to catch up using reasonable amounts of data (if they have to verify each epoch message themselves). Not sure where cutoff point will be, but we can discuss this more later.

CIPs/0017.md
lucasege: Not naive at all - this is something we've been mulling over as well.
The circuit that generates these proofs has a "hard-coded" limit of how many epochs a proof can span (I believe right now we're targeting 120*), so while we can't roll up to 1000 like in your example, we could roll up to this limit. 

That said, while we can reduce redundant proofs like you say, currently I'm thinking we shouldn't. The reasons being that `lightest` clients may still ask about blocks within epochs between boundaries, so removing intermediary proofs could hurt the performance there. The other reason is that proofs aren't very large and there likely won't be too many, so the tradeoffs seem worth it here.

* we may be able to change this limit dynamically, but that's not super relevant here

CIPs/0017.md
```diff
@@ -0,0 +1,101 @@
...
+## Overview
+
+In order to continue supporting the dream of a seamless mobile sync experience, the lightest sync protocol must be extended to support PLUMO Zk-SNARKs which can quickly verify transitions from epoch X to epoch X + Y, where Y could be on the order of 10s-100s of epochs.
+
+The prover/verifier for plumo proofs have already been implemented, and the exposed go interface is available [here](https://github.com/celo-org/celo-bls-go)
+
+## Goals
+
+- Support adding/storing proofs on full nodes
+- Gossip these proofs to propogate them through the network
```
lucasege, line 17:
```suggestion
- Gossip these proofs to propagate them through the network
```

CIPs/0017.md
```diff
@@ -0,0 +1,101 @@
...
+
+- Date: 2020-07-27
+- Author: @lucasege
+- Status: WIP
+
+## Overview
+
+In order to continue supporting the dream of a seamless mobile sync experience, the lightest sync protocol must be extended to support PLUMO Zk-SNARKs which can quickly verify transitions from epoch X to epoch X + Y, where Y could be on the order of 10s-100s of epochs.
+
+The prover/verifier for plumo proofs have already been implemented, and the exposed go interface is available [here](https://github.com/celo-org/celo-bls-go)
```
lucasege, line 12:
```suggestion
The prover/verifier for plumo proofs have already been implemented, and the exposed go interface is available [here](https://github.com/celo-org/celo-bls-go)

More information on Plumo can be found in the [reference paper](https://docs.zkproof.org/pages/standards/accepted-workshop3/proposal-plumo_celolightclient.pdf)
```

CIPs/0017.md
```diff
@@ -0,0 +1,101 @@
...
+[plumoProof_0, plumoProof_1, ...]
+Reply to GetPlumoProofMsg. The items in the list are plumo proof structs, {proof: []byte, firstEpoch: uint, lastEpoch: uint, versionNumber: uint}.
+This may validly contain no plumo proofs if the GetPlumoProofMsg included all proofs this node is aware of.
+```
+
+This will likely be accompanied by a new `proofFetcher` object to be in charge of fetching these proofs from peers while providing protections against DoS attacks. This feels like overkill, but seems necessary to prevent attacks and maintain state.
+
+The `les` protocol must then also be extended via at least two messages:
+```
+GetCertifiedEpochsValidatorSet - Sends the current highest epoch known to the sender. Server responds with the minimal set of proofs & headers to give the light client the fastest sync time.
```
lucasege, line 60:
```suggestion
GetCertifiedEpochsValidatorSet - Sends the current highest epoch known to the server. Server responds with the minimal set of proofs & headers to give the light client the fastest sync time.
```

CIPs/0017.md
lucasege, line 92:
```suggestion
Introducing bugs to mobile client syncing. Possible effects of a faulty implementation include not being able to sync to the network, and introducing a security vulnerability which would allow the node to sync to an attack controlled chain. Bugs that may be introduced via faulty verification mechanisms will be mitigated via chaos-testing (malicious testers), fuzz testing, e2e tests and strong unit tests.
```

lucasege: There are potential risks inherent to the setup, but these are mitigated via crypto & opsec techniques - will need to ask the folks working on the setup for more details.

Assuming a secure setup and snark construction malicious provers' impact should be mitigated via the same tactics as malicious les servers already are. A malicious prover's proofs should not propagate throughout the network because of the verification step, and their impact on their `les` peers should be eliminated by verification checks once the light client has synced and is receiving blocks from multiple peers. This may also be mitigated earlier via multiple proofs from multiple peers, but I'm not sure on the details yet.

lucasege:
Good points here @psivesely , thanks for correcting and verifying things I've been guessing/approximating at: https://github.com/celo-org/celo-proposals/pull/41#issuecomment-667634302. I'll pull your comments out for cleanliness:

> The verification key is quite short, especially because the actual instance that's verified is a hash of the initial and final epochs. The prover key is the one that's very long.

Good point, I was referring to a point Kobi had mentioned about not wanting to store these on chain and I wrongly assumed it was because they were long. Either way, we've been planning to store VKs within the client, so if they are indeed short that works great.

> You don't need this to verify the proof. If the proof verifies it implies this number is < 1/3 of validators.

Interesting, I've just been referring to the [BLS-go snark API](https://github.com/celo-org/celo-bls-go/blob/master/snark/snark.go) which seems to require this info, would it be possible to create a new `EpochBlock` struct without that field then? Not a big deal since I believe the number is constant barring gov changes, but curious why that field exists if this is the case, or if I'm misunderstanding.

> Zero-knowledge proofs and interactive proofs always use randomness, but we're using a non-zk version of the non-interactive Groth16 proof system, which results in unique proofs given the same initial and final epoch (short of the existence of a fork that merges back at some later point in the epoch message chain).
Just wanted to correct something: the honest non-zk-variant groth16 prover will produce the same proof for an instance x, but anyone can randomize any valid proof for x to create another valid proof. E.g., if (A, C, B) in G1^2 x G2 then so is (rA, C, r^(-1)B) for any r in Fp.

This makes sense, we haven't been working on any assumption of determinism just in case of an "attack" like this - good to hear that those worries weren't unfounded. I don't think there would be any significant optimizations we could implement if we assumed proofs were deterministic anyway so this works for me.

> The client will already have in its state the validator keys for the initial epoch and needs the keys for the final epoch. Keys are one element in group 2 of BLS12-377, so they are 95B each. I think it just makes sense to send the 100 keys instead of the diff because a maximum saving 9.5kB is not worth adding the logic of the diff.

I agree in principle here, but I'll take a look when I'm implementing this to see if this tradeoff is really worth it. I'll also take a look at val set churn, if the churn over ~120 epochs on average is << 100 then it may be worth the diff logic.

lucasege:
>  So since they're small, the hash-to-field of the initial epoch of each proof should be stored as well so they don't need to be recomputed.

lucasege:
> One thing missing from discussion here is batch verification. Batch verification refers to using a modified version of a verification algorithm to verify multiple SNARKs/ signatures/ etc. together faster than they could be independently verified using the standard verification algorithm. If we ensure the proofs we store line up (e.g., (0, 120), (120, 240), (240, 256)), then in our case there is another optimization we can use: following the example above, a client can receive 3 proofs pi_1,pi_2,pi_3 for the 3 time periods above, but they only need to receive the 9.5kB of final validator keys at epoch 256, and a hash-to-field (of size 2 field elements over BW6-761 = 96B) for both the validator keys at 120 and 240. So since they're small, the hash-to-field of the initial epoch of each proof should be stored as well so they don't need to be recomputed.

This is a great point - this was mentioned briefly but I've neglected to add it. Thanks for going deep on this, storing the `hash-to-group` value within each proof makes sense to me as well then, I'll add that to the spec.

CIPs/0017.md
lucasege: This may have been unclear in the wording, trying to make it more clear now.

The version is not related to the validator set, only the VK.

psivesely:
> > You don't need this to verify the proof. If the proof verifies it implies this number is < 1/3 of validators.
> 
> Interesting, I've just been referring to the [BLS-go snark API](https://github.com/celo-org/celo-bls-go/blob/master/snark/snark.go) which seems to require this info, would it be possible to create a new `EpochBlock` struct without that field then? Not a big deal since I believe the number is constant barring gov changes, but curious why that field exists if this is the case, or if I'm misunderstanding.

You do need it. I corrected myself here: https://github.com/celo-org/celo-proposals/pull/41#issuecomment-669291180.

> > Zero-knowledge proofs and interactive proofs always use randomness, but we're using a non-zk version of the non-interactive Groth16 proof system, which results in unique proofs given the same initial and final epoch (short of the existence of a fork that merges back at some later point in the epoch message chain).
> > Just wanted to correct something: the honest non-zk-variant groth16 prover will produce the same proof for an instance x, but anyone can randomize any valid proof for x to create another valid proof. E.g., if (A, C, B) in G1^2 x G2 then so is (rA, C, r^(-1)B) for any r in Fp.
> 
> This makes sense, we haven't been working on any assumption of determinism just in case of an "attack" like this - good to hear that those worries weren't unfounded. I don't think there would be any significant optimizations we could implement if we assumed proofs were deterministic anyway so this works for me.

This only holds for proofs directly received from honest provers. See update on this: https://github.com/celo-org/celo-proposals/pull/41#issuecomment-668743912

> > The client will already have in its state the validator keys for the initial epoch and needs the keys for the final epoch. Keys are one element in group 2 of BLS12-377, so they are 95B each. I think it just makes sense to send the 100 keys instead of the diff because a maximum saving 9.5kB is not worth adding the logic of the diff.
> 
> I agree in principle here, but I'll take a look when I'm implementing this to see if this tradeoff is really worth it. I'll also take a look at val set churn, if the churn over ~120 epochs on average is << 100 then it may be worth the diff logic.

I just remembered client has to do a fair amount of work to decompress the public keys after receiving them in order to actually use them to check block signatures, and that the decompressed keys should be cached by them (pretty sure this is implemented already). To avoid re-doing the decompression work you'd want to implement basically the same logic on the client side to make the diff. So actually now I'm in favor of sending just the diff. I can think of a diff format that for 120 keys is 120B + however many bytes the new public keys are. Lmk if you have a shorter one :D

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+// My rough calculations show this struct is ~439 bytes for a non-batched proof with no val set differences.
+LightPlumoProof struct {
+  Proof             []byte // Serialized proof. Latest estimate from @psi = 383 bytes
+  FirstEpoch        LightEpochBlock // First epoch this proof is over
+  LastEpoch         LightEpochBlock // Last epoch this proof is over
+  VersionNumber     uint // Proof version number, indicates the corresponding verification key to use
+  LastHashToField  []byte  // Hash to field of Last epoch's `snark.EpochBlock` struct. To be used for batch verification.
+                           // If there is no batch verification, this will be empty. May need an additional boolean to signify this.
+                           // If batch verification should be used, newValidators and deletedValidators will be empty.
+  NewValidators     []blsCrypto.SerializedPublicKey // ValSetDiff(FirstEpoch, LastEpoch), added validators
```
asaj, line 90: We use the `NewValidators` and `DeletedValidators` to determine the final validator set. How is that final validator set ordered, and is the hash of the validator set that is proved by the SNARK dependent on this ordering?

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+  DeletedValidators *big.Int // ValSetDiff(FirstEpoch, LastEpoch): bitmap of deleted validators from the `firstEpoch` valset.
+}
+```
+
+The light clients will need to store verification keys corresponding to the existing versions of the snark circuit, as elaborated in #upgradeability. 
+
+There are two proposed extensions for the `les` protocol:
+
+Option 1:
```
asaj, line 99: After coming back to this I'm having trouble understanding why this handles incomplete proof propagation better than approach 2? The only case I've thought of is where this approach is better is when the full node would have proofs that start more than 240 epochs ago, in which case we'd presumably just ask a different full node.

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+  Server responds with `EpocHeadersOrProofs` dependent on `handleLogic`.
+  Both responses are subject to message size limits limiting the number of proofs/headers.
+EpochHeadersOrProofs: Contains the set of proofs & headers to help the client, as requested in `GetEpochHeadersOrProofs`.
+```
+
+Option 2: This option benefits from assumptions on the pattern of proof generation (likely at least one for every ~120 epochs) and by assuming proofs are well distributed (all full nodes have all proofs). The idea here is to pass a bitmap to the `GetEpochHeadersOrProofs` msg that indicates if the client knows the validator set for the past `SIZE_OF_BITMAP` (~240?) epochs. This allows the full node to optimize the set of proofs sent back, while minimizing computation for the client. The main win here would be in a case like this: the client uses a proof to sync to epoch 120, then syncs further to epoch 121, so it knows the validator set of 120 and 121. Then the client goes down for > 120 epochs, and when it comes back the server can figure out to send a proof from 120-240 that we assume is generated, rather than sending 119 epoch headers to verify.
```
asaj, line 114: I'm not sure that this approach actually relies on those assumptions. The light node gives the full node all the information it needs in order to construct a reasonably efficient batch of proofs/headers to sync the light node.

The light client could include in that request, the maximum number of proofs/headers it's willing to verify.

If the full node cannot meet that request, it returns an empty response. If it *can* meet the request, great!

We can have these values be configurable, such that they can be tightened up towards the optimal values of `(CEIL(epochs / 120), 0)`

CIPs/0017.md
asaj: I think what you might be talking about is that we need some way to be able to write an Istanbul validator snapshot for a given epoch after a proof is verified. Currently, I believe the only way for light clients to write a snapshot is to validate an epoch header.

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+
+It may be preferred to combine both options, and send a bitmap with `handleLogic == false` in option 1.
+
+Regardless of the option chosen, the `les` client must then be modified to request proofs & headers during `lightest` sync. This will likely replace the function `getEpochOrNormalHeaders` within `eth/downloader`. This will also involve a new function to update the underlying validator set and blockchain data, which will be built to work with checkpoints as well.
+
+Upon receiving proofs and headers in a response message, the light client will verify headers the same way as before, and will verify proofs via a call into `celo-bls-go`'s `verify` fn.
+
+The verify function requires information on an epoch's validator set, so this will likely require some access changes (private -> public) and exposing more data structures to allow the validator set to be queried and modified upon proof verification.
+
+### Upgradeability
```
asaj, line 134: One thing that might be worth covering is how we will handle gaps in our epoch -> validator set mapping. e.g. what happens if a lightest node wants to check if a transaction was included in a block from a previous epoch for which it doesn't have the validator set? IIRC you solved a similar problem at one rung lower in the validator set > block header > merkle proof hierarchy of the light client protocol.

 think it's fine to not implement the solution for now but it would be nice to know that the message design here will cover that case. (it seems like it does)

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+### Full node changes (RPC and `eth` protocol)
+We propose to add two new RPCs: 
+* `AddProof(proof []byte, firstEpoch uint, lastEpoch uint, versionNumber uint)`
+* `Proofs() ([][]byte)`.
+
+`versionNumber` here refers to the snark circuit version, for more info refer to #upgradeability below.
+
+When an eventual "proving service" generates a new proof it will publish this proof to a full node that it is running, which will then propagate the proof throughout the network.
+
+This `AddProof` function will then utilize the `verificationKey` corresponding to the `versionNumber` and the corresponding validator sets for these epochs to verify the proof. If successful, it will then store the proof in a new database `plumoproofdata`, keyed by the combination of its first and last epoch.
```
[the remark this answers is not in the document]
lucasege: Yes, I've built it that way just missed updating it here

lucasege, line 33:
```suggestion
This `AddProof` function will then utilize the `verificationKey` corresponding to the `versionNumber` and the corresponding validator sets for these epochs to verify the proof. If successful, it will then store the proof in a new database `plumoproofdata`, keyed by the combination of its first and last epoch and the proof's version number. 
```

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+  FirstEpoch uint
+  LastEpoch uint
+  VersionNumber uint
+}
+
+// Metadata for each epoch to construct `EpochBlock` for the snark's verify fn
+// May be unnecessary if we can make assumptions on `MaxNonSigners`
+LightEpochBlock struct {
+  Index         uint // Epoch Index 
+  MaxNonSigners uint // MinQuorumSize()
```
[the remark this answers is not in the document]
psivesely: We have to set a limit for validator set size in the circuit before doing the trusted setup. It's okay to be under that limit, we pad with dummy public keys. So yes this is fine as long as B is less than this limit.

CIPs/0017.md
psivesely: Yes, the ordering is important. Maybe there's already a diff message format, but my idea was to begin with a v byte map, where v is the maximum validator set size, each byte of the map tells you what has happened to the validator key that was in that position last time. If the first byte is 0, then that validator key has remained in the same position. If it's 8, then the 9th key from the last validator set is now the first position. If it's 255, then a new public key not present in the last validator set has taken that position. The rest of the diff message is the list of new public keys, in order of appearance. This gives us v bytes + 95n bytes, where n≤v is the number of new validator public keys.

CIPs/0017.md
psivesely: I'm personally in favor of there being one and only set of proofs and corresponding instance info per epoch that a server will serve. Basically, if clients go offline for a while and get sufficiently far behind, they should just start at the genesis epoch again and verify to the present that way. Maybe after 2 weeks or so downloading the non-SNARK epoch messages becomes more bandwidth/CPU intensive versus just re-bootstrapping with SNARKs. This also very much simplifies proof management these proof serving servers need to do, only storing the minimal number of proofs to get from the genesis to current epoch.

CIPs/0017.md
psivesely: Sorry, general public, to link a Slack thread you can't see, but some discussion on this in https://celo-org.slack.com/archives/CLPE043GT/p1597444936109200

psivesely:
I feel some of my comments regarding my opinions on what proofs to store have been unclear, so to explain my current thinking a bit more clearly I propose storing the following set: {[120i, 120i+j] | i in (0,1,2,...), j in (1,2,...,120)}. We want to support two operations:

- `bootstrap()`: a new ultralight client hardcoded with the genesis state has just joined the network---let's get it the latest committee public keys so it can start transacting. The server sends the client the minimum number of proofs from the genesis validator set to the current committee and a single validator diff. There maybe be a window between when an epoch begins and when the server receives a proof that includes the last epoch message, so in this case the server sends proofs/diff up to the penultimate epoch as well as the last epoch message and its multisig. As noted above, if a client goes offline for ~ a couple weeks it will be more data and CPU efficient for them to re-bootstrap instead of downloading and verifying the non-SNARK epoch messages for that time span. The exact crossover point will depend on how much time has passed since the first epoch, but we can just set some fixed value independent of time where it will be cheaper for at least the next 5 years to re-bootstrap because we'll be on some new protocol version doing different stuff by then anyway.
- `verify_txs([txid_1,...,txid_n])`: an ultralight client just (re-)bootstrapped and would like to recover it's transaction history---assuming it already got a list of txs to/from its address from some index searching service, lets provide it with proofs that all those transactions are valid. Given a vector of transaction identifiers, the server sends the minimal proof chain leading up to each one, minus the proofs that will already be sent because they are required for previous epochs. The server then sends the relevant block headers, their multisignatures (and accompanying bitmaps), and the Merkle tree proof of tx inclusion (if multiple transactions are queried in the same block, the server can reduce communication requirements by not sending the same sibling nodes twice nor nodes the client will be able to compute and cache given previous inclusion proofs it receives). The client batch verifies both the SNARKs and the multisignatures.

CIPs/0017.md
asaj: Hmm I'm not sure I see why that's necessary. The light client can just ask to sync from the latest epoch N for which it has the validator set where e.g. `N % 120 == 0`.

CIPs/0017.md
asaj: When we do validator set diffs for a single epoch transition, i.e. N -> N+1, IIRC the validator set diff implies the order. Given your existing validator set, first you delete all the removed validators, leaving "holes" that are then filled, in order, by the added validators.

However it's not clear to me whether or not that works here

CIPs/0017.md
asaj: I think that's just about transaction inclusion, right? There are other reasons it's useful to be able to fetch and validate headers, so even if there were other ways to prove transaction inclusion the scheme I described above would still be useful

asaj:
> As noted above, if a client goes offline for ~ a couple weeks it will be more data and CPU efficient for them to re-bootstrap instead of downloading and verifying the non-SNARK epoch messages for that time span

I think there's an option you're not considering, which is to download and verify plumo proofs starting, not at the latest epoch for which the light client has the validator set, but from the last epoch for which the full node is able to prove up to the current epoch with plumo proofs only.

asaj:
> I propose storing the following set: {[120i, 120i+j] | i in (0,1,2,...), j in (1,2,...,120)}

This is what I was thinking as well 👍

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+When an eventual "proving service" generates a new proof it will publish this proof to a full node that it is running, which will then propagate the proof throughout the network.
+
+This `AddProof` function will then utilize the `verificationKey` corresponding to the `versionNumber` and the corresponding validator sets for these epochs to verify the proof. If successful, it will then store the proof in a new database `plumoproofdata`, keyed by the combination of its first and last epoch. 
+
+To propagate these proofs throughout the network, we also propose to extend the `eth` protocol with three new message types: 
+```
+NewPlumoProof (0x019)
+[[firstEpoch, lastEpoch, versionNumber], ...]
+Specify one or more new plumo proofs which have appeared on the network.
+This is intended to be honored with a proceeding GetPlumoProofsMsg to request the proof.
```
lucasege, line 40:
```suggestion
This indicates to other peers that a new proof is available by sending out its metadata. Peers are then expected to send a `GetPlumoProofsMsg` to request this proof by its metadata.
```

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+  MaxNonSigners uint // MinQuorumSize()
+}
+
+// All of the data needed by the light client to verify and utilize a proof.
+// This is also crafted to allow for batch verifications using `FirstHashToField`
+// My rough calculations show this struct is ~439 bytes for a non-batched proof with no val set differences.
+LightPlumoProof struct {
+  Proof             []byte // Serialized proof. Latest estimate from @psi = 383 bytes
+  FirstEpoch        LightEpochBlock // First epoch this proof is over
+  LastEpoch         LightEpochBlock // Last epoch this proof is over
```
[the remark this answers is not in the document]
lucasege: Agreed that this seems to imply that capability, but I can't confirm the prover's capabilities here:
@psivesely do you know?

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+```
+// Same as in the `eth` protocol, minimal proof metadata
+ProofMetadata struct {
+  FirstEpoch uint
+  LastEpoch uint
+  VersionNumber uint
+}
+
+// Metadata for each epoch to construct `EpochBlock` for the snark's verify fn
+// May be unnecessary if we can make assumptions on `MaxNonSigners`
```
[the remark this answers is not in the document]
lucasege: Yeah good point, if this is easy for clients then agreed it should be done that way.

Still seems we need the `LastEpoch` struct, though

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+LightEpochBlock struct {
```
lucasege, line 74:
```suggestion
ProofInputs struct {
```

CIPs/0017.md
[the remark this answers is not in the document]
lucasege: Looks like this is addressed by the comment above, agreed this seems possible

CIPs/0017.md
lucasege: Yup exactly, that's my understanding of the necessary changes

CIPs/0017.md
lucasege: Hm good catch, I'll take a look at how the `ValSetDiff` is calculated:
if the way the valset is updated like how you describe, then so long as the `AddedValidators` produced from epoch X to epoch X + 100 maintains the order of each intermediary epoch (I think it should be? If not, wouldn't be that hard to do) then this should be easy to fit into the existing diff format. Of course will think in more detail to verify this

CIPs/0017.md
lucasege: I spoke with @nategraf about this yesterday, one specific case he brought up that I hadn't thought of is the "release process" for new circuits.
Say we upgrade the circuit to v2, this implies a new VK that full nodes must receive in order to validate and accept/propagate further proofs. Depending on how this is implemented, we could be left in a situation where some full nodes have more proofs by nature of having v2 proofs.
The end goal is of course is to allow light clients to verify they are syncing via the fastest available method, so this "proofInventory" is one way, another would be to receive all available proofs from all peers and then follow the shortest path, but that seems even more cumbersome to implement.

CIPs/0017.md
psivesely: It's one thing to have the protocol preserve the order of validators from epoch to epoch (i.e., if validator A is in position 8 in epoch t, then it remains in 8 in epoch t+1 even if the validators in 7 and/or 9 are replaced), but what if A is removed and re-added to the set? We can't also ensure its position is waiting for it. Hence the permutation/ removal bitmap suggestion.

CIPs/0017.md
psivesely: Yes, clarifications below.

CIPs/0017.md
psivesely: Can you give me some examples of other reasons?

psivesely:
> I think there's an option you're not considering, which is to download and verify plumo proofs starting, not at the latest epoch for which the light client has the validator set, but from the last epoch for which the full node is able to prove up to the current epoch with plumo proofs only.

Yeah, I was just over simplifying, but to give a more optimized re-bootstrapping parameter, let t be the latest epoch for which a client has verified the validator public key set (using any combination of SNARKs and multisigs), let t + ∆ be the current epoch, let s = floor(t / 120), and let s' = floor((t+∆)/120). Then when ∆ > 5/4(s'-s+1) + 10 the client should verify s'-s+1 SNARK proofs to re-bootstrap, and when ∆ ≤ 5/4(s'-s+1) + 10 the client should verify ∆ multisigs to re-bootstrap. This is based on verification time, and not on the amount of data that needs to be downloaded. I could try to provide bounds for that as well, but would need an estimate for expected verifier churn.

The bounds above are based on the following assumptions/observations. Batch Groth16 requires n+3 pairings and a n+2 G<sub>1</sub> multiplications and batch BLS requires n+1 pairings. If we let a G<sub>1</sub> exponentiation be the unit cost and assume a pairing costs 4 units, then we arrive at the bounds above.

CIPs/0017.md
lucasege: Agreed this approach doesn't rely on these assumptions, I was just noting that it benefits from them and produces a more optimal flow if they hold.

Maybe this idea of a max # proofs/headers can solve for the potential problem above of full nodes not having every proof? I'll draw out a potential network flow and think through some data structures for a couple solutions and see if I can decide on a solution

CIPs/0017.md
lucasege: Agreed the plan is for the protocol to be accommodating for this kind of request, but I won't prioritize the strict implementation of these requests yet as you mention

psivesely:
The bounds above can also be used to optimize for clients who wish to verify txs/ block headers in multiple epochs that are close together. Instead of providing a proof to each epoch as I suggest above, if any two are close enough in the sense of the bound above then you can just send the SNARKs needed to get to the first one and the diffs, multisigs, bitmaps the client needs to compute and verify the public key set for the next one.

CIPs/0017.md
```diff
@@ -0,0 +1,181 @@
...
+GetProofInventory - Asks the server what proofs it has. Server responds with `[]proofMetadata` of all available proofs.
+GetEpochHeadersOrProofs:
+  Client sends (highestEpoch uint, handleLogic boolean, requestedProofs []proofMetadata)
+    `highestEpoch` is the current highestEpoch the client knows of.
+    `handleLogic`: If true, indicates the server should send back the best set of proofs/headers it has
+                   If false, the server should look at `requestedProofs` and send back proofs requested, with headers to fill gaps as appropriate.
+    `requestedProofs`: Set of proofs' metadata requested by the client
+  Server responds with `EpocHeadersOrProofs` dependent on `handleLogic`.
+  Both responses are subject to message size limits limiting the number of proofs/headers.
+EpochHeadersOrProofs: Contains the set of proofs & headers to help the client, as requested in `GetEpochHeadersOrProofs`.
```
nategraf, line 112: Based on our discussion yesterday, I would suggest breaking this message into `GetHeaders/Headers` (which exists) and `GetProofs/Proofs` and only using this message set when the client wants explicit control over what is returned. Option two would use a separate message to share it's sync status and respond with a mixture of `Headers` and `Proofs` messages controlled by the server.

CIPs/0017.md
```diff
@@ -0,0 +1,181 @@
...
+
+Option 2:
+```
+GetEpochHeadersOrProofs:
+  Client sends (highestEpoch uint, knownValSets *big.Int, maxResponses uint)
+    `knownValSets` is a bitmap of the last ~240 or so epochs that the client knows the val set for
+    `maxResponses` indicates the maximum number of (proofs + headers) the client is willing to accept
+  Server computes the optimal path of proofs & headers that the client can use to sync fastest and send them back using `EpochHeadersOrProofs`.
+EpochHeadersOrProofs:
+  Server sends this back from `GetEpochHeadersOrProofs` with the optimal set of proofs and headers this client can utilize.
```
nategraf, line 135: This can be broken up into `Headers` (which exists) and `Proofs` and sent as a continuous stream of messages. I would suggest this to simplify the network protocol.

CIPs/0017.md
```diff
@@ -0,0 +1,181 @@
...
+
+`peerScore`: This is the total number of proofs and headers needed to sync by this peer. For each peer, subtract the range of each proof in `optimalPath` from `numEpochs`, the result is the number of necessary remaining headers. Add `optimalPath`.length to this for the peer's score. 
+
+Whichever peer has the lowest score indicates the shortest path (best set of proofs) for our client. This could also be short-circuited if the optimal score is provided. Once the best peer is chosen, the client requests the set of proofs and headers from this peer that it has assessed as optimal.
+
+Option 2:
+```
+GetEpochHeadersOrProofs:
+  Client sends (highestEpoch uint, knownValSets *big.Int, maxResponses uint)
+    `knownValSets` is a bitmap of the last ~240 or so epochs that the client knows the val set for
```
nategraf, line 131: I mentioned before, but we should do some thinking about when a bitmap is more or less efficient than a range based representation (i.e. sending a list of `[start, end]` `uint32` pairs that represent that ranges of epochs the light client knows the validator sets for.)

CIPs/0017.md
```diff
@@ -0,0 +1,181 @@
...
+    `maxResponses` indicates the maximum number of (proofs + headers) the client is willing to accept
```
nategraf, line 132: We should generalize this concept to a "verification cost" metric based on the estimates that @psivesely is discussing, plus a cost metric in terms of data usage, which we could include in a unified number of separate it out.

I am not sure whether this max cost function will work as intended in the case where proofs are well distributed. My thinking is that the client will always want to sync (i.e. it will never decide to stay offline because the cost to sync is too high) so if a server cannot fullfil it's request becasuse the provided max cost is too restrictive, we can either assume that another peer will have a lower cost, and "shop around", or if that is a bad assumtion, the client will simply raise the max cost, having wasted a round trip. As a result, I'm not sure we should include this parameter.

CIPs/0017.md
```diff
@@ -0,0 +1,181 @@
...
+
+`proofRange` : For each proof, the difference between `LastEpoch` and `FirstEpoch`'s indices.
+
+`peerScore`: This is the total number of proofs and headers needed to sync by this peer. For each peer, subtract the range of each proof in `optimalPath` from `numEpochs`, the result is the number of necessary remaining headers. Add `optimalPath`.length to this for the peer's score. 
+
+Whichever peer has the lowest score indicates the shortest path (best set of proofs) for our client. This could also be short-circuited if the optimal score is provided. Once the best peer is chosen, the client requests the set of proofs and headers from this peer that it has assessed as optimal.
+
+Option 2:
+```
+GetEpochHeadersOrProofs:
```
nategraf, line 129: We may also need to include information about what proof versions the client is able to accept with this request.

CIPs/0017.md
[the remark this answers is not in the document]
lucasege: Yeah sorry should have removed this.
I was originally wondering if the client could potentially generate this themselves, making this unnecessary, but I believe we need it for the `LastEPoch` still.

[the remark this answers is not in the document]
lucasege: On the "breaking change" point, this wouldn't introduce any potential for breaking changes, the worst that could happen is some wasted bandwidth (sending something the client doesn't need), but I believe that's already been fixed in the design, I'll update this to reflect that.

CIPs/0017.md
```diff
@@ -0,0 +1,165 @@
...
+// Metadata for each epoch to construct `EpochBlock` for the snark's verify fn
+// May be unnecessary if we can make assumptions on `MaxNonSigners`
+LightEpochBlock struct {
+  Index         uint // Epoch Index 
+  MaxNonSigners uint // MinQuorumSize()
+}
+
+// All of the data needed by the light client to verify and utilize a proof.
+// This is also crafted to allow for batch verifications using `FirstHashToField`
+// My rough calculations show this struct is ~439 bytes for a non-batched proof with no val set differences.
```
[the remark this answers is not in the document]
lucasege: Yup it does. Each public key is 96 bytes, so yes the size will be dominated around >4 new validators, the size is roughly `431 + (96 * n)`.

I'm not sure the exact size of the deleted validators bitmap, I'm using 16 bytes as an estimate here which would be 128 bits, but I'm not sure if there is additional necessary overhead or if we need to increase the bits' size.

431 comes from: 383 (proof) + 8 (firstEpochIndex) + 16 (lastEpochBlock) + 8 (versionNumber) + 16 (deletedValBitmap)

CIPs/0017.md
lucasege: Agreed and I'll add this specifically to the design, although I would still like to verify how simply this can be achieved. I agree it simplifies the abstract protocol, but I'm not 100% sure that managing the concurrency of both responses potentially out of sync will be that easy.

CIPs/0017.md
lucasege: Especially if we're going with option 1 (proof inventory) this added parameter is unnecessary.

If we do add it, though, this idea of a verification cost makes sense to me and is an improvement to this idea.

CIPs/0017.md
lucasege: Yes good point

CIPs/0017.md
lucasege: Agreed I'll think more about this and update now that it seems this is the decided (in my mind) path forward

CIPs/0017.md
```diff
@@ -0,0 +1,182 @@
+
+# CIP [0017]: Integrate Plumo Proofs into the lightest sync protocol
+
+- Date: 2020-07-27
+- Author: @lucasege
```
YazzyYaz, line 5: hi @lucasege thank you for submitting this CIP, I would like you to update it to follow the newly activated CIP process which has it's newly updated template here: https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

YazzyYaz (changes requested):

PR #43, opened 2020-09-09
[0019] JWT signer for DID-JWT

daoauth:
JWT signer for DID-JWT

### DIFF
--- a/CIPs/0019.md
+++ b/CIPs/0019.md
@@ -1,0 +1,49 @@
+# CIP [0019]: JWT signer for DID-JWT
+
+- Date: 2020-09-09
+- Author: @daoauth
+- Status: DRAFT
+
+## Overview
+- Now signer interface has 4 methods (signTransaction, signPersonalMessage, getNativeKey, decrypt).
+- And add a one more signer for did-jwt.
+- Because of security, jwt signer and tx signer are using same pk. so it need store in same wallet.
+- It's maybe very useful and strong interface for DIDs. VCs, and more. (eg. HTTP Authentication header type)
+```
+Authorization: Bearer eyJhbGciOiJIUzI1NiIXVCJ9TJV...r7E20RMHrHDcEfxjoYZgeFONFh7HgQ
+```
+
+## Goals
+- Make a simple DID signer.
+- And many DID methods and web authorization use it.
+
+## Proposed Solution
+- did-jwt has a own [simple signer](https://github.com/decentralized-identity/did-jwt/blob/master/src/SimpleSigner.ts).
+- Make a new interface & method with this source.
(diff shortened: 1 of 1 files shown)


CIPs/0019.md
```diff
@@ -0,0 +1,49 @@
+# CIP [0019]: JWT signer for DID-JWT
+
+- Date: 2020-09-09
+- Author: @daoauth
```
YazzyYaz, line 4: hi @daoauth  thank you for submitting this CIP, I would like you to update it to follow the newly activated CIP process which has it's newly updated template here: https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

YazzyYaz (changes requested):

CIPs/0019.md
daoauth: yes. I updated. but However, I couldn't write a few items. And I don't know how to write the Category and Track.

CIPs/0019.md
```diff
@@ -0,0 +1,72 @@
+---
+cip: 0019
```
YazzyYaz, line 2: As per my other point, the CIP number will be changed as only editors can specify one. I'll get back to you on the name later as it's not the most urgent part.

CIPs/0019.md
```diff
@@ -0,0 +1,72 @@
...
+discussions-to: https://github.com/celo-org/celo-proposals/issues/73
+status: Draft
+type: 
+category: 
+created: 2020-09-09
+license: Apache 2.0
+---
+
+## Simple Summary
+JWT encryption is required for web and did support.
```
YazzyYaz, line 14: Can you define the acronym for JWT before using it throughout the doc? It makes for easier reading.

Also what do you mean by "did support"?

Further more, can you explain why JWT is required? Is it for security, for speed, etc.?

CIPs/0019.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Abstract
```
YazzyYaz, line 16: I'd like a definition of did-jwt or a link to it for clarity

CIPs/0019.md
daoauth: DID and JWT definitions have been updated.

CIPs/0019.md
daoauth: did-jwt definitions have been updated.

CIPs/0019.md
YazzyYaz: @prestwich would this be Ring 2?

YazzyYaz: @daoauth for Track, it's "Standards"

PR #44, opened 2020-09-09
[0020] Extensible precompile for hash functions

prestwich:
A CIP for an extensible precompile for hash functions. Starts to specify the precompile itself, as well as a method for adding new hash functions. The intention is to provide a standardized process for upgrading the Celo EVM's cryptographic functionality with minimal developer overhead.

This CIP is currently a WIP. I would love feedback on the contract specification and review process. I have filled in a few common hash functions, as well as one extremely-targeted function (Blake2Xs).

TODOs:
- [x] Specification for Blake2Xs operation mode
- [ ] Benchmarks for gas consumption
- [x] (After review) An implementation of the precompile in go

### DIFF
--- a/CIPs/0020.md
+++ b/CIPs/0020.md
@@ -1,0 +1,127 @@
+# CIP [0020]: Generic hash function support via extensible precompile
+
+- Date: 2020-09-4
+- Author: @prestwich (James Prestwich)
+- Status: WIP
+
+### Terminology
+
+The key words "MUST", "MUST NOT", "SHOULD", and "MAY" in this document are to be interpreted as described in 
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.html).
+
+## Overview
+
+As a descendant of the EVM, Celo supports only a handful of common hash 
+functions. This restricts the design-space of Celo protocols, and 
+
+## Goals
+
+- Provide support for common cryptographic hash functions
+- Expand options for on-chain verification of protocols not specifically 
+designed for EVM compatibility
+    - e.g. Verification of off-chain signatures using Blake2Xs, including Celo's 
+    own epoch commitments within a smart contract.
+    - e.g. Verification of remote Proof of Work from a wide variety of chains
+- Provide a clear route for acceptance and adoption of additional hash functions
+
+## Proposed Solution
+
+Introduce a new precompiled contract at address [TBD] that provides access to 
+a set of pre-determined hash functions. The input to this pre-compile consists 
+of a one-byte selector, a function-specific parameter block, and the preimage. 
+The output of this pre-compile is dependent on the selector, and may be fixed- 
+or variable-length. If the selector is unknown (not on the current 
+compatibility list), or the function-specific, the precompile MUST return an 
+error.
+
+### Proposing an extension
+
+To be eligible for inclusion, a proposed hash function meet the following 
+criteria:
+
+1. There MUST be a standardized specification for the function.
+1. There MUST be an existing high-quality implementation in Go.
+    1. The proposer SHOULD provide an implementation of the precompile 
+    extension.
+1. The function MUST be pure.
+    1. This is intended to disqualify `ethash` and other stateful functions.
+1. The function MUST NOT be memory intensive
+    1. This is intended to disqualify "memory-hard" hash functions and other 
+    potential DoS vectors
+    1. E.g. `scrypt`, `argon`
+1. The function SHOULD NOT be a combination of other hash functions
+    1. Prefer adding each component separately, and building the composition
+    as a Solidity function.
+    1. This is intended to disqualify common PoW-specific combinations.
+    1. E.g. `sha256d`, `X11`,
+
+Once these criteria are met the function must be proposed via a pull request. 
+The pull request SHOULD include a link to an existing implementation and 
+concrete use case for the new hash function. The pull request MUST be named 
+according to the following template: `[CIP-0020 Extension] DigestFunctionName`.
+
+The pull request MUST update this CIP with the following information:
+
+1. A new row in the table below
+    1. The status MUST be set to `Proposed`.
+    1. The selector MUST be marked `TODO`. Selectors will be assigned by the 
+    reviewer, not the proposer.
+1. If any function-specific parameters are required:
+    1. A link to full documentation in a new file in the `CIP-0020/` folder.
+1. If the ouput is not a fixed-length digest:
+    1. A link to full documentation in a new file in the `CIP-0020/` folder.
+
+
+The pull request MUST NOT:
+
+1. Modify any existing table row or text outside of the table.
+1. Conflict with the selector or name of any existing function.
+
+### Review of new hash functions
+
+Reviewers MAY ask the proposer for additional justification or to clarify 
+decisions made with respect to the input and output formats. Reviewers SHOULD 
+accept any proposal that meets all criteria and contains a well-tested 
+implementation.
+
+
+
+### Supported Digest Functions
+
+<!--
+Update using https://www.tablesgenerator.com/markdown_tables
+-->
+
+| Selector | Function Name | Extended Docs | Input Format  | Output Format | Link to specification                                            | Status   |
+|----------|---------------|---------------|---------------|---------------|------------------------------------------------------------------|----------|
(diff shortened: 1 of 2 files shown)


CIPs/0020.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Proposed Solution
+
+Introduce a new precompiled contract at address [TBD] that provides access to 
+a set of pre-determined hash functions. The input to this pre-compile consists 
+of a one-byte selector, a function-specific parameter block, and the preimage. 
+The output of this pre-compile is dependent on the selector, and may be fixed- 
+or variable-length. If the selector is unknown (not on the current 
+compatibility list), or the function-specific, the precompile MUST return an 
+error.
```
nategraf, line 35: Does `return an error` here mean an error indicator value, or does it need to revert the transaction?

CIPs/0020.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Proposing an extension
+
+To be eligible for inclusion, a proposed hash function meet the following 
+criteria:
+
+1. There MUST be a standardized specification for the function.
+1. There MUST be an existing high-quality implementation in Go.
```
nategraf, line 43: Does a `cgo` interface to an implementation in another language (e.g. Rust) count as a Go implementation?

CIPs/0020.md
```diff
@@ -0,0 +1,127 @@
...
+    1. The proposer SHOULD provide an implementation of the precompile 
+    extension.
+1. The function MUST be pure.
+    1. This is intended to disqualify `ethash` and other stateful functions.
+1. The function MUST NOT be memory intensive
```
nategraf, line 48: Could we allow memory hard hash functions given that their gas costs are appropriately tuned? (This obviously might result in the de-facto ban of all memory-hard hash functions except with the most trivial parameters)

nategraf (commented):
We should include some criteria for how to assign the gas cost for a newly introduced hash function. My suggestion would be to use the [validator machine recommendations](https://docs.celo.org/getting-started/baklava-testnet/running-a-validator-in-baklava#hardware-requirements) and create a benchmark harness that can be used to compare the cost of Keccak with the proposed hash function. We could then assign a gas cost relative to the gas cost of Keccak

CIPs/0020.md
prestwich: it means the same as in EIP-2537. The output of the EVM staticcall should be `(false, revertData)`

CIPs/0020.md
prestwich: It seems unwise to offer functions that are designed to consume resources in a strictly resource-constrained environment. Do you have a use case in mind?

CIPs/0020.md
prestwich: The intent here is "someone who's not cLabs maintains this code". Might be better phrased as `there MUST be an existing Go interface to a high-quality implementation`

prestwich:
> We should include some criteria for how to assign the gas cost for a newly introduced hash function. My suggestion would be to use the [validator machine recommendations](https://docs.celo.org/getting-started/baklava-testnet/running-a-validator-in-baklava#hardware-requirements) and create a benchmark harness that can be used to compare the cost of Keccak with the proposed hash function. We could then assign a gas cost relative to the gas cost of Keccak. This would still ignore a lot of variables (e.g. crypto extensions in the CPU, non-linear scaling of runtime with machine specs, difficult to model hash parameters) but it could be a good baseline requirement,

yeah, benching against the recommended hardware is the best approach. I'd use the sha2 precompile as a baseline, rather than the keccak opcode tho.

> and fairly reasonable for functions such as BLAKE2s which are similar to Keccak in interface)

The blake family has the most interface departure from other hash functions because they're so flexible. Ideally we should support keying, personalization, tree hashing mode, etc. Which means their input blob is the most complex, and they need the most work to benchmark. That's why the initial draft includes only the blake variant (blake2Xs) for which we have a concrete and immediate use case

prestwich:
A draft implementation is available here: https://github.com/celo-org/celo-blockchain/pull/1171

prestwich:
> > We should include some criteria for how to assign the gas cost for a newly introduced hash function. My suggestion would be to use the [validator machine recommendations](https://docs.celo.org/getting-started/baklava-testnet/running-a-validator-in-baklava#hardware-requirements) and create a benchmark harness that can be used to compare the cost of Keccak with the proposed hash function. We could then assign a gas cost relative to the gas cost of Keccak. This would still ignore a lot of variables (e.g. crypto extensions in the CPU, non-linear scaling of runtime with machine specs, difficult to model hash parameters) but it could be a good baseline requirement,
> 
> yeah, benching against the recommended hardware is the best approach. I'd use the sha2 precompile as a baseline, rather than the keccak opcode tho.

Update to my thinking here:
I'd use the keccak opcode pricing for the keccak family of hash functions. Sha2 otherwise

CIPs/0020.md
nategraf: I was thinking about verifying the results of an Ethhash, scrypt, or Argon2 PoW output, mostly with applications such as bridging in mind, but also for PoW as rate limiting for dApps. None of these are a pressing need though, and it's quite likely that the gas costs would make not a concern in reality, so we could leave this intentionally unaddressed for now.

prestwich: Ah, those are specifically and intentionally excluded

nategraf (commented):
Giving this a look again, here are some thoughts on what we should also include here:
* Who are the reviewers for the addition of new hash functions? Are they the CIP editors, or a different set of individuals. (e.g. the singleton set {james@prestwi.ch})
* Should new proposals still get a callout and rough consensus at the ACD call? I imagine that it would be much simpler than your average CIP (essentially just a check that they followed the process laid out here and a quick chance to fish for concerns) or do we want to allow for the addition of new hash functions without going through the ACD call at all?
* Test vector and fuzzing should be a required part of the proposal. The largest risk in this process in this whole process is that a bug in the implementation could result in a panic on a bad (e.g. malicious) input vector and that is the main issue I am thinking of here.
* Gas cost assignment is likely the trickiest decision to be made in this process. Having a clear guideline for setting the gas estimate would alleviate this concern. I would recommend the benchmarking process include creating a benchmark permanently added to the test suite. DoS is the second-largest risk to the network in these proposals, and that is the main concern I would like us to address here.

prestwich:
> Who are the reviewers for the addition of new hash functions? Are they the CIP editors, or a different set of individuals. (e.g. the singleton set {james@prestwi.ch})

Reviewers are participants in the ACD call. I don't think editors should necessarily be required to review the technical merit or desirability of the hash function.

> Should new proposals still get a callout and rough consensus at the ACD call?

Yes definitely. Rough consensus still required, but a stronger bias towards acceptance.

> Test vector and fuzzing should be a required part of the proposal.

I can make this explicit 👍 

> Gas cost assignment is likely the trickiest decision to be made in this process

Do we have any prior work here? My conception was to build a small standardized benchmark suite to go along with this CIP. It should accept an interface (that includes test gen) and output iterations/second which can be used to generate gas based on per-second targets

prestwich:
> Should be `https://www.blake2.net/blake2x.pdf`?

Yep thanks 👍

prestwich:
The draft implementation is ready to be run in dev mode

https://github.com/celo-org/celo-blockchain/tree/prestwich/cip-0020

CIPs/0020.md
```diff
@@ -0,0 +1,167 @@
+# CIP [0020]: Generic hash function support via extensible precompile
```
YazzyYaz, line 1: I'd remove this line just to ensure the preamble remains at the beginning of the line. That way when we have a CIP site, it can scrape easily from the preamble.

CIPs/0020.md
```diff
@@ -0,0 +1,167 @@
...
+
+---
+cip: 20
+title: Generic hash function support via extensible precompile
+author: James Prestwich <james@prestwi.ch>
+discussions-to: <URL>
```
YazzyYaz, line 7: Hi @prestwich can you add a url link? It can be either a Github issue called "Discussions for Extensible precompile for hash functions" or something on the forum.celo.org page.

YazzyYaz:
Approving this PR now
@nategraf and @prestwich I'd encourage that we carry the discussions about this CIP in the discussions-to link added in the preamble to preserve the conversation.

YazzyYaz (approved):

PR #45, opened 2020-09-22
doc: Create CIP-0000 Celo Improvement Proposal Process

YazzyYaz:
After discussions with @prestwich @pranaymohan @mcortesi I have drafted a CIP Process proposal to expand our existing process.

This CIP document is based on EIP process with the following important information:

* Unlike the EIP, this CIP Process will adhere to the Celo Community Tenets and the Celo Code of Conduct.

* This document assumes an existing All-Core Dev Call process exists (it needs to be specified, I'm happy to facilitate those as they need to be public calls that include the community).

* This document assumes there's a website for the CIPs such as cips.celo.org. One doesn't exist yet but I can easily add it after since it'll follow the eips.ethereum.org template.

* This CIP process includes the categorization of Rings in the Standard track of proposals.

Please let me know your feedback on this CIP Process proposal.


CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+* An **Informational CIP** describes a Celo design issue, or provides general guidelines or information to the Celo community, but does not propose a new feature. Informational CIPs do not necessarily represent Celo community consensus or a recommendation, so users and implementers are free to ignore Informational CIPs or follow their advice.
+
+It is highly recommended that a single CIP contain a single key proposal or new idea.
+
+A CIP must meet certain minimum criteria. It must be a clear and complete description of the proposed enhancement. The enhancement must represent a net improvement as to be determined by the CIP editors during the PR review process. For more information on the CIP Editor role, see the appropriate section below.
+
+
+# CIP Workflow
+## Shepherding a CIP
+Parties involved in the process are you, the champion or _CIP author_, the CIP editors, and the Celo Core Developers (core developers are the developers working on the Celo client).
```
nambrot, line 41: In general, this document seems to be heavily focused on celo-client/celo-blockchain as the relevant domain, but I would suggest to expand the domain to the Celo Platform as a whole. Of course there are special considerations for celo-blockchain related CIPs (e.g. hard forks)

prestwich: can you be more specific about the scope expansion you're interested in? which repos and packages do you think should fall under the CIP process as opposed to being facilitated by a separate CGP process or not having a formal process?

YazzyYaz: If we are looking at the whole Celo Platform like say Valora, it's probably not a good idea to involve it in the CIP process as it will greatly delay updates and fixes on Valora.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+Building consensus with the Celo community must adhere to the [Celo Code of Conduct](https://celo.org/code-of-conduct) and the [Celo Community Tenets](https://celo.org/community). Your CIP must also be in-line with Celo’s Community Tenets. For example, proposals like changing consensus mechanisms to Proof-of-Work doesn’t follow Celo’s vision because it won’t make Celo accessible to everyone with a smartphone and will therefore not help advance Celo.
+
+## Core CIPs
+For Core CIPs, given that they require Celo’s client implementation to be considered **Final** (see “CIPs Process” below), the blockchain author will need to either provide an implementation for the celo-blockchain client or convince clients to implement your CIP.
```
oneeman, line 50: Should this say "CIP author" instead of "blockchain author"?

YazzyYaz: Good catch!

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+	* ❌ Draft — Reasons for denying Draft status include being too unfocused, too broad, duplication of effort, being technically unsound, not providing proper motivation or addressing backwards compatibility, or not in keeping with the  Celo values and code of conduct.
+* **Draft** — Once the first draft has been merged, you may submit follow-up pull requests with further changes to your draft until such point as you believe the CIP to be mature and ready to proceed to the next status. A CIP in draft status must be implemented to be considered for promotion to the next status (ignore this requirement for core CIPs).
+	* ➡️ Last Call — If agreeable, the CIP editor will assign Last Call status and set a review end date (review-period-end), normally 14 days later.
+	* ❌ Last Call — A request for Last Call status will be denied if material changes are still expected to be made to the draft. We hope that CIPs only enter Last Call once, so as to avoid unnecessary noise on the RSS feed.
+* **Last Call** — This CIP will listed prominently on the  [https://cips.celo.org/](https://cips.celo.org/) website. **NOTE: CIP website is not live yet**
```
oneeman, line 84: This should say "will be listed" (looks like a mistake present in EIP-1 as well)

prestwich:
Can we write down some ways we want this process to differ from the EIP process? I know we already mentioned the code of conduct. What other specific values would we like reflected in our improvement proposal process?

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+# CIP Types
+There are three types of CIP:
+
+* A **Standards Track CIP** describes any change that affects most or all Celo implementations, such as a change to the network protocol, a change in block or transaction validity rules, proposed application standards/conventions, or any change or addition that affects the interoperability of applications using Celo. Furthermore Standard CIPs can be broken down into the following categories. Standards Track CIPs consist of two parts, a design document and implementation.
+	* **Core** - improvements requiring a consensus fork, as well as changes that are not necessarily consensus critical but may be relevant to  “core dev” discussions, and the validator/node strategy.
+	* **Governance** - improvements to the governance proposal to the Celo platform including parameter changes. It should explain the reason for the change, enumerate the smart contract methods that will be called with parameters, and describe the impact and any risks. 
+	* **Networking** - improvements to networking protocol specifications such as devp2p, and proposed improvements like whisper and swarm.
```
asaj, line 28: Assuming this was inherited from EIPs? How relevant is this to Celo given the current usage of whisper and swam?

CIP-0000.md
nambrot: I believe possible domains that could possibly fall under the CIP process that are not `celo-blockchain` are:
- The Celo Core Contracts (Governance, Proof-of-Stake, etc.) Examples include https://github.com/celo-org/celo-proposals/pull/17 or https://github.com/celo-org/celo-proposals/pull/42
- "core-protocol" primitives on the Celo platform (i.e. the identity protocol, including the on-chain `Attestations` contract but also off-chain components such as the `attestation-service` or `ODIS` and `ARKE`.)
- Another example would be offchain-storage, a la [CIP8](https://github.com/celo-org/celo-proposals/blob/master/CIPs/0008.md)

In general, we have a whole team at cLabs who is dedicated towards advancing all non-`celo-blockchain` related protocol efforts which we have tried to define as broadly speaking as building blocks for applications in the Celo ecosystem to build on. Of course the boundary towards applications can get blurry (would a wallet signer interface such as `WalletConnect` count?)

I'd love to hear how y'all think this view of the "Celo protocol" should (or shouldn't) be part of the CIP process

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+* A **Meta CIP** describes a process surrounding Celo or proposes a change to (or an event in) a process. Process CIPs are like Standards Track CIPs but apply to areas other than the Celo protocol itself. They may propose an implementation, but not to Celo’s codebase; they often require community consensus; unlike Informational CIPs, they are more than recommendations, and users are typically not free to ignore them. Examples include code of conduct updates, procedures, guidelines, changes to the decision-making process, and changes to the tools or environment used in Celo development. Any Meta CIP is also considered a Process CIP.
+* An **Informational CIP** describes a Celo design issue, or provides general guidelines or information to the Celo community, but does not propose a new feature. Informational CIPs do not necessarily represent Celo community consensus or a recommendation, so users and implementers are free to ignore Informational CIPs or follow their advice.
+
+It is highly recommended that a single CIP contain a single key proposal or new idea.
```
asaj, line 34: How do you see this wrt smart contract upgrades? The tooling currently supports upgrading all contracts at once rather than individually to ensure compatibility and a clear audit trail

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+A CIP must meet certain minimum criteria. It must be a clear and complete description of the proposed enhancement. The enhancement must represent a net improvement as to be determined by the CIP editors during the PR review process. For more information on the CIP Editor role, see the appropriate section below.
+
+
+# CIP Workflow
+## Shepherding a CIP
+Parties involved in the process are you, the champion or _CIP author_, the CIP editors, and the Celo Core Developers (core developers are the developers working on the Celo client).
+
+Before you begin writing a formal CIP, you should vet your idea. Ask the Celo community first if an idea is original to avoid wasting time on something that will be be rejected based on prior research. It is thus recommended to open a discussion thread on Issues section of the CIP process to do this, but you can also use the Celo Discord and the [Celo Forum](https://forum.celo.org) and engage with the community about your idea.
```
asaj, line 43: Should we link the discord here?

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+	* ❌ Last Call — A request for Last Call status will be denied if material changes are still expected to be made to the draft. We hope that CIPs only enter Last Call once, so as to avoid unnecessary noise on the RSS feed.
+* **Last Call** — This CIP will listed prominently on the  [https://cips.celo.org/](https://cips.celo.org/) website. **NOTE: CIP website is not live yet**
+	* ❌ — A Last Call which results in material changes or substantial unaddressed technical complaints will cause the CIP to revert to Draft.
+	* ➡️ Accepted (Core CIPs only) — A successful Last Call without material changes or unaddressed technical complaints will become Accepted.
+	* ✅ Final (Non-Core CIPs) — A successful Last Call without material changes or unaddressed technical complaints will become Final.
+* **Accepted (Core CIPs only)** — This status signals that material changes are unlikely and Celo client developers should consider this CIP for inclusion. Their process for deciding whether to encode it into their clients as part of a hard fork is not part of the CIP process.
```
asaj, line 88: This may need to read "Celo protocol" if it covers e.g. core smart contract changes

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+Participation in the Celo project and the CIP process is informed by the Celo [Community Tenets](https://celo.org/community) and subject to the  [Code of Conduct](https://celo.org/code-of-conduct).
+
+# CIP Types
+There are three types of CIP:
+
+* A **Standards Track CIP** describes any change that affects most or all Celo implementations, such as a change to the network protocol, a change in block or transaction validity rules, proposed application standards/conventions, or any change or addition that affects the interoperability of applications using Celo. Furthermore Standard CIPs can be broken down into the following categories. Standards Track CIPs consist of two parts, a design document and implementation.
+	* **Core** - improvements requiring a consensus fork, as well as changes that are not necessarily consensus critical but may be relevant to  “core dev” discussions, and the validator/node strategy.
+	* **Governance** - improvements to the governance proposal to the Celo platform including parameter changes. It should explain the reason for the change, enumerate the smart contract methods that will be called with parameters, and describe the impact and any risks.
```
nambrot, line 27: For a Celo on-chain governance proposal, how does this process interact with the formal on-chain process of upvotes, approvals and actual voting? I.e does the equivalent CIP first reach `Final` stage before it should be voted on, or can that already happen in `Draft`?

YazzyYaz: It's a good question, I'm not 100% on what to do about Governance here. I was also thinking maybe it makes sense to move Governance out of it the equation and back to the CGP side.

The one thing I think we want to achieve with CIP process is to target things that need to be standardized, which is easy to follow from the EIP/BIP/PEP process since it's concerned with maintaining and upgrading a standard.

For Governance, we probably need to have a discussion about whether it makes sense to include it in the CIP process the more I think about it.

CIP-0000.md
asaj: +1 to all of these falling under the CIP process.

IIUC, the current proposal does allow for these to be covered

Core contracts could be *core* or *governance* (or both, if the implementation is *core* and the adoption is *governance*)
Similarly, other core primitives would be *core*
Standards for off-chain storage would be, IIUC, *crc*

CIP-0000.md
YazzyYaz: I've listed those as examples inherited from EIPs. I can remove them if needed for clarity. For swarm, EF recently released this: https://swarm.ethereum.org/

Whisper, I've seen working in the wild between different networks. It can be useful to send messages from one node to others.

CIP-0000.md
YazzyYaz: It's a good question, I guess it goes back to whether the smart contract upgrade is a parameter change that's part of defining the Celo standard or if it's something more related to maintenance and refactoring. Perhaps smart contracts need to be part of the CGP instead.

CIP-0000.md
YazzyYaz: Good call, will do.

CIP-0000.md
nambrot: Yeah I can definitely see the case for governance to be outside of the CIP process, IMO it's mostly an implementation detail to possibly achieve some standard, but it's likely never the standard itself. The one nice thing about governance is that it's currently the only process that receives the vote/approval of the Celo community as a whole

CIP-0000.md
nambrot: Yeah I agree that the current proposal can cover these, just some of the language in the proposal (like the above `core developers are the developers working on the Celo client`) might not feel the most inclusive in that regard

CIP-0000.md
prestwich: Generally, the EIPs and BIPs are marked `Final` when they are activated. I expect a governance proposal would be marked `Final` once it has taken affect

CIP-0000.md
prestwich: I'd like to push back on ODIS and off-chain storage being part of the regular CIP process. These services are offered by specific entities, rather than core functionality of a distributed system. The goal of the core CIP process is to define the consensus protocol sufficiently that a second implementation could reach consensus with celo-blockchain. It is not to allow other people to use CLabs services like ODIS. Unless ODIS and CIP8 are critical to the continued function of the chain in a way I don't understand, they ought to be informational CIPs describing standards that use the Celo network (a la ERC20, EIP1967, etc), or not CIPs at all. I would like to exclude as many off-chain services as possible from this protocol. CIP8 certainly shouldn't be subject to the same review process as core CIPs that change consensus behavior. This process is slow, painful, and easily politicized.

While we're on the subject of consensus behavior, we should extend the CIP process to cover any smart contract that implements part of the consensus process or is called outside of the normal user-driven tx processing flow. The fee checks. The core contracts should be considered part of the `celo-blockchain` codebase, as they determine node behavior and can cause consensus failures between client versions

CIP-0000.md
prestwich: > The one nice thing about governance is that it's currently the only process that receives the vote/approval of the Celo community as a whole

This is the thing that is scariest about the process, not what's nice about it :thinking:

CIP-0000.md
YazzyYaz: I fully support @prestwich that off-chain services can be specified under non-Core category. Luckily, this CIP document has that specific category that applies to those off-chain services.

```
* **CRC** - (Celo [Request for Comments](https://en.wikipedia.org/wiki/Request_for_Comments)) application-level standards and conventions, including contract standards such as token standards, name registries, URI schemes, library/package formats, and wallet formats.
```

We can use the CRC category proposed for any service level standards off-chain.

The main goal of "Core" is to be able to specify the Celo Standard, just as "Core" for EIPs specifies the Ethereum Standard, and the main benefits are outlined by James. Long term, I hope it allows the generation of a yellow paper defining the precise specification of the Celo Standard. 

I also agree that for smart contracts, if they're defining behavior that affects how consensus is reached and other behavior like activation of a network upgrade from a specific hard fork block or the activation of a specific gas limit, then they are under the Core Category of the Standards track.

With CRCs, we can include as many off-chain services that need standardization as possible. I can help everyone draft them if needed, and we can discuss them in our debut calls for the CIP process.

asaj: > Unless ODIS and CIP8 are critical to the continued function of the chain in a way I don't understand, they ought to be informational CIPs describing standards that use the Celo network (a la ERC20, EIP1967, etc)

I agree with this wrt CIP8. RE ODIS, I guess it's a question of this off-chain service is considered part of e.g. the attestation protocol primitive.

> While we're on the subject of consensus behavior, we should extend the CIP process to cover any smart contract that implements part of the consensus process or is called outside of the normal user-driven tx processing flow

> I also agree that for smart contracts, if they're defining behavior that affects how consensus is reached and other behavior like activation of a network upgrade from a specific hard fork block or the activation of a specific gas limit, then they are under the Core Category of the Standards track.

I agree with these statements and would suggest extending it to cover the stability and attestation protocols, despite not necessarily affecting consensus or node behavior.

One open question I have is how we treat upgrades to core contracts under the proposed process. Seemingly, proposed changes should go through the CIP process before being merged to master, but also these changes will need to be proposed and adopted via on-chain governance.

CIP-0000.md
mcortesi: Devp2p certainly is part of this

CIP-0000.md
asaj: IMO, the stability protocol, for example, is part of the Celo standard, and changes to the stability protocol (e.g. a different formula for calculating bucket sizes) fit the "core" category.

It's tricky, because adopting such a change will ultimately result in a governance proposal. In our existing process, changes like this would first go through as a CIP, after which they would be merged into master, and then later as a CGP to upgrade all contracts to the latest on `master` (which would effectively bundle a number of CIPs, as well as changes to contracts too small to necessitate a CIP).

Curious to get your thoughts on all of this, and how you were thinking core contract changes fit into the proposed process.

CIP-0000.md
asaj: Ah I guess I meant the usage of whisper/swarm in Celo.

Good to know that whisper more mature, we looked into this a couple years ago and couldn't find compelling examples of it being used in the wild

CIP-0000.md
timmoreton: My view is that one of our primary goals should be to ensure that apps like Valora built on top of the Celo platform can inter-operate, and that when a new feature/service/process is required/desired, it is low-friction to have that adopted as part of the Celo Platform.  In fact, the layer where we have seen most need for a standards process is around things like identity, comment encryption, QR code formats, etc - because the loss of value to the platform in having approaches to these things diverge is huge. Clearly services like ODIS test that "Celo Platform" boundary but these should not be seen as cLabs services in the long term, just like the oracles.  At the very minimum, any contract owned by Governance should be part of the CIP process.

For me, it is a distant secondary goal to have a second clean-room implementation of celo blockchain. Many successful open source projects are "standards by reference implementation" and further implementations reduce development velocity.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+	* **Core** - improvements requiring a consensus fork, as well as changes that are not necessarily consensus critical but may be relevant to  “core dev” discussions, and the validator/node strategy.
+	* **Governance** - improvements to the governance proposal to the Celo platform including parameter changes. It should explain the reason for the change, enumerate the smart contract methods that will be called with parameters, and describe the impact and any risks. 
+	* **Networking** - improvements to networking protocol specifications such as devp2p, and proposed improvements like whisper and swarm.
+	* **Interface** - improvements around client [API/RPC] specifications and standards, and also certain language-level standards like method names and contract ABIs.
+	* **CRC** - (Celo [Request for Comments](https://en.wikipedia.org/wiki/Request_for_Comments)) application-level standards and conventions, including contract standards such as token standards, name registries, URI schemes, library/package formats, and wallet formats.
```
timmoreton, line 30: I think it's good to have an advisory process alongside the CIP process here, but I don't think these are all good examples of the things we would want to make only advisory (see discussion below)

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+In addition to making sure your idea is original, it will be your role as the author to make your idea clear to reviewers and interested parties, as well as inviting editors, developers and community to give feedback on the aforementioned channels. You should try and gauge whether the interest in your CIP is commensurate with both the work involved in implementing it and how many parties will have to conform to it. For example, the work required to implement a Core CIP will be much greater than for a CRC and the CIP will need to be reviewed carefully by the Celo community. Negative community feedback will be taken into consideration and may prevent your CIP from moving past the Draft stage.
+
+Building consensus with the Celo community must adhere to the [Celo Code of Conduct](https://celo.org/code-of-conduct) and the [Celo Community Tenets](https://celo.org/community). Your CIP must also be in-line with Celo’s Community Tenets. For example, proposals like changing consensus mechanisms to Proof-of-Work doesn’t follow Celo’s vision because it won’t make Celo accessible to everyone with a smartphone and will therefore not help advance Celo.
```
timmoreton, line 47: not sure I follow the example here?

timmoreton: also I think this is more about how the community evaluates CIPs. I think it's fine to write a PoW CIP, but the author should expect the community to look think about the Community Tenets when evaluating and considering the proposal.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Core CIPs
+For Core CIPs, given that they require Celo’s client implementation to be considered **Final** (see “CIPs Process” below), the blockchain author will need to either provide an implementation for the celo-blockchain client or convince clients to implement your CIP.
+
+The best way to get client implementers to review your CIP is to present it on a Celo AllCoreDevs call. You can request to do so by posting a comment linking your CIP on an agenda issue for an upcoming call on the Github repository.
```
timmoreton, line 52: I feel we should encourage an async method here- e.g. by posting it in the `#governance` channel on Discord and on the Forum (we should also probably choose which one of those is primary!)

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+The Celo AllCoreDevs call serve as a way for client implementers and stakeholders to do two things. First, to discuss the technical merits of CIPs. Second, to coordinate CIP implementation for network upgrades.
```
timmoreton, line 54: ditto

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+These calls generally result in a “[rough consensus](https://en.wikipedia.org/wiki/Rough_consensus)” around what CIPs should be implemented. Rough consensus, first used by IETF to describe how working groups would proceed, rests on the assumptions that CIPs are not contentious enough to cause a network split and that they are technically sound. It also states that 51% of working group voting on a specific action isn't rough consensus, while 99% is better than rough consensus.
+
+⚠️ The CIPs process and AllCoreDevs call were not designed to address contentious non-technical issues, but, due to the lack of other ways to address these, often end up entangled in them. This puts the burden on core developers to try and gauge community sentiment, which hinders the technical coordination function of CIPs and AllCoreDevs calls. If you are shepherding a CIP, you can make the process of building community consensus easier by making sure that the Github issue thread for your CIP includes or links to as much of the community discussion as possible and that various stakeholders are well-represented.
```
timmoreton, line 58: not sure we need this paragraph

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+_In short, your role as the champion is to write the CIP using the style and format described below, shepherd the discussions in the appropriate forums, and build community consensus around the idea._
+
+## CIP Process
+When we talk about the community from the perspective of the CIP Process, we are talking about all the validators, stakeholders, core-developers, dapp developers, volunteers and active participants involved in the CIP Process.
+
+Rough Consensus, as mentioned in the previous section, is used to measure community sentiment about a particular proposal. Rough Consensus is measured mostly on the All-Core Devs calls, but it can also happen outside of those calls on Github or Celo Forum comments. Rough Consensus will be determined by the Hardfork Coordinator and All-Core Dev Call organizer based on feedback of the network participants during the call.
```
timmoreton, line 65: let's find an off-chain async polling mechanism rather than do this on a call.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+
+# Transferring CIP Ownership
+It occasionally becomes necessary to transfer ownership of CIPs to a new champion. In general, we’d like to retain the original author as a co-author of the transferred CIP, but that’s really up to the original author. A good reason to transfer ownership is because the original author no longer has the time or interest in updating it or following through with the CIP process, or has fallen off the face of the ‘net (i.e. is unreachable or isn’t responding to email). A bad reason to transfer ownership is because you don’t agree with the direction of the CIP. We try to build consensus around a CIP, but if that’s not possible, you can always submit a competing CIP.
+If you are interested in assuming ownership of a CIP, send a message asking to take over, addressed to both the original author and the CIP editor. If the original author doesn’t respond to email in a timely manner, the CIP editor will make a unilateral decision.
+
+# CIP Editors
+The current CIP editors are:
+`* James Prestwich (@prestwich)`
+`* Yaz Khoury (@YazzyYaz)`
```
timmoreton, line 172:
```suggestion
`* Yaz Khoury (@YazzyYaz)`
`* Amy Slawson (@aslawson)`
```

timmoreton: Amy has stewarded a lot of the governance process so far -- @aslawson , happy to be involved?

Also would love non-cLabs representation here, esp someone outside of North America

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+The editors don’t pass judgment on CIPs. We merely do the administrative & editorial part.
+
+# Style Guide
+When referring to a CIP by number, it should be written in the hyphenated form CIP-X where X is the CIP’s assigned number.
+
+# History
+This document was derived heavily from [EIP-1](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md) written by Hudson Jameson and Martin Becze which in turn was derived from [Bitcoin’s BIP-0001](https://github.com/bitcoin/bips) written by Amir Taaki which in turn was derived from  [Python’s PEP-0001](https://www.python.org/dev/peps/) . In many places text was simply copied and modified. Although the PEP-0001 text was written by Barry Warsaw, Jeremy Hylton, and David Goodger, they are not responsible for its use in the Celo Improvement Process, and should not be bothered with technical questions specific to Celo or the CIP. Please direct all comments to the CIP editors.
+
+# Copyright
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
```
timmoreton, line 198: Everything else we do is Apache 2 where possible.

timmoreton (commented):
Thanks for taking the time to progress this! In particular, I really like the clear description of stages and results from each stage.  Things that I think need further discussion:
* (Major) Scope of CIPs and definition of what consistutes the platform. 
* (Minor) Way of notifying, straw polling etc on CIPs. I think we should try and make this an async  process as far as possible.

CIP-0000.md
YazzyYaz: I think, in reference to the organizing what to include on ACD calls, we probably should stick to Github. Each call can have it's own issue link and people can comment on what they want included in the call. If breakout sessions are needed post calls to discuss things more, new issue tickets can be created per breakout session.

Discord would make it super hard to track things for a specific topic as things are more organic. Forum can function the same way as a Github issue ticket as a discussion space for a proposed CIP.

CIP-0000.md
YazzyYaz: Due to avoiding sybil attacks, IETF Working Groups proposing Rough Consensus as a measure during a call has been pretty successful in the past in a lot of the standards being specified.

If we do off-chain polling like say a poll on Discord about people's opinions on somethings, we heavily risk being sybil attacked by someone pushing for a particular direction. I've seen it happy many times, a sudden stream of sockpuppets entering a chat room, mostly voting and preaching a particular proposal.

What we can do is add off-chain polling as a secondary signal, but I am very hesitant to use it to inform consensus.

Consensus has to be rough at least for upcoming hardforks. If there are any disagreements due to majority-minority votes, we are risking network splits.

CIP-0000.md
YazzyYaz: Yeah, happy to add Amy if she wants to be an editor. Also would love some non-cLabs representation if anyone in the community would like to be involved.

CIP-0000.md
YazzyYaz: I can remove the example if needed.

CIP-0000.md
prestwich: the call is an important part of Ethereum's process, in my opinion. It ensures that devs and stakeholders to show up and advocate for their point of view, and that decisions are made in a fixed cadence

CIP-0000.md
YazzyYaz: Interesting. I normally would go with Apache 2 but I'm using Creative Commons here because improvement proposals aren't code but specifications on code. I've been on calls with Software Freedom Law Center about using CC-BY-SA along with a DCO (in case one is worried about anonymous contributors doing malicious things) for blockchain improvement proposals. However, I don't know what's best and I believe this is something we can ask the legal team. If needed, I think we can even reach out to the Software Freedom Law Center to see what would work best for CIPs.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+The CIP proposer, the **author** is responsible for building consensus within the community and documenting whether there is rough consensus on their proposal. 
+
+Because the CIPs are maintained as text files in a versioned repository, their revision history is the historical record of the feature proposal.
+
+Participation in the Celo project and the CIP process is informed by the Celo [Community Tenets](https://celo.org/community) and subject to the  [Code of Conduct](https://celo.org/code-of-conduct).
+
+# CIP Types
+There are three types of CIP:
+
+* A **Standards Track CIP** describes any change that affects most or all Celo implementations, such as a change to the network protocol, a change in block or transaction validity rules, proposed application standards/conventions, or any change or addition that affects the interoperability of applications using Celo. Furthermore Standard CIPs can be broken down into the following categories. Standards Track CIPs consist of two parts, a design document and implementation.
```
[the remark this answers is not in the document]
YazzyYaz: There are cases of Meta tracks that aren't describing standards but a collection of standards, as in the case of referring to Hardfork Implementations from Ethereum (EIPs used) as references.

I'd like to keep Meta track because we will need to use those to refer to all adopted EIPs.

Informational ones aren't used as much, more as a nice-to-have but maybe useful in case they need to be used down the line. I am not strongly opinionated on Informational but Meta is very useful.

prestwich:
Writing down some thoughts from an earlier conversation with @asaj, I think that we should move away from the simple Core/non-Core division in the EIP process. We're establishing some formalism around the organically grown Celo Protocol, so it makes sense that everyone has a different definition of "core." The EIP process is geared towards managing risk from consensus updates. Clearly ODIS and CIP-8 are essential services around the consensus system, but are distinct from it and shouldn't be subjected to the same review and risk management process.

What I'd propose instead is drawing an analogy to [protection rings](https://en.wikipedia.org/wiki/Protection_ring) and incorporating "rings" into our CIP process as follows:

- **Ring 0** - changes that can cause the Celo chain to fail
    - Consensus process changes
    - Tx processing behavior
    - All changes requiring binary updates (hard forks)
    - Core contract changes that can result in permanent partitions
- **Ring 1** - changes that can cause the Celo chain to operate incorrectly
    - Oracle changes (inasmuch as they affect validator rewards)
    - CP-DOTO parameter changes (inasmuch as they affect inflation, etc)
    - New fee tokens
    - Many core contracts (e.g. validation logic)
- **Ring 2** - Standards for essential services not delivered by consensus
    - CIP-8 encryption/decryption scheme
    - ODIS attestation scheme
- **Ring 3** - User-facing standards
    - ERC-20, ERC-777, EIP-1967, etc.

The goal of Core EIPs  is to pass off a concrete technical specification to the core devs for implementation and later activation. Analogously, the goal for Ring 0 and Ring 1 is to hand off the specification for implementation and then to hand off the implemented specification to the CGP process for activation. 

This handoff is not required for Ring 2 or Ring 3, which indicates that they should be subjected to a different process. My contention is that Ring 2 & 3 are best suited to CRCs or informational CIPs rather than normative CIPs. I don't think that we want to mandate updates to ODIS be passed through a formal standards track process with ACD call review right now.

CIP-0000.md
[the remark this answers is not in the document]
timmoreton: OK thanks for the explanation - SGTM

CIP-0000.md
timmoreton:
```suggestion
Building consensus with the Celo community must adhere to the [Celo Code of Conduct](https://celo.org/code-of-conduct). Further, your proposal is much more likely to build consensus if you it aligns with the [Celo Community Tenets](https://celo.org/community) and furthers the Celo project's mission.
```

CIP-0000.md
timmoreton: I'm happy with any of the above, but we'll just need to pick one!

CIP-0000.md
timmoreton: how do they deal with timezones, meeting efficiency, etc? there are certainly individuals who have a lot to contribute but whose English proficiency makes it harder for them to participate in calls than async. howo you feel we can accommodate this and also make this a productive call?

YazzyYaz: 3PM UTC is normally best time that makes everyone happy since it includes most of the user timezones. The calls are mostly for technical specifications and improvements to the Celo Protocol. They will be recorded, they can be transcribed and translated via Google Translate later if needed to allow everyone to review them. Community calls are a huge help for technical specifications of the protocol since debates need a natural way to flow in regards to updating of the client while measuring rough consensus.

nategraf:
I think the protection ring analogy works quite well here. One question I do have is the difference between Ring 2 and 3. I'm having trouble seeing the partition between the CIP8 specification and things like ERC20, as both as opt-in and do not pose risks to users that do not opt to use them. If there isn't a useful distinction, maybe we should collapse Rings 2 and 3 in this proposal.

prestwich:
> I'm having trouble seeing the partition between the CIP8 specification and things like ERC20, as both as opt-in and do not pose risks to users that do not opt to use them

This distinction comes out of the discussion with @nambrot and others above about the meaning of the word "Core" in the CIP process. We wanted some way to express that attestations and other actively managed services are core parts of the Celo Platform in a way that ERC20 is not, while maintaining separate process requirements.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+discussions-to:
+author: Yaz Khoury (@YazzyYaz), James Prestwich (@prestwich), Pranay Mohan (@pranaymohan), Mariano Cortesi (@mcortesi)
+created: 2020-09-22
+license: CC0 1.0
+- - - -
+
+# Abstract
+A Celo Improvement Proposal (CIP) describe standards for the Celo platform, including the core protocol specifications, SDK, and contract standards. It is a design document providing information to the Celo community, or describing a new feature for Celo or its processes or environment, a rationale for the proposal, detailed solution including technical specifications, and, if any, a list of potential risks. 
+
+The CIP proposer, the **author** is responsible for building consensus within the community and documenting whether there is rough consensus on their proposal.
```
nategraf, line 16: We should document what "rough consensus" means, and some examples of how to document it.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+* A **Meta CIP** describes a process surrounding Celo or proposes a change to (or an event in) a process. Process CIPs are like Standards Track CIPs but apply to areas other than the Celo protocol itself. They may propose an implementation, but not to Celo’s codebase; they often require community consensus; unlike Informational CIPs, they are more than recommendations, and users are typically not free to ignore them. Examples include code of conduct updates, procedures, guidelines, changes to the decision-making process, and changes to the tools or environment used in Celo development. Any Meta CIP is also considered a Process CIP.
+* An **Informational CIP** describes a Celo design issue, or provides general guidelines or information to the Celo community, but does not propose a new feature. Informational CIPs do not necessarily represent Celo community consensus or a recommendation, so users and implementers are free to ignore Informational CIPs or follow their advice.
```
nategraf, line 32: I don't quite understand why this exists as part of the CIP process. IIUC, this might be something like a description of how a node operator might execute a key rotation, or some recommendations to users on how to pick a validator for voting. If that it correct, then the forum or documentation website (docs.celo.org) seem like better places for these.

CIP-0000.md
nategraf: I agree that Discord is a great place to float an idea. I think we should recommend the forum as the canonical place to have long-form discussion, and not recommend it be done in GitHub issues. If nothing else, the Forum has a better interface. (Also we want to reduce the proliferation of communication channels as much as possible.)

nategraf (commented):
One of the issues we've seen so far is that PRs will sit for a very long time, and it makes it difficult to track down CIPs. We should make clear recommendations on why to merge the PRs, and ensure that PRs don't need to sit open until the CIP is finalized. (i.e. that the process can be completed in multiple stages)

CIP-0000.md
YazzyYaz: It's linked in the document.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+Before you begin writing a formal CIP, you should vet your idea. Ask the Celo community first if an idea is original to avoid wasting time on something that will be be rejected based on prior research. It is thus recommended to open a discussion thread on Issues section of the CIP process to do this, but you can also use the Celo Discord and the [Celo Forum](https://forum.celo.org) and engage with the community about your idea.
+
+In addition to making sure your idea is original, it will be your role as the author to make your idea clear to reviewers and interested parties, as well as inviting editors, developers and community to give feedback on the aforementioned channels. You should try and gauge whether the interest in your CIP is commensurate with both the work involved in implementing it and how many parties will have to conform to it. For example, the work required to implement a Core CIP will be much greater than for a CRC and the CIP will need to be reviewed carefully by the Celo community. Negative community feedback will be taken into consideration and may prevent your CIP from moving past the Draft stage.
```
nategraf, line 45: We might need to clarify what kind of negative feedback might block a CIP, as I can imagine any significant proposal will have detractors. Especially in the case of CRCs, or other non-core designs, we would not want someone's politically motivated attacks to be able to block something the community wants. It may be something we hash out as we get experience in the process, but it may be worth talking about now, drawing on your experience from the Ethereum community.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+
+The best way to get client implementers to review your CIP is to present it on a Celo AllCoreDevs call. You can request to do so by posting a comment linking your CIP on an agenda issue for an upcoming call on the Github repository.
+
+The Celo AllCoreDevs call serve as a way for client implementers and stakeholders to do two things. First, to discuss the technical merits of CIPs. Second, to coordinate CIP implementation for network upgrades. 
+
+These calls generally result in a “[rough consensus](https://en.wikipedia.org/wiki/Rough_consensus)” around what CIPs should be implemented. Rough consensus, first used by IETF to describe how working groups would proceed, rests on the assumptions that CIPs are not contentious enough to cause a network split and that they are technically sound. It also states that 51% of working group voting on a specific action isn't rough consensus, while 99% is better than rough consensus.
```
YazzyYaz, line 56: cc @nategraf it's defined here.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+	* ➡️ Draft — If agreeable, CIP editor will assign the CIP a number (generally the issue or PR number related to the CIP) and merge your pull request. The CIP editor will not unreasonably deny a CIP.
+	* ❌ Draft — Reasons for denying Draft status include being too unfocused, too broad, duplication of effort, being technically unsound, not providing proper motivation or addressing backwards compatibility, or not in keeping with the  Celo values and code of conduct.
+* **Draft** — Once the first draft has been merged, you may submit follow-up pull requests with further changes to your draft until such point as you believe the CIP to be mature and ready to proceed to the next status. A CIP in draft status must be implemented to be considered for promotion to the next status (ignore this requirement for core CIPs).
+	* ➡️ Last Call — If agreeable, the CIP editor will assign Last Call status and set a review end date (review-period-end), normally 14 days later.
+	* ❌ Last Call — A request for Last Call status will be denied if material changes are still expected to be made to the draft. We hope that CIPs only enter Last Call once, so as to avoid unnecessary noise on the RSS feed.
```
nategraf, line 83: We don't otherwise mention an RSS feed. Is this something we are going to implement? I might advocate for leaning on the Forum instead, as it has a pretty good system for managing notifications.

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+* **Accepted (Core CIPs only)** — This status signals that material changes are unlikely and Celo client developers should consider this CIP for inclusion. Their process for deciding whether to encode it into their clients as part of a hard fork is not part of the CIP process.
+	* ⬅️ Draft — The Core Devs can decide to move this CIP back to the Draft status at their discretion. E.g. a major, but correctable, flaw was found in the CIP.
+	* ❌ Rejected — The Core Devs can decide to mark this CIP as Rejected at their discretion. E.g. a major, but uncorrectable, flaw was found in the CIP.
+	* ✅ Final — Standards Track Core CIPs must be implemented in the Celo client before it can be considered Final. When the implementation is complete and adopted by the community, the status will be changed to “Final”.
+* **Final** — This CIP represents the current state-of-the-art. A Final CIP should only be updated to correct errata.
+
+Other exceptional statuses include:
+* **Active** — Some Informational and Process CIPs may also have a status of “Active” if they are never meant to be completed. E.g. CIP 1000 (this CIP).
```
nategraf, line 95:
```suggestion
* **Active** — Some Informational and Process CIPs may also have a status of “Active” if they are never meant to be completed. E.g. CIP 0000 (this CIP).
```

CIP-0000.md
```diff
@@ -0,0 +1,198 @@
...
+* **Superseded** — A CIP which was previously Final but is no longer considered state-of-the-art. Another CIP will be in Final status and reference the Superseded CIP. A CIP cannot move on from this state.
+
+## What belongs in a successful CIP?
+Each CIP should have the following parts:
+* **Preamble** - [RFC 822 style headers](https://www.w3.org/Protocols/rfc822/) containing metadata about the CIP, including the CIP number, a short descriptive title (limited to a maximum of 44 characters), and the author details. See below for details.
+* **Abstract** - A short (~200 word) description of the technical issue being addressed.
+* **Motivation** (*optional) - The motivation is critical for CIPs that want to change the Celo protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the CIP solves. CIP submissions without sufficient motivation may be rejected outright.
+* **Specification** - The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow implementation in the celo-blockchain client.
```
nategraf, line 106: It should not be restricted the the `celo-blockchain` client here.

CIP-0000.md
prestwich: I think that certain standards ought to be documented alongside the protocol. An example of an informational BIP that should certainly live in the BIPs repo is the BIP32/39/44/49/84 suite. They describe key generation, serialization, and portability formats that are widely used throughout the crypto ecosystem. While you can use Bitcoin without them, your app will not be able to participate in the ecosystem fully without them

nategraf:
We mention rough consensus as the standard for moving a CIP forward and getting it accepted, but it's unclear who makes up the group that needs to reach consensus? Is it the core developers, the CELO holders, Celo users, Dapp developers...

CIP-0000.md
YazzyYaz: The RSS feed is found here: https://github.com/ethereum/EIPs

The code base there can adapted to the CIP process so we can have a cool looking website like cips.celo.org with an RSS feed added

CIP-0000.md
YazzyYaz: removed it for now.

CIP-0000.md
YazzyYaz: For CRCs, my opinion is that, since they don't really affect consensus but just a design specification, even if they receive negative feedback, it's still ok to implement them since they fall in the Ring 2-3 of Standards track. The only reason they shouldn't be accepted by editors is if they're not technically sound but clearly a CRC that isn't pushing the Celo Protocol forward and doesn't align with Celo Tenets and Community Code of Conduct.

YazzyYaz:
@nategraf great questions. For who makes up group for rough-consensus, it's those stakeholders on the call since we are using the ACD as a public working group for the Celo Protocol.

For dealing with unmerged PRs, that will be the job of CIP editors (@prestwich and I), but I also will be updating all CIPs to follow new CIP process and template after we Activate this CIP on a community call. 

As long as CIPs follow the right template format, it's OK to merge them even if they're in Draft form which I'll be doing after we approve the new process.

CIP-0000.md
YazzyYaz: I redefined it as `core developers are the developers working on the Celo Platform`.

CIP-0000.md
YazzyYaz: Resolving since we have moved to a Ring level categorization instead of a core/non-core one.

CIP-0000.md
nambrot: Very interesting. What do you think would be a good heuristic for distinguishing such a informational CIP to a Ring 2 or 3 CIP. Arguably, key derivation and portability are pretty crucial for wallet interoperability at this point, however that certainly wasn't the case when these BIPs were originally authored (I assume). Could they "graduate" in that sense?

CIP-0000.md
```diff
@@ -0,0 +1,195 @@
...
+* **Final** — This CIP represents the current state-of-the-art. A Final CIP should only be updated to correct errata.
+
+Other exceptional statuses include:
+* **Active** — Some Informational and Process CIPs may also have a status of “Active” if they are never meant to be completed. E.g. CIP 0000 (this CIP).
+* **Abandoned** — This CIP is no longer pursued by the original authors or it may not be a (technically) preferred option anymore.
+	* ➡️ Draft — Authors or new champions wishing to pursue this CIP can ask for changing it to Draft status.
+* **Rejected** — A CIP that is fundamentally broken or a Ring 0-1 CIP that was rejected by the Core Devs and will not be implemented. A CIP cannot move on from this state.
+* **Superseded** — A CIP which was previously Final but is no longer considered state-of-the-art. Another CIP will be in Final status and reference the Superseded CIP. A CIP cannot move on from this state.
+
+## What belongs in a successful CIP?
```
nambrot, line 98: Can we move this into the template https://github.com/celo-org/celo-proposals/blob/master/CIPs/template.md ?

CIP-0000.md
nambrot: Very interesting. What do you think would be a good heuristic for distinguishing such a informational CIP to a Ring 2 or 3 CIP. Arguably, key derivation and portability are pretty crucial for wallet interoperability at this point, however that certainly wasn't the case when these BIPs were originally authored (I assume). Could they "graduate" in that sense?

CIP-0000.md
YazzyYaz: I prefer keeping it here if that's ok so everyone has one place to understand the entire process. I will also be updating the template after this Draft is approved and merged because it needs to be expanded as well.

nambrot: Sorry yeah I should have not used the word move, but copy!

YazzyYaz: Makes sense! Will add it as well to the template guide :)

CIP-0000.md
```diff
@@ -0,0 +1,195 @@
...
+	* **Ring 3** - User-facing standards and CRCs (Celo [Request for Comments](https://en.wikipedia.org/wiki/Request_for_Comments)) application-level standards and conventions, including contract standards such as token standards, name registries, URI schemes, library/package formats, and wallet formats. Also includes improvements around client [API/RPC] specifications and standards, and also certain language-level standards like method names and contract ABIs.. Examples include ERC-20, ERC-777, EIP-1967, etc.
+* A **Meta CIP** describes a process surrounding Celo or proposes a change to (or an event in) a process. Process CIPs are like Standards Track CIPs but apply to areas other than the Celo protocol itself. They may propose an implementation, but not to Celo’s codebase; they often require community consensus; unlike Informational CIPs, they are more than recommendations, and users are typically not free to ignore them. Examples include code of conduct updates, procedures, guidelines, changes to the decision-making process, and changes to the tools or environment used in Celo development. Any Meta CIP is also considered a Process CIP.
```
asaj, line 30: nit: "Process CIP" and "Meta CIP" are used here interchangeably

CIP-0000.md
```diff
@@ -0,0 +1,195 @@
...
+
+Building consensus with the Celo community must adhere to the [Celo Code of Conduct](https://celo.org/code-of-conduct). Further, your proposal is much more likely to build consensus if you it aligns with the [Celo Community Tenets](https://celo.org/community) and furthers the Celo project's mission.
+
+## Ring 0 and Ring 1 CIPs
+For Ring 0-1 CIPs, given that they require Celo’s client implementation to be considered **Final** (see “CIPs Process” below), the CIP author will need to either provide an implementation for the celo-blockchain client or convince clients to implement your CIP.
```
asaj, line 49: nit: convince core devs?

asaj: Generally I find it confusing every where we refer to "clients", would suggest we refrain from implying multiple Celo clients until others exist

CIP-0000.md
```diff
@@ -0,0 +1,195 @@
...
+
+The best way to get client implementers to review your CIP is to present it on a Celo AllCoreDevs call. You can request to do so by posting a comment linking your CIP on an agenda issue for an upcoming call on the Github repository.
```
asaj, line 51: For example, consider `s/client implementers/core devs` here and below

CIP-0000.md
```diff
@@ -0,0 +1,195 @@
...
+
+Rough Consensus, as mentioned in the previous section, is used to measure community sentiment about a particular proposal. Rough Consensus is measured mostly on the All-Core Devs calls, but it can also happen outside of those calls on Github or Celo Forum comments. Rough Consensus will be determined by the Hardfork Coordinator and All-Core Dev Call organizer based on feedback of the network participants during the call. 
+
+
+Following is the process that a successful Ring 2-3 CIP will move along:
+```
+[ WIP ] -> [ DRAFT ] -> [ LAST CALL ] -> [ FINAL ]
+```
+
+Following is the process that a successful Ring 0-1 or non-Standards CIP will move along:
```
asaj, line 70: nit: why put rings 0-1 after 2-3?

asaj (approved):
LGTM!

prestwich:
@asaj nit: we can't be sure that other celo-blockchain forks don't exist unless we go inspect all those validator nodes 👀 

that aside, what's preventing us from adopting this policy? It sounds like we have buy-in for the high-level sketch? any outstanding issues to address?

CIP-0000.md
YazzyYaz: It's fairly more complex therefore I added it after since some people are happy with proposing changes on 2-3 ring layer.

PR #57, opened 2020-10-28
[0022] Standard for Injected web3 Provider

daoauth:
## Overview
- Standard to find out the type of web3 provider to support dapps running in web browsers

## Goals
- In the case of web DApps, you can only use [DSRV Celo Extension Wallet](https://chrome.google.com/webstore/detail/celo-desktop-wallet/kkilomkmpmkbdnfelcpgckmpcaemjcdh) or MetaMask Extension wallet, but a standard is required to distinguish them.
- Through these standards, each provider can be identified and transactions can be sent to the Celo blockchain through appropriate operations.

## Proposed Solution
- Ready `ethereum.isMetaMask` property for MetaMask. Through this property, it can be seen that it is a provider injected in MetaMask.
- So, it is suggested that providers injected in Celo have a property with the syntax of `celo[providerName]`.
  * eg. Some injected provider has `celo.isDesktop` property. It's injected by DSRV Celo Extension wallet
  * eg. For DApp browser has `celo.isMobile` too

## Useful Links
- [MetaMask](https://docs.metamask.io/guide/ethereum-provider.html#properties)

### DIFF
--- a/CIPs/0022.md
+++ b/CIPs/0022.md
@@ -14,8 +14,8 @@
 ## Proposed Solution
 - Ready `ethereum.isMetaMask` property for MetaMask. Through this property, it can be seen that it is a provider injected in MetaMask.
 - So, it is suggested that providers injected in Celo have a property with the syntax of `celo[providerName]`.
-  * eg. Some injected provider has `celo.isExtWallet` property. It's injected by DSRV Celo Extension wallet
-  * eg. For DApp browser has `celo.isDexFair` too
+  * eg. Some injected provider has `celo.isDesktop` property. It's injected by DSRV Celo Extension wallet
+  * eg. For DApp browser has `celo.isMobile` too
 
 ## Useful Links
 - [MetaMask](https://docs.metamask.io/guide/ethereum-provider.html#properties)
--- a/CIPs/0022.md
+++ b/CIPs/0022.md
@@ -30,9 +30,9 @@ To increase usability and accessibility by allowing celo DApps to be used not on
 
 
 ## Implementation
-[Celo web signer](https://github.com/dexfair/celo-web-signer)
-[Celo remix plugin](https://github.com/dexfair/celo-remix-plugin)
-[DSRV Celo Extension Wallet](https://github.com/dsrvlabs/celo-extension-wallet)
+- [Celo web signer](https://github.com/dexfair/celo-web-signer)
+- [Celo remix plugin](https://github.com/dexfair/celo-remix-plugin)
+- [DSRV Celo Extension Wallet](https://github.com/dsrvlabs/celo-extension-wallet)
 
 ## Security Considerations
 
--- a/CIPs/0022.md
+++ b/CIPs/0022.md
@@ -1,21 +1,40 @@
-# CIP [0022]: Standard for Injacted web3 Provider
+---
+cip: 0022
+title: Standard for Injacted web3 Provider
+author: @daoauth @NAKsir-melody
+discussions-to:
+status: Draft
+type:
+category:
+created: 2020-10-29
+license: Apache 2.0
+---
 
-- Date: 2020-10-29
-- Author: @daoauth @NAKsir-melody
-- Status: DRAFT
+## Simple Summary
+Standard to find out the type of web3 provider to support dapps running in web browsers
 
-## Overview
-- Standard to find out the type of web3 provider to support dapps running in web browsers
-
-## Goals
+## Abstract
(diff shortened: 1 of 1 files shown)


CIPs/0022.md
```diff
@@ -0,0 +1,21 @@
+# CIP [0022]: Standard for Injacted web3 Provider
+
+- Date: 2020-10-29
+- Author: @daoauth @NAKsir-melody
```
YazzyYaz, line 4: hi @daoauth  thank you for submitting this CIP, I would like you to update it to follow the newly activated CIP process which has it's newly updated template here: https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

YazzyYaz (changes requested):

CIPs/0022.md
daoauth: @YazzyYaz I updated CIP, But the number is wrong, I registered first, but another CIP is using it first. It is suggested to use the pull number. (Like EIP)

CIPs/0022.md
```diff
@@ -0,0 +1,40 @@
+---
+cip: 0022
```
YazzyYaz, line 2: Just to note that we will be changing the CIP number as only editors can assign a number to a CIP

CIPs/0022.md
```diff
@@ -0,0 +1,40 @@
...
+title: Standard for Injacted web3 Provider
+author: @daoauth @NAKsir-melody
+discussions-to: https://github.com/celo-org/celo-proposals/issues/74
+status: Draft
+type:
+category:
```
YazzyYaz, line 8: @prestwich seems to me this is a Ring 4 category due to specifying RPC endpoints that don't really impact the network per se. Any thoughts on it?

CIPs/0022.md
```diff
@@ -0,0 +1,40 @@
...
+To increase usability and accessibility by allowing celo DApps to be used not only in mobile environment but also in desktop environment.
+
+## Specification
+- Ready `ethereum.isMetaMask` property for MetaMask. Through this property, it can be seen that it is a provider injected in [MetaMask](https://docs.metamask.io/guide/ethereum-provider.html#properties).
+- So, it is suggested that providers injected in Celo have a property with the syntax of `celo[providerName]`.
+  * eg. Some injected provider has `celo.isDesktop` property. It's injected by DSRV Celo Extension wallet
+  * eg. For DApp browser has `celo.isMobile` too
+
+## Test Cases
+
```
YazzyYaz, line 30: What are test cases, if any, needed to be specified to ensure this proposal is correct? Do curl commands suffice?

CIPs/0022.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Implementation
+- [Celo web signer](https://github.com/dexfair/celo-web-signer)
+- [Celo remix plugin](https://github.com/dexfair/celo-remix-plugin)
+- [DSRV Celo Extension Wallet](https://github.com/dsrvlabs/celo-extension-wallet)
+
+## Security Considerations
```
YazzyYaz, line 37: Any security considerations we should be thinking about here with the addition of new RPC endpoints?

CIPs/0022.md
daoauth: It was suggested to add a parameter to find out what kind of provider is injected. For example, in the case of MetaMask, the MetaMask's definition is already done through a isMataMask parameter. Likewise, isDesktop means a provider injected by the DSRV wallet. And we can use the functions of MetaMask or DSRV wallet through web3.

CIPs/0022.md
daoauth: Since it is a proposal to use MetaMask and other web extension wallets(eg. DSRV's extension wallet, DApp browser), so The commands of transaction or sign message are done through web3 like general DApps. so there is no security for this proposal.

CIPs/0022.md
```diff
@@ -0,0 +1,40 @@
+---
+cip: 0022
+title: Standard for Injacted web3 Provider
```
prestwich, line 3: This should be `Injected`

CIPs/0022.md
prestwich: Ring 3, I believe

CIPs/0022.md
daoauth: I'll fix it thanks

CIPs/0022.md
prestwich: @YazzyYaz these aren't new RPC endpoints, rather they're properties that providers can expose to help dapps use the provider correctly. 

@daoauth does a corresponding EIP exist? Can we expand the specification to list suggested properties for injected providers?

CIPs/0022.md
prestwich: @daoauth can you update type to `Informational` and category to `Ring 3`? I will merge as a draft after that

prestwich (changes requested):

PR #58, opened 2020-10-29
Add CGP [0014]: cUSD Rewards

ewilz:
A CIP to deliver CELO rewards to holders of cUSD in order to incentivize engagement with valora and the celo platform. The exact details are still being worked out. Here is a list of key parameters as they currently stand:
- Reward: 6% annual
- Scheduled reward delivery: Monthly 
- Asset of payout: CELO 
- Max qualifying balance: 500 cUSD per verified address
- Qualifying criteria: has completed at least 3 attestations (automatic enrollment)

### DIFF
--- a/CGPs/0014.md
+++ b/CGPs/0014.md
@@ -1,0 +1,45 @@
+# CGP [0014]:  cUSD Rewards
+
+- Date: 2020-10-25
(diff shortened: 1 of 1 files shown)


cla-bel (approved):
LGTM

PR #60, opened 2020-10-29
update CIP8 with encryption spec

AlexBHarley:


CIPs/0008.md
```diff
@@ -38,24 +39,22 @@ This http address should be referred to as the root. The expectation here is tha
...
-    - file contents: the data unencrypted, or encrypted to session keys available to the recipients
+  - `dataPath`: `/profile/name` or `/profile/picture`
+  - file contents: the data unencrypted, or encrypted to session keys available to the recipients
 - Releases:
-    - `dataPath`: `/releases/${hashOfRelease}`
-    - file contents: The signature of the key on the hash of the release
+  - `dataPath`: `/releases/${hashOfRelease}`
+  - file contents: The signature of the key on the hash of the release
 
 If no assocation between to the `dataPath` should be leaked, the `dataPath` itself could be encrypted. The recipient would then just have to decrypt all `dataPath`s (which could be available in an index file).
```
nategraf, line 56: Is this still going to be the recommendation? With the "indirect" encryption technique, the only one we are shipping in version 1 of the implementation, this is indeed still a concern. In the "direct" encryption technique, no semantic information about the data is leaked through a plaintext data path. So an updated suggestion here may be to use the "direct" encryption strategy for cases where the dataPath itself leaks non-public information. (Or to think of a way to about including the information in the data path)

nategraf (commented):
One thing that would help future implementers is to standardize on RFC 2119 definitions for "MUST", "MUST NOT", "SHOULD" and "MAY". Example https://github.com/celo-org/celo-proposals/pull/44/files (This is not a strong requirement, but is quite helpful to third parties)

CIPs/0008.md
```diff
@@ -64,14 +63,56 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
 
 More reliance on off-chain data and misunderstanding such reliance could lead to inconsistency or poor usability. Ultimately this is traded off against cost of storing data on-chain or in centralized services.
 
-## Useful Links
+## Specification
+
+### Signatures
+
+All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster. Signatures should be stored along side the payload under `${dataPath}.signature` and to construct these signatures the signature specification [EIP712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) is used. Where it doesn’t make sense to include the payload itself in the EIP712 object we instead include the keccak256 hash hex encoded.
```
nategraf, line 70:
```suggestion
All file contents should always be signed by an authorized signer of the account to prevent tampering by the file hoster. Client may reject any unsigned data received from the storage root. Signatures should be stored along side the payload under `${dataPath}.signature` and to construct these signatures the signature specification [EIP712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) is used. Where it doesn’t make sense to include the payload itself in the EIP712 object we instead include the keccak256 hash hex encoded.
```

nategraf, line 70: Can we copy in the EIP-712 type from your implementation? That's an important part of this spec because clients cannot verify the signatures without it.

CIPs/0008.md
```diff
@@ -64,14 +63,56 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
...
+
+It is worth noting that this does not protect against the adversarial storage provider withholding or returning outdated data. Additionally, this does not protect against the adversary swapping out files. If the order, including the existence, of the data is critical, then the developer should consider falling back to on-chain data, as consensus on the current view of the data is required. A possible partial workaround is to add expiry to the schema.
+
+### Privacy
+
+We commonly expect users to share data only with a set of authorized peers. A profile picture, under the path /profile/picture might be shared only with the user’s friends. Creating and publishing a number of ciphertexts to a related path (e.g. /profile/picture/{recipient address}), each encrypted with a friend's public key would be one way to achieve this. It would ensure data integrity and confidentiality.
```
nategraf, line 78: Just formatting the paths as code strings.
```suggestion
We commonly expect users to share data only with a set of authorized peers. A profile picture, under the path `/profile/picture` might be shared only with the user’s friends. Creating and publishing a number of ciphertexts to a related path (e.g. `/profile/picture/{recipient address}`), each encrypted with a friend's public key would be one way to achieve this. It would ensure data integrity and confidentiality.
```

CIPs/0008.md
```diff
@@ -64,14 +63,56 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
...
+
+A problem with indexing the ciphertexts by recipient is that it reveals publicly every recipient of a given message. In the case of profile pictures, this would make public every “friend” with whom the picture is shared, and in the case of payment requests it would reveal that a request is pending from the user to a given peer. This data may be valuable to attackers looking to craft a spear-phishing attack, and may violate some users’ expectations of privacy.
+
+To solve this we present a scheme that maintains both sender and receiver privacy, to compute the storage path it requires the publisher's private data encryption key and the receivers public data encryption key. The path can be computed as follows:
```
nategraf, line 82: This version of the scheme is only designed to protect recipient privacy. We have a not a bit further down about forward compatibility with sender privacy, but I don't think we should mention it here.
```suggestion
To solve this we present a scheme that maintains recipient privacy, to compute the storage path it requires the publisher's private data encryption key and the receivers public data encryption key. The path can be computed as follows:
```

CIPs/0008.md
```diff
@@ -64,14 +63,56 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
...
+
+```
+label = PRF(ECDH(A, B), A || B || data_path)
+ciphertext path = "/ciphertexts/" || hex(label)
+```
+
+Where A is the sender’s public key, B is the recipient’s public key, ECDH is a non-interactive key exchange requiring the private key for either A or B to evaluate, PRF is the pseudorandom function HMAC-BLAKE2s-256, "ciphertexts" is a cosmetic path URL path prefix, and data pathis the URL path suffix. ECDH produces a secp256k1 elliptic curve point, which is encoded and used as the 64B key for PRF. Elliptic curve points (e.g. public keys and shared secrets) are represented in SEC1-v2 compressed representation. The message input to the PRF consists of the concatenation of the public key of the sender and the public key of the recipient (both in SEC1-v2 compressed representation), and the UTF8 encoding of the data path.
```
nategraf, line 89:
```suggestion
Where A is the sender’s public key, B is the recipient’s public key, ECDH is a non-interactive key exchange requiring the private key for either A or B to evaluate, PRF is the pseudorandom function HMAC-BLAKE2s-256, "ciphertexts" is a cosmetic path URL path prefix, and data path is the URL path suffix. ECDH produces a secp256k1 elliptic curve point, which is encoded and used as the 64B key for PRF. Elliptic curve points (e.g. public keys and shared secrets) are represented in SEC1-v2 compressed representation. The message input to the PRF consists of the concatenation of the public key of the sender and the public key of the recipient (both in SEC1-v2 compressed representation), and the UTF8 encoding of the data path.
```

nategraf, line 89: If I understand our implementation correctly, the line about using the ECDH output as a 64B key is slightly inaccurate. I believe we get back from the node crypto API the 33B compressed form, and should make that the standard. https://github.com/celo-org/celo-monorepo/pull/5091/files#diff-baa5b1d1de4d5708ed55809c3c630d36442ddf5fb84eb26e62ef9f52a0930c3eR10

Note: This does not represent a reduction in entropy.
```suggestion
Where A is the sender’s public key, B is the recipient’s public key, ECDH is a non-interactive key exchange requiring the private key for either A or B to evaluate, PRF is the pseudorandom function HMAC-BLAKE2s-256, "ciphertexts" is a cosmetic path URL path prefix, and data pathis the URL path suffix. ECDH produces a secp256k1 elliptic curve point, which is encoded  in SEC1-v2 compressed format and used as the key for PRF. Elliptic curve points (e.g. public keys and shared secrets) are represented in SEC1-v2 compressed representation. The message input to the PRF consists of the concatenation of the public key of the sender and the public key of the recipient (both in SEC1-v2 compressed representation), and the UTF8 encoding of the data path.
```

CIPs/0008.md
```diff
@@ -64,14 +63,56 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
...
+
+`A || B` is prepended to the data path for future compatibility with sender hiding. If both the sender and receiver upload messages to the same path prefix in order to share an “anonymity pool” and wish to, e.g., mutually share photos under the data path “profile/photo,” then this ensures there’s a unique ciphertext path for A sending a photo to B and B sending a photo to A.
+
+A ciphertext path is easily computable from the data path and public keys of the sender and recipient, and knowledge of either private key. It is infeasible to compute without access to one of the two private keys, and it is also infeasible to determine the sender or recipient. (Although in CIP8 the sender is known because storage roots are not shared.)
```
nategraf, line 93:
```suggestion
A ciphertext path is easily computable from the data path and public keys of the sender and recipient, and knowledge of either private key. It is infeasible to compute without access to one of the two private keys, and it is also infeasible to determine the sender or recipient. (Although in this specification, the sender is known because storage roots are not shared.)
```

CIPs/0008.md
```diff
@@ -64,14 +63,56 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
...
+
+### Encryption
+
+To encrypt a file and share it with recipients, we first encrypt the message using a symmetric key, and then use asymmetric encryption to send that key to all the recipients. This is done as follows. For each file a fresh 16 byte message keyand a fresh 16 byte IVare chosen at random. We encrypt the data (bytes) to produce the file ciphertext, where
```
nategraf, line 97:
```suggestion
To encrypt a file and share it with recipients, we first encrypt the message using a symmetric key, and then use asymmetric encryption to send that key to all the recipients. This is done as follows. For each file a fresh 16 byte message key and a fresh 16 byte IV are chosen at random. We encrypt the data (bytes) to produce the file ciphertext, where
```

CIPs/0008.md
```diff
@@ -64,14 +63,56 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
...
 
-* [The original Metadata CIP](https://github.com/celo-org/CIPs/blob/master/CIPs/0003.md)
-* Links to related CIPs or other documents
+```
+file_ciphertext = IV || E(message_key, IV, data)
+```
 
-## Implementation
+and E is AES-128-CTR. The file ciphertext is stored at `{data_path}.enc` and its corresponding ECDSA signature at `{data_path}.signature`. ECIES encryptions of message keyis stored at the ciphertext path of `{data_path}.key` for each message recipient. Key encryptions should be signed as described above.
```
nategraf, line 103:
```suggestion
and E is AES-128-CTR. The file ciphertext is stored at `{data_path}.enc` and its corresponding ECDSA signature at `{data_path}.enc.signature`. ECIES encryptions of message key are stored at the ciphertext path of `{data_path}.key` for each message recipient. Key encryptions should be signed as described above.
```

CIPs/0008.md
```diff
@@ -64,14 +63,56 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
...
 
-* To follow.
-* Or a link to a PR.
+As an example, if the publisher wants to make their photo available with the data path `/profile/photo.jpg` they will choose a random message key and use it to encrypt the photo bytes. This encrypted photo will be stored at `/profile/photo.jpg.enc`, and a signature over the encrypted photo and path at `/profile/photo.jpg.enc.signature`. For each recipient, the publisher will calculate a ciphertext path, `/ciphertexts/<label>`, using the recipient's public key, for the data path /profile/photo.jpg.key and place the message key ECIES ciphertext there. A signature over the ciphertext and ciphertext path will be produced and stored at `/ciphertexts/<label>.signature`. Note that on that storage provider there is no actual data stored at `/profile/photo.jpg` or `/profile/photo.jpg.key`. Below is a directory structure of this example.
```
nategraf, line 105:
```suggestion
As an example, if the publisher wants to make their photo available with the data path `/profile/photo.jpg` they will choose a random message key and use it to encrypt the photo bytes. This encrypted photo will be stored at `/profile/photo.jpg.enc`, and a signature over the encrypted photo and path at `/profile/photo.jpg.enc.signature`. For each recipient, the publisher will calculate a ciphertext path, `/ciphertexts/{label}`, using the recipient's public key, for the data path `/profile/photo.jpg.key` and place the message key ECIES ciphertext there. A signature over the ciphertext and ciphertext path will be produced and stored at `/ciphertexts/{label}.signature`. Note that on that storage provider there is no actual data stored at `/profile/photo.jpg` or `/profile/photo.jpg.key`. Below is a directory structure of this example.
```

CIPs/0008.md
```diff
@@ -64,14 +63,56 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
...
+
+```
+profile
+├── photo.jpg.enc
+└── photo.jpg.enc.signature
+ciphertexts
+├── <ciphertext label for “/profile/photo.jpg.key”>
+└── <ciphertext label for “/profile/photo.jpg.key”>.signature
```
nategraf, line 113: Just using curly braces to match style with the rest of this document.
```suggestion
├── {ciphertext label for “/profile/photo.jpg.key”}
└── {ciphertext label for “/profile/photo.jpg.key”}.signature
```

CIPs/0008.md
```diff
@@ -64,14 +63,56 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
...
+```
+
+## Useful Links
 
-Add any implementation considerations here. For all proposals going through the governance process, this section should reference the code implementing the proposed change. It’s recommended to get community feedback before writing any code.
+- [The original Metadata CIP](https://github.com/celo-org/CIPs/blob/master/CIPs/0003.md)
```
nategraf, line 118:
```suggestion
- [The original Metadata CIP](https://github.com/celo-org/CIPs/blob/master/CIPs/0003.md)
- [CIP8 encryption client implementation in `celo-monrepo`](https://github.com/celo-org/celo-monorepo/pull/5091)
```

CIPs/0008.md
```diff
@@ -2,15 +2,16 @@
 
 - Date: 2020-05-05
 - Author: @nambrot
```
YazzyYaz, line 4: hi @nambrot  thank you for submitting this CIP, I would like you to update it to follow the newly activated CIP process which has it's newly updated template here: https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-template.md

YazzyYaz (changes requested):

CIPs/0008.md
```diff
@@ -64,14 +67,114 @@ I considered doing storing all data on IPFS, however the tooling around IPFS as
...
+
+To solve this we present a scheme that maintains both recipient privacy, to compute the storage path it requires the publisher's private data encryption key and the receivers public data encryption key. The path can be computed as follows:
+
+```
+label = PRF(ECDH(A, B), A || B || data_path)
+ciphertext path = "/ciphertexts/" || hex(label)
+```
+
+Where A is the sender’s public key, B is the recipient’s public key, ECDH is a non-interactive key exchange requiring the private key for either A or B to evaluate, PRF is the pseudorandom function HMAC-BLAKE2s-256, "ciphertexts" is a cosmetic path URL path prefix, and data path is the URL path suffix. ECDH produces a secp256k1 elliptic curve point, which is in SEC1-v2 compressed format and used as the key for PRF. Elliptic curve points (e.g. public keys and shared secrets) are represented in SEC1-v2 compressed representation. The message input to the PRF consists of the concatenation of the public key of the sender and the public key of the recipient (both in SEC1-v2 compressed representation), and the UTF8 encoding of the data path.
```
nambrot, line 150: Did we resolve the question of using `BLAKE2s` here?

nambrot (approved):

YazzyYaz (approved):

PR #67, opened 2020-11-02
[0022]: Upgrade epoch SNARK data for flexibility, slashability and security

kobigurk:
A CIP for "epoch SNARK data" changes that supports validator sets up to size 150, enables slashing and addresses the "future subcommitte" attack. It's still WIP and is worked on by @nategraf, @lucasege, @mstraka100, @mrsmkl, @pranaymohan and me. I'd love to receive comments and discuss any aspect of it.


CIPs/0022.md
```diff
@@ -0,0 +1,119 @@
...
+
+where `validator_set` is the concatenation of the results of `encode_public_key_le` on each validator public key.
+
+We propose to change it to the following. We first define the parameters `MAX_VALIDATORS = 150` and `SEC_PARAM = 128`. We then specify the new encoding.
+
+```
+Blake2Xs(
+    encode_integer_16_le(epoch number) || 
+    BHHash(
+        bytes_le_to_bits_le(truncate(parent_hash), SEC_PARAM/8) ||
```
oneeman, line 79: The closing parenthesis for truncate is misplaced, this line should instead say:
```
        bytes_le_to_bits_le(truncate(parent_hash, SEC_PARAM/8)) ||
```

kobigurk: Oops, thanks!

CIPs/0022.md
```diff
@@ -0,0 +1,119 @@
...
+    encode_integer_16_le(epoch number) || 
+    BHHash(
+        bytes_le_to_bits_le(truncate(parent_hash, SEC_PARAM/8)) ||
+        encode_integer_32_le(maximum_non_signers) || 
+        pad(validator set, MAX_VALIDATORS, G1_GENERATOR)
+    )
+)
+```
+
+where `validator_set` is the concatenation of the results of `encode_public_key_le` on each validator public key, `parent_hash` is the hash of the parent block header and
```
oneeman, line 86: I think it'd be good to clarify that parent block here refers to the block that's one epoch prior (the parent in the chain of epoch blocks, not in the chain of blocks in general).  It's logical in the context of lightest sync and Plumo, where only epoch blocks are involved, but in general when I hear "parent block" I think of the previous block in the blockchain.

kobigurk: Very good point. What do you think about the new phrasing?

oneeman:
I might be missing something, but in the PRs I see the entropy include two block hashes (`epochHash` and `parentHash`), while in the CIP I only see mention of the parent hash.

CIPs/0022.md
oneeman: Looks good to me.

kobigurk:
> I might be missing something, but in the PRs I see the entropy include two block hashes (`epochHash` and `parentHash`), while in the CIP I only see mention of the parent hash.

You're completely right. Thanks for catching it. Added!

CIPs/0022.md
```diff
@@ -0,0 +1,120 @@
...
+1. It only works correctly when the validator set is at its maximum size. 
+2. As it's designed to be SNARK-friendly, the encoding uses the Bowe-Hopwood SNARK-friendly hash function on the entire elected validator set. As a result, showing a double-signing proof currently requires hashing the entire elected validator set with a non EVM-friendly hash, having too high gas costs.
+3. It's vulnerable to a "future subcommittee" attack, where an attacker that compromises enough validators can pre-sign encodings with future epoch numbers and use them when their compromised subcommittee is elected in the future.
+
+This CIP upgrades the encoding to address all three topics.
+
+## Motivation
+
+- Support validator sets which are not full up to a specified maximum.
+- Create the necesasry infrastructure for on-chain slashing in the case of double-signing.
```
pranaymohan, line 44:
```suggestion
- Create the necessary infrastructure for on-chain slashing in the case of double-signing. 
```

How about a bullet for the future subcommittee attack?

CIPs/0022.md
```diff
@@ -0,0 +1,120 @@
...
+We define a few helper functions:
+* `encode_integer_X_le` - encodes an unsigned_integer having X bits to bits in a little-endian order.
+* `encode_public_key_le` - encodes a BLS12-377 point to a compressed form in bits - the little-endian bits of the x coordinate and a single bit whether y is the greatest of the two possibilities.
+* `bits_be_to_bytes_le` - converts big-endian bits to little-endian bytes.
+* `bytes_le_to_bits_be` - converts little-endian bytes to big-endian bits.
+* `bytes_le_to_bits_le` - converts little-endian bytes to little-endian bits.
+* `BHHash` - Bowe-Hopwood hash, as specified in the [Zcash protocol specification](https://github.com/zcash/zips/blob/master/protocol/protocol.pdf), on the Edwards curve defined over the same base field as BLS12-377, as specified in the [Zexe paper](https://eprint.iacr.org/2018/962.pdf).
+* `Blake2Xs` - the Blake2s-based variant of Blake2X, defined in the [Blake2X paper](https://blake2.net/blake2x.pdf).
+* `pad(array, n, padding_element)` - enlarge the array to size `n`, adding `padding_element` in the new places.
+* `truncate(array, n)` - truncate the array to the first `n` elements.
```
pranaymohan, line 57:
```suggestion
* `pad(array, n, padding_element)` - enlarges the array to size `n`, adding `padding_element` in the new places.
* `truncate(array, n)` - truncates the array to the first `n` elements.
```

CIPs/0022.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Test Cases
+
+* **WIP** - We provide the [plumo-prover](https://github.com/celo-org/plumo-prover) codebase, which consumes data from a Celo deployment and generates Plumo proofs. It will be run on private testnets before the test period and on Alfajores and Baklava during the test period.
+* **WIP** - We provide a proof-of-concept implementation of a slashing contract that utilizes [CIP20](https://github.com/celo-org/celo-proposals/compare/prestwich/cip-0020) for Blake2Xs and [CIPX]() for BLS12-377 operations.
+
+## Implementation
+
+* **WIP** - The changes are implemented as PRs on [celo-blockhain](https://github.com/celo-org/celo-blockchain) ([#1158](https://github.com/celo-org/celo-blockchain/pull/1158)), [celo-bls-snark-rs](https://github.com/celo-org/celo-bls-snark-rs) ([#183](https://github.com/celo-org/celo-bls-snark-rs/pull/183)) and [celo-bls-go](https://github.com/celo-org/celo-bls-go).
```
pranaymohan, line 107:
```suggestion
* **WIP** - The changes are implemented as PRs on [celo-blockchain](https://github.com/celo-org/celo-blockchain) ([#1158](https://github.com/celo-org/celo-blockchain/pull/1158)), [celo-bls-snark-rs](https://github.com/celo-org/celo-bls-snark-rs) ([#183](https://github.com/celo-org/celo-bls-snark-rs/pull/183)) and [celo-bls-go](https://github.com/celo-org/celo-bls-go).
```

CIPs/0022.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Deployment Considerations
+
+* This CIP can in fact be deployed as a soft-fork, in the sense that only validators MUST upgrade. Full nodes and light-clients do not verify the correctness of this field. Nevertheless, we suggest to deploy it in a hard-fork to minimize the amount of upgrades validators have to perform.
+
+* The CIP puts hard upper-bound on the number of validators to be 150.
```
pranaymohan, line 113:
```suggestion
* The CIP puts a hard upper-bound on the number of validators to be 150.
```

pranaymohan (commented):
looks great! mostly nits

CIPs/0022.md
```diff
@@ -0,0 +1,127 @@
...
+
+
+## Rationale
+
+A single Plumo proof can support a fixed amount of validators and epochs. We believe that 150 validators and 130 epochs (~= 130 days) will satisfy Celo's needs for the next 2 years.
+
+Moving the epoch number outside the `BHHash` allows for an efficient on-chain double-signing proof - this proof doesn't need the information inside the `BHHash`, it's just required to show there are two aggregate signatures by a super majority on the same epoch number and distinct opaque outputs of `BHHash`.
+
+Adding the current and parent block header hashes prevents the "future subcommittee" attack since it adds an unpredictable value that the attacker can't predict in advance and therefore can't pre-sign, and establishes a chain of epoch headers. Specifically, the header contains the state hash, which in turn contains the result of the Celo randomness beacon. We take only `SEC_PARAM/8` bytes, since, assuming the bytes are sufficiently unpredictable, an attacker would need to pre-sign around `2^128` different messages in order to succeed.
```
oneeman, line 105: Some more details about the future subcommittee attack would be good.  For one thing, I had a hard time finding information about it by googling it.  But, also, at first I thought since the entropy includes two header hashes, we only need to take `SEC_PARAM/16` bytes from each.  Then I thought maybe not, because the attack could perhaps be done during one epoch, so the parent hash could be know.  But then it wasn't clear to me why there should be `SEC_PARAM/8` bytes from the parent hash.  Is it to ease the SNARK by having the two hashes be the same length?  Regardless, I guess my point is just that more detail would be good on these points, both about the attack (can be a link to details elsewhere), and about the requirements for the chain of epoch headers that are behind the choice to include `SEC_PARAM/8` for the parent hash too.

kobigurk:
@oneeman thanks a lot for the comments. I added some text to expand on it. Let me know if it makes sense.

nategraf (commented):
Is there a particular reason to include constants (e.g. the cofactor for BLS12-377) that are part of the BLS12-377 specification? If not, you might make this slightly more concise by leaving them out and pointing to a reference for the curve parameters.

kobigurk:
> Is there a particular reason to include constants (e.g. the cofactor for BLS12-377) that are part of the BLS12-377 specification? If not, you might make this slightly more concise by leaving them out and pointing to a reference for the curve parameters.

Good question. I aimed for the CIP to be as self-contained as possible. So for example, there are sometimes "cofactor clearing" methods that use a multiple of the cofactor, which is still fine for security purposes, but results in a different point. Another reason that I had is that it's just not very well documented elsewhere besides code.

CIPs/0022.md
```diff
@@ -0,0 +1,146 @@
+# CIP [0022]: Upgrade epoch SNARK data for flexibility, slashability and security
```
YazzyYaz, line 1: Hi @kobigurk I'd remove this title as the preamble will have all the information needed. Preamble will scraped later with a CIP website that I will integrate later.

CIPs/0022.md
```diff
@@ -0,0 +1,146 @@
...
+
+---
+cip: 22
+title: Upgrade epoch SNARK data for flexibility, slashability and security
+author: Kobi Gurkan <me@kobi.one>
+discussions-to: <URL>
```
YazzyYaz, line 7: Can you add a discussions-to url? It can be as easy as opening a Github ticket that is called: "Discussions to  Upgrade epoch SNARK data for flexibility, slashability and security" and linking the URL here. Or, if you prefer, you can make one on the forum.celo.org.

YazzyYaz (approved):

PR #77, opened 2020-11-12
Update CGP13 with proposalID 13

nambrot:


CGPs/0013.md
```diff
@@ -3,7 +3,7 @@
 - Date: 2020-10-16
 - Author(s): @nambrot
 - Status: DRAFT
```
m-chrzan, line 5: I don't remember off the top of my head, do we update proposal status to something like "PROPOSED" during the governance process?

m-chrzan (approved):

PR #84, opened 2020-11-17
[0026] Precompile for getting BLS public key

mrsmkl:
First draft.


CIPs/cip-draft-bls-publickey.md
```diff
@@ -0,0 +1,68 @@
+---
+cip: 
+title: Precompile for getting BLS public key for a given validator in a historical block
+author: Sami Mäkelä (@mrsmkl)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): <Ring 0, 1, 2, 3>
```
kobigurk, line 8: It's ring 0 since it's a consensus change

CIPs/cip-draft-bls-publickey.md
```diff
@@ -0,0 +1,68 @@
...
+
+Adding a precompile that returns the BLS public key of a given validator in a given block.
+
+## Abstract
+
+Returns the BLS public key of a validator in the same way as the existing precompile for getting the validator address.
+
+## Motivation
+
+Currently it's not possible to get the validator BLS public key from the chain. However the keys are needed to implement slashing in the case of bad epoch snark data (see CIP-20).
```
kobigurk, line 24:
```suggestion
Currently it's not possible to get the validator BLS public key used for consensus from the chain. It's only possible to get the current BLS key that would be active in the next epoch. However the keys are needed to implement slashing in the case of bad epoch snark data (see CIP-20).
```

CIPs/cip-draft-bls-publickey.md
```diff
@@ -0,0 +1,68 @@
...
+
+To be implemented.
+
+## Implementation
+
+Branch mrsmkl/precompile-bls-key
+
+## Security Considerations
+
+The precompile is veru similar to an existing one, so there shouldn't be too many security implications. Mostly have to be sure that all the edge cases are handled properly and to make sure there's no non-deterministic behavior.
```
kobigurk, line 64:
```suggestion
The precompile is very similar to an existing one, so there shouldn't be too many security implications. Mostly have to be sure that all the edge cases are handled properly and to make sure there's no non-deterministic behavior.
```

CIPs/cip-draft-bls-publickey.md
```diff
@@ -0,0 +1,68 @@
+---
```
kobigurk, line 1: We need to mention gas costs somewhere.
As for the methodology to derive it, one possibility is, paraphrasing from @prestwich:
* We need to choose a "gas per second" number
* Perform benchmarks and set the gas based on that

Another suggestion by @oneeman is relative pricing - take an existing precompile, benchmark your new precompile and the existing one, and price using the ratio.

oneeman: An approach combining them would be to take some existing precompiles and their current gas costs, benchmark them and calculate their implied gas per second numbers, then take the mean or median of those and price using that.

CIPs/cip-draft-bls-publickey.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Specification
+
+The number of precompile is 0x... The input is interpreted in the same way as precompile for getting the validator address: it is 64 bytes, first 32 bytes are an integer representing the index of validator to get, and the second 32 bytes are an integer representing the block to access. These input values have the same conditions as the precompile for getting validator address: index must be smaller than the number of validator for the given block, and block number must be smaller than current block number.
+
+Returns the BLS public key in uncompressed format as 192 byte string. In the case of Celo, the BLS public key is a point in the BLS12-377 elliptic curve defined over Fp2 (G2). A value of Fp2 is represented as two numbers smaller than the base field modulus. These numbers are encoded as 48 byte big endian strings (function `BE`). So if we have a point (a+bi,c+di), it is serialized as `BE(a) || BE(b) || BE(c) || BE(d)`.
```
prestwich, line 30: any thought given to returning 256 bytes (4 64-byte numbers) to maintain evm word alignment? with 48-byte numbers, the caller will always have to read the middle of some memory words and mask appropriately

CIPs/cip-draft-bls-publickey.md
prestwich: I know the matter team has been doing this for various Ethereum precompiles lately. There's a bit of a difference here, as all ethereum precompiles are pure computation while Celo has precompiles that access state (and thus should be considerably more expensive than benching might imply)

prestwich: note: state access costs scale with state size. so any precompile that accesses state should be repriced regularly

CIPs/cip-draft-bls-publickey.md
mrsmkl: It's probably best to return what's most efficient to handle in the EVM, I just described what's happening in the rust code for now.

prestwich: using the 64-byte format instead of the 48-byte format would also match the EIP2537/2539 format and allow you to directly pass the returned block to the precompile (if you manage memory carefully). I think I'd strongly recommend switching to 64-byte

YazzyYaz (approved):

PR #87, opened 2020-11-18
Remove blake2xs from cip20 specification

prestwich:
This PR removes blake2Xs from the CIP20 spec. Reasoning: existing implementations are generally of low quality, and it can be instantiated from successive applications of blake2s.


YazzyYaz (approved):
LGTM

PR #102, opened 2020-12-01
CGP 0016

alecps:


nambrot (approved):
An important CGP, thanks for driving, the voters on Celo will thank you!

YazzyYaz (approved):

PR #109, opened 2020-12-03
splits cgp16 into cgp16 and cgp18

alecps:
Based on feedback from the governance call, I've split these proposals into two.

### DIFF
--- a/CGPs/0018.md
+++ b/CGPs/0018.md
@@ -1,4 +1,4 @@
-# CGP [0018]: Update baselineQuorumFactor and Referendum Stage Duration
+# CGP [0018]: Update referendumStageDuration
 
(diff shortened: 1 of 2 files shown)


CGPs/0018.md
```diff
@@ -0,0 +1,35 @@
+# CGP [0018]: Update baselineQuorumFactor and Referendum Stage Duration
```
nambrot, line 1: You probably want to also split the titles

alecps: Great catch 😂

nambrot (approved):

prestwich:
@nambrot feel free to merge if you think it's ready. I don't have context for the changes

PR #111, opened 2020-12-09
CIP 30 & 31 - add BLS EIPs as new CIPs

prestwich:
@YazzyYaz please assign numbers and I'll update with discussion issues :)


prestwich:
@YazzyYaz want to check and see if these are ready for merge now :)

YazzyYaz (approved):

PR #114, opened 2020-12-10
update cip21 to expand specification

prestwich:
I tried to expand and update the CIP to clarify the implementation and expected behavior. Please review to ensure I got it right 👍

### DIFF
--- a/CIPs/cip-0021.md
+++ b/CIPs/cip-0021.md
@@ -17,7 +17,7 @@ The key words "MUST", "MUST NOT", "SHOULD", and "MAY" in this document are to be
 
 ## Simple Summary
 
-Validators are paid with rewards to maintain the network. Their pay rate is lowered if they were failed to sign attestations for a fixed duration, called the lookback window. This proposal suggests adjusting the lookback window on-chain via community governance.
+Validators are paid with rewards to maintain the network. Their pay rate is lowered if they failed to sign blocks for a fixed duration (the "lookback window"). This proposal suggests adjusting the lookback window on-chain via community governance.
 
 ## Abstract
 
--- a/CIPs/cip-0021.md
+++ b/CIPs/cip-0021.md
@@ -10,21 +10,26 @@ created: 2020-09-25
 license: Apache 2.0
 ---
 
+### Terminology
+
+The key words "MUST", "MUST NOT", "SHOULD", and "MAY" in this document are to be interpreted as described in
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.html).
+
 ## Simple Summary
 
-Validators are paid with rewards to maintain the network. Their pay rate is lowered if they were considered to be down for a fixed duration, called lookback window. This proposal suggests moving the lookback window on chain, to be adjusted by the community via governance proposals.
+Validators are paid with rewards to maintain the network. Their pay rate is lowered if they were failed to sign attestations for a fixed duration, called the lookback window. This proposal suggests adjusting the lookback window on-chain via community governance.
 
 ## Abstract
 
 Each epoch, the Celo protocol elects validators trusted to sign mined blocks and, in return, they recieve Celo gold at the end of the epoch. What happens if an elected validator fails to sign the blocks it's been assigned? During an epoch, the Celo protocol will keep track of each validator's score, as a percentage, and the rewards paid will be proportional to their score.
 
-The purpose of this score is to make sure the validator is doing the work it's been assigned. A validator's score can be penalized for various reasons. This proposal is focused on the uptime metric.
+The purpose of this score is to make sure the validator is doing the work it has been assigned. A validator's score can be penalized for various reasons. This proposal is focused on the uptime metric.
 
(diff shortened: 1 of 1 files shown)


andres-dg (approved):

CIPs/cip-0021.md
```diff
@@ -10,21 +10,26 @@ created: 2020-09-25
...
 
+### Terminology
+
+The key words "MUST", "MUST NOT", "SHOULD", and "MAY" in this document are to be interpreted as described in
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.html).
+
 ## Simple Summary
 
-Validators are paid with rewards to maintain the network. Their pay rate is lowered if they were considered to be down for a fixed duration, called lookback window. This proposal suggests moving the lookback window on chain, to be adjusted by the community via governance proposals.
+Validators are paid with rewards to maintain the network. Their pay rate is lowered if they were failed to sign attestations for a fixed duration, called the lookback window. This proposal suggests adjusting the lookback window on-chain via community governance.
```
mcortesi, line 20: "attestations" might be confusing since we have the 'attestions service', i would use 'blocks'

prestwich: Updated to ` blocks` and did some minor cleanup around it

prestwich:
Force push was to rebase onto latest master

CIPs/cip-0021.md
```diff
@@ -35,24 +40,71 @@ The core devs have been recieving complaints from the validators saying this win
...
+    ...
+  }
 ```
 
-This way, the parameter will be handled by the BlockchainParameters contract and it will be fetched with a static view call from the node (layer 1). This would mean a hard fork, replacing any reference to the parameter with a contract call.
+Starting at `HARDFORK_BLOCK`, block creation and validation MUST read the lookback window from chain state by calling the `BlockchainParameters` contract, rather than reading chain paramaters. Reward calculations are unaffected.
+
+If the return value is outside the safe window, the node MUST clamp the return value to that range such that `MIN_SAFE_LOOKBACK_WINDOW <= window <= MAX_SAFE_LOOKBACK_WINDOW`. If the result is greater than the epoch size, the node MUST discard the result and use the epoch size instead.
+
+If the `BlockchainParameters.sol` contract errors or returns 0, the node MUST ues the default value `MIN_SAFE_LOOKBACK_WINDOW`.
```
mcortesi, line 66:
```suggestion
If the `BlockchainParameters.sol` contract errors or returns 0, the node MUST use the default value `MIN_SAFE_LOOKBACK_WINDOW`.
```

CIPs/cip-0021.md
```diff
@@ -35,24 +40,71 @@ The core devs have been recieving complaints from the validators saying this win
...
 
-This solution would also imply a contract update.
+```python
+def clamp_lookback_window(window: int) -> int:
+    # clamp to range
+    window = max(window, MIN_SAFE_WINDOW)
+    window = min(window, MAX_SAFE_WINDOW)
```
mcortesi, line 72:
```suggestion
    window = max(window, MIN_SAFE_LOOKBACK__WINDOW)
    window = min(window, MAX_SAFE_LOOKBACK__WINDOW)
```

CIPs/cip-0021.md
```diff
@@ -35,24 +40,71 @@ The core devs have been recieving complaints from the validators saying this win
...
+      window >= MIN_SAFE_LOOKBACK_WINDOW &&window <= MAX_SAFE_LOOKBACK_WINDOW
+    );
+    ...
+  }
 ```
 
-This way, the parameter will be handled by the BlockchainParameters contract and it will be fetched with a static view call from the node (layer 1). This would mean a hard fork, replacing any reference to the parameter with a contract call.
+Starting at `HARDFORK_BLOCK`, block creation and validation MUST read the lookback window from chain state by calling the `BlockchainParameters` contract, rather than reading chain paramaters. Reward calculations are unaffected.
+
+If the return value is outside the safe window, the node MUST clamp the return value to that range such that `MIN_SAFE_LOOKBACK_WINDOW <= window <= MAX_SAFE_LOOKBACK_WINDOW`. If the result is greater than the epoch size, the node MUST discard the result and use the epoch size instead.
```
mcortesi, line 64:
```suggestion
If the return value is outside the safe window, the node MUST clamp the return value to that range such that `MIN_SAFE_LOOKBACK_WINDOW <= window <= MAX_SAFE_LOOKBACK_WINDOW`. If the result is greater than the epoch size, the node MUST discard the result and use `epochSize - 2` instead.
```

mcortesi (approved):
approved minus some typos

CIPs/cip-0021.md
prestwich: Added a short comment about`epochSize` of 3 or less being invalid. If you're good with it, lets get this merged

mcortesi (approved):

CIPs/cip-0021.md
mcortesi: approved!

PR #89, opened 2020-11-19
CIP 28 - Split Etherbase

andres-dg:

PR #115, opened 2020-12-10
Prestwich/cip28 update

prestwich:
This PR rebases #89 onto master, and expands the specification to reflect the current implementation. It is a new PR to avoid merge issues after the rebase.

Please carefully check the specification updates, as I'm not as familiar with the code as you are 👍


mcortesi (approved):
Great!

PR #119, opened 2020-12-11
doc: Move CIP-27 CIPs to Last Call

YazzyYaz:


prestwich (changes requested):
I think we are missing 26 and 28

prestwich (approved):

PR #121, opened 2020-12-15
CIP 32 - Attestation Node Incentives

codyborn:
[Discussion](https://github.com/celo-org/celo-proposals/issues/144)


CIPs/cip-0032.md
```diff
@@ -0,0 +1,117 @@
...
+
+Despite Validator uptime on Celo being phenominal, the maintenance of the AS nodes are [sometimes neglected](https://metabase.celo-networks-dev.org/public/dashboard/b0a27650-1d62-4645-81d7-26ff7546ff0d?date_filter=past2weeks~). Today the implicit motivation to run an AS is to contribute to the health of the network which can help procure more votes from those that are interested in the success of Celo. This proposal attempts to make the incentive more explicit by introducing a slashing penalty in the case of prolonged AS downtime.
+
+In order to slash effectively, we will need an on-chain measure of AS availability. We're researching ways to provide this measurement in a manner that ensures validators are correctly rewarded when their AS is healthy and correctly punished when their AS is down. Until we have a large number of users, it's difficult to isolate a signal we can be confident in as a community. Therefore, we're starting with a conservative approach of measuring a minimum completion threshold over a period of time.
+
+## Specification
+
+- `B` is the number of consecutive blocks in which an elected Asstestation Service is required to complete at least one attestation.
```
aslawson, line 30:
```suggestion
- `B` is the number of consecutive blocks in which an elected Attestation Service is required to complete at least one attestation.
```

CIPs/cip-0032.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Defining parameters
+
+#### AttestationSlasher
+
+- `B` This will be chosen based on the minimal period that it is likely for a validator to have recieved an attestation request. We'll start conservative and look at a period where every validator should have received 10 attestations. In the month of November 2020, there was an average of 94 attestations requested per day (more in December). Therefore, over 11 days, an average of 1034 attesations would be requested, providing an expected value of ~10 attestations for each of the 100 issuers. Since we produce a block every 5 seconds, 11 days is equivalent to 190,080 blocks.
```
aslawson, line 74: Curious, with this parameter at the suggested value, would any validators have been slashable in the past few months?

11 days to be down if very long...but I understand due to low traffic, services get fewer chances to perform. makes sense to start conservative and update when we have more traffic

CIPs/cip-0032.md
```diff
@@ -0,0 +1,117 @@
...
+- `P` Similarly we'll start conservatively with the amount of stake slashed. At the time of writing this, only two active validators have an excess stake of less than 100 CELO (85 CELO each). If we set a penalty of 40 CELO, then these validators can be slashed twice before incurring any loss in epoch rewards.
+- `R` The reward for slashing should cover both the gas costs as well as the infrastructure costs to monitor for offenses.
+
+#### MinStakeSlasher
+
+- `P` The penalty is zero since the validator has already been slashed to get to this point.
+- `R` Similarly, the reward for slashing should cover both the gas costs as well as the infrastructure costs to monitor for offenses.
```
aslawson, line 81: I'm not super familiar with where the funds for slashing rewards come from. Do they get distributed with the general epoch rewards or does a portion of the penalty go to the slashing? If the later, how do we fund the reward if the penalty becomes 0?

CIPs/cip-0032.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Signal online
+
+In the unlikely event that a validator was never selected for an attestation during the period of `B` blocks, it's possible for a malicious actor to request multiple attestations with the goal of randomly getting paired with this validator and intentionally leaving the attestation incomplete. To prevent this, we're building into the AS the ability to signal to the `AttestationSlasher` that the service is still online if a slashing event becomes close by signing and submitting a transaction to the slasher contract. This requires that the AS signer key to be minimally funded to pay the gas fees of the transaction, however this adds more piece of mind for validators to be sure they won't be incorrectly slashed. Additionally, we'll be adding two new Prometheus events to the AS to alert a validator when the slashing period is nearly reached and another event to alert when the AS signer key's balance is low.
```
aslawson, line 58: What would trigger this signal? Does the attestation service pro-actively send this transaction and how is it monitoring its status of being "slashable"?

aslawson: Also how do we make this resistant to fake "signals", a service that just sends a signed `SignalOnline` regularly?

CIPs/cip-0032.md
```diff
@@ -0,0 +1,117 @@
...
+2. Voting frequency is every 3 months, which is not immediate enough to motivate an operator to regularly monitor and maintain their AS availability.
+
+### Governance Slashing
+
+We have a generic slashing contract which allows anyone to slash for any reason via a governance proposal. This process is good for one-off slashing events, but requires a large amount of coordination for each event. We wanted to build something more automatic where the rules are stated explicitly ahead of time.
+
+### Additional Rewards
+
+Rather than introducing a slashing policy which punishes poor performance, we could add additional reward to validators for great AS performance. We haven't ruled this option out and will likely institute it once we have more granular metrics (see [Using completion rate](#using-completion-rate)). Today, many validators put a good amount of effort into maintaining a highly available Attestation Service. Slashing based on poor performance ensures that everyone is putting in this same level of effort for the same compenstation.
```
aslawson, line 117: I think we should highlight this at the top as to why slashing is beneficial. I could see validators initially reading this and being turned off by introducing another way for them to lose rewards with no obvious upside. But the truth is that MOST of our validators do a really good job and put in a lot of effort doing that despite not currently receiving extra compensation. Slashing ensures that all others are doing the same in providing the value that will grow our ecosystem and the network as a whole.

CIPs/cip-0032.md
```diff
@@ -0,0 +1,117 @@
...
+mapping(address => uint256) public lastSlashed;
+```
+
+A slashing event may occur when the slashing initiator calls `slash()` with the slashable address. Before slashing, the slashing contract checks the following:
+
+1. The validator has been elected continuously for the past `B` blocks.
+2. The issuer was selected for at least one attestation in the past `B` blocks.
+3. The validator wasn't slashed by the `AttestationSlasher` within the past `B` blocks.
+4. The address hasn't completed an attestation in the past `B` blocks.
+5. The addres hasn't submitted `SignalOnline` event in the last `S` blocks. (see [below](#signal-online)).
```
aslawson, line 52:
```suggestion
5. The address hasn't submitted `SignalOnline` event in the last `S` blocks. (see [below](#signal-online)).
```

CIPs/cip-0032.md
```diff
@@ -0,0 +1,117 @@
...
...asher` would directly call `LockedGold.slash()`, only slashing stake and not removing the validator. This decreases the potential risk of having multiple validators removed simultaneously due to an overly aggressive `AttestationSlasher` policy. If the validator is slashed it may be that their locked stake drops below the `validatorLockedGoldRequirements` (currently 10K CELO). By dropping below `validatorLockedGoldRequirements` alone, the validator is not removed from the validator set but they will cease to earn epoch rewards. If they continue to get slashed by the `AttestationSlasher`, it may be that they drop to a near-zero stake while still remaining to be a validator. This is a dangerous situation, since this validator is performing an important role for the network but has nothing at stake. To resolve this tension, we'll introduce a new `MinStakeSlasher` which will force a validator to be de-affiliated if their balance drops below the `penalty + reward` threshold of any of the registered slasher contracts. Once a validator has been de-affiliated, they'll need to return their stake amount to the `validatorLockedGoldRequirements` requirement before being able to affiliate again.
```
kevjue, line 68: nit:  I *think* the implication that you are alluding here is that these validators will effectively be removed when going below `validatorLockedGoldRequirements` after the next epoch election.  It would be nice if that's explicitly stated.

CIPs/cip-0032.md
kevjue, line 74: Since a validator must be elected for B contiguous blocks, would having B this large end up making only a fraction of the validators slashable for attestations?

CIPs/cip-0032.md
kevjue, line 58: Seems like one of the main reasons for signalOnline is to handle the case of non completed attestations from a "down" user.  (e.g. the cause of a non completed attestation is the user's fault, as opposed to the AS's).

I am not very familiar with the attestation protocol, so this may not be technically feasible, but perhaps one heuristic to figure out if it's more likely that it's the user's fault is if they didn't complete any other attestations (I believe 3 completed is required by the user?).

Ofc this is just a heuristic (e.g. the user could be malicious and intentionally not complete an attestation for a specific validator), so if it is more likely the user, then perhaps the validator is only fractionally slashed (or something along those lines).

CIPs/cip-0032.md
kevjue: Looks like @aslawson already asked this question :).

CIPs/cip-0032.md
nambrot, line 30: Just a nit but I'm not the biggest fan of super short variable names, especially in specs

CIPs/cip-0032.md
```diff
@@ -0,0 +1,117 @@
...
+
+If we observe that a validator is relying too heavily upon the `SignalOnline` functionality when they're receiving legitimate attestation requests we can add mitigations such as introducing an increasing financial penalty for subsquent signals or by decreasing the time that signal will last for (lowering `S`) per user.
+
+### Requiring Attestation Service
+
+**Note that the slasher expects each validator to be running an Attestation Service.** Unlike today, where it's optional for a Validator to run an attestation node, this proposal would make it mandatory. Having more attestation nodes in the pool improves the security of the phone number verification by making it more difficult to game the random selection of a set of colluding issuers. This proposal recommends providing at least three weeks from time of merge of this CIP to the time of enacting the slashing mechanism to give everyone an opportunity to set up their Attestation Service.
```
nambrot, line 64: I like the spirit of diversifying the attestors, but also requiring reluctant validators to run AS might decrease the overall reliability of the attestation protocol

CIPs/cip-0032.md
carterqw2, line 58:
```suggestion
In the unlikely event that a validator was never selected for an attestation during the period of `B` blocks, it's possible for a malicious actor to request multiple attestations with the goal of randomly getting paired with this validator and intentionally leaving the attestation incomplete. To prevent this, we're building into the AS the ability to signal to the `AttestationSlasher` that the service is still online if a slashing event becomes close by signing and submitting a transaction to the slasher contract. This requires that the AS signer key to be minimally funded to pay the gas fees of the transaction, however this adds more peace of mind for validators to be sure they won't be incorrectly slashed. Additionally, we'll be adding two new Prometheus events to the AS to alert a validator when the slashing period is nearly reached and another event to alert when the AS signer key's balance is low.
```

carterqw2: In practice, it sounds more like "if the AS appears to be available it doesn't get penalized". In this case maybe we could track the ratio of the AS being selected to the number of "initialized attestations" instead?

CIPs/cip-0032.md
carterqw2: Could this value be dynamic and provided by an oracle based on the situation in the current period?

CIPs/cip-0032.md
codyborn: The rewards come from the staked funds that are being slashed. The reward is first given, then the remaining amount is slashed. You're correct that the penalty should actually be equal to the reward in this case. I was originally thinking about the penalty being the addition to the reward but it's clearer if we assume it to be the total amount.

CIPs/cip-0032.md
codyborn: I hear the concern. Currently, 96/100 elected validators have registered Attestation Service claims so I don't think it will be too big of a lift. The alternative is that we allow validators to opt-out by deregistering an AS, but this doesn't feel fair to those that opt-in to support the network.

CIPs/cip-0032.md
```diff
@@ -0,0 +1,117 @@
...
+
+Encourage liveness and global coverage of the Attestation Service (AS) through the incentivization of successful customer onboarding. This CIP will introduce slashing opportunities in the case of long term outages of the AS.
+
+## Abstract
+
+The Attestation Service performs a critical role in user onboarding when linking a phone number to a Celo address. When AS nodes are down or have poor coverage they lead to a degraded user experience. This proposal focuses on explicit incentives to run a highly available AS.
+
+## Motivation
+
+Despite Validator uptime on Celo being phenominal, the maintenance of the AS nodes are [sometimes neglected](https://metabase.celo-networks-dev.org/public/dashboard/b0a27650-1d62-4645-81d7-26ff7546ff0d?date_filter=past2weeks~). Today the implicit motivation to run an AS is to contribute to the health of the network which can help procure more votes from those that are interested in the success of Celo. This proposal attempts to make the incentive more explicit by introducing a slashing penalty in the case of prolonged AS downtime.
```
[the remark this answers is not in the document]
codyborn: Great questions. 

Regarding the positive incentive (additional reward): With this proposal we have a min-bar measurement on-chain that shows the validator is running an attestation service, but it doesn't indicate how well they're doing it. They could be failing 90% of the time and still not get slashed. I wanted to save the positive reward once we have confidence in a metric beyond this min-bar measure.

Regarding the requirement that all validators run the attestation service: Given that 96/100 already are running it, I think it's fair to make it a requirement that everyone runs it and shares the responsibility.

CIPs/cip-0032.md
codyborn: Looking at the 11 days from 1/1 to 1/11 inclusive, there are two such validators that are registered that have 100% failure rates (0xeD32008dB382d078bABB0eB6F91Df0207d9098990, x3BEbD7Ed4cA0ce7BFd69fC6aeF31dDD21C460158)
Here's the query for future:
```
SELECT DISTINCT issuer
FROM `celo-testnet-production.analytics_eksportisto.attestation_state_with_user_data`
WHERE timestamp >= "2021-01-01"
  AND timestamp <= "2021-01-11"
AND issuer NOT IN
(SELECT DISTINCT issuer FROM `celo-testnet-production.analytics_eksportisto.attestation_state_with_user_data`
  WHERE timestamp >= "2021-01-01"
  AND timestamp <= "2021-01-11"
  AND isCompleted)
  ```
  This doesn't include the four validators which don't have attestation services registered.

Comment (codyborn) [reply CIPs/cip-0032.md:74]: > Could this value be dynamic and provided by an oracle based on the situation in the current period?
I think to start with, it'd be safest to use governance proposals to adjust the parameters, but I can totally imagine this threshold being a function of the total number of attestations over the past week/month.

File: CIPs/cip-0032.md (hunk 4 above)
Comment (codyborn) [reply CIPs/cip-0032.md]: The service would query the new slasher contract which tracks the last successful attestation. If the time period becomes too close to the slashable event, then the attestation service will automatically submit a `SignalOnline` transaction if the account is funded. 

We will need to create a dashboard to monitor the use of this signal. If it looks like a validator is using it as a way to bypass running a healthy service, then we'll need to either:
1. Remove the feature. This feature is a temporary safety mechanism in case we have low onboarding rates. If we have collective confidence in the min-bar, then we can remove this safety feature.
2. Add mitigations such as introducing an increasing financial penalty for subsequent signals or by decreasing the time that signal will last for (lowering `S`) per user.

Comment (codyborn) [reply CIPs/cip-0032.md]: @kevjue I was also thinking along the same lines and already added it [our dashboard](https://metabase.celo-networks-dev.org/public/dashboard/b0a27650-1d62-4645-81d7-26ff7546ff0d) (see `OnlyMissingIssuer` column in the `Missed Attestations` table). `OnlyMissingIssuer` implies that this validator was the only one that the user missed when completing attestations, indicating it's likely the fault of the validator and not the user. This is metric is nice because it is derived purely from on-chain data. Once we're more confident in the metric we can use this to reward validators who run with exceptional services.

The SignalOnline feature does help cover this gap, but it also can cover cases of an outage caused by cLabs (ex. Valora or Komenci are having issues), or simply unexpected lulls in user onboarding. Once we build more confidence in this min-bar metric, we can remove the SignalOnline feature.

Comment (codyborn) [reply CIPs/cip-0032.md]: @carterqw2 by "initialized attestations" do you mean the number of attestations that were completed for that issuer? The concern with ratios for now is that they're gamable by malicious actors (ex. I can make other validators look bad or make my own look good). In addition to the metric that Kevin called out, I'd like to also explore adding an on-chain attestation from Komenci indicating that the user solved a reCaptcha and device checks. This attestation can be used to build confidence in the on-chain metric and make it more difficult to game.

File: CIPs/cip-0032.md (hunk 9 above)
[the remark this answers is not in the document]
Comment (aslawson) [reply CIPs/cip-0032.md]: Gotcha -- I see in the intro, you now mentioned it as a future possibility and included the benefit to having all validators put in the same work for the same compensation. Looks good 👍

File: CIPs/cip-0032.md (hunk 2 above)
Comment (aslawson) [reply CIPs/cip-0032.md]: Ya, makes sense to start simple and conservative. Can always tweek the static value or make it dynamic later.

File: CIPs/cip-0032.md (hunk 4 above)
Comment (aslawson) [reply CIPs/cip-0032.md]: Under what conditions would the attestation service send and non-send this signal? Does attestation service itself maintain a timer to know when its close to a slashable event and check if it's had a completion in that time period?

I guess my hesitancy is if this would negate most slashable events (eg. the two you found in your query above appear to be running attestation services, so they would probably send a signal...)

I think we have a conservative min-bar suggested and we can decrease as we feel confident, so not sure if this is needed. Plus, it adds more complexity. WDYT?

Comment (aslawson) [reply CIPs/cip-0032.md]: Or can we just make sure we are running the attestation bot in mainnet at a rate similar to the slash event rate to ensure that each validator at least gets a chance for 1 non-malicious request

Comment (codyborn) [reply CIPs/cip-0032.md]: Good points Amy. I agree that this feature does add complexity and the benefits could be realized using other mechanisms. Since its purpose is to mitigate the risks until we build up more confidence in the slashing rules, we can get the same benefits by simply lowering the slashable penalty to a trivial amount to begin with. Once everyone is confident we can ramp it up. WDYT?

Comment (aslawson) [reply CIPs/cip-0032.md]: Ya, I like that approach 👍

File: CIPs/cip-0032.md [hunk 10]
```diff
@@ -0,0 +1,120 @@
...
+
+In order to slash effectively, we will need an on-chain measure of AS availability. We're researching ways to provide this measurement in a manner that ensures validators are correctly rewarded when their AS is healthy and correctly punished when their AS is down. Until we have a large number of users, it's difficult to isolate a signal we can be confident in as a community. Therefore, we're starting with a conservative approach of measuring a minimum completion threshold over a period of time.
+
+## Specification
+
+- `MinAttestationSlashablePeriod` is the number of consecutive blocks in which an elected Attestation Service is required to complete at least one attestation.
```
asaj, line 32: What initial values are you proposing for these variables, and how did you come to those values?

CIPs/cip-0032.md
```diff
@@ -0,0 +1,120 @@
...
+
+We have a generic slashing contract which allows anyone to slash for any reason via a governance proposal. This process is good for one-off slashing events, but requires a large amount of coordination for each event. We wanted to build something more automatic where the rules are stated explicitly ahead of time.
+
+### Signal online
+
+We considered introducing a mechanism for the AS to signal that it was still online in the case that it never had the opportunity to be selected for an attestation. The goal of this mechanism was to add more assurance that a validator wouldn't be incorrectly slashed if they didn't have a chance to naturally prove that they were online. Instead we're mitigating this risk in the following ways:
+
+1. Starting with a conservative slashing penalty.
+2. Requiring a single attestation in a time period where we'd naturally expect 10 to have occurred.
+3. Slashing can only occur if the AS was selected for at least one attestation in the time period.
```
asaj, line 118: This feels like it contradicts (2), why "one" instead of "ten"?

CIPs/cip-0032.md
```diff
@@ -0,0 +1,120 @@
+---
+cip: 32
+title: Attestation Incentives
+author: Cody Born <cody@clabs.co>
+discussions-to: https://github.com/celo-org/celo-proposals/issues/144
+status: Draft
+type: Standards
+category: Ring 1
```
YazzyYaz, line 8: @prestwich I believe this is accurate, but wanted to double check. It adds slashing via a core contract if a validator's attestation service isn't up, but it doesn't necessarily impact Celo blockchain's functionality (only impacts Valora's given its reliance on the attestation service).

prestwich: 👍

CIPs/cip-0032.md
codyborn: Please see the `Defining parameters` section below.

CIPs/cip-0032.md
```diff
@@ -0,0 +1,120 @@
...
+
+In order to slash effectively, we will need an on-chain measure of AS availability. We're researching ways to provide this measurement in a manner that ensures validators are correctly rewarded when their AS is healthy and correctly punished when their AS is down. Until we have a large number of users, it's difficult to isolate a signal we can be confident in as a community. Therefore, we're starting with a conservative approach of measuring a minimum completion threshold over a period of time.
+
+## Specification
+
+- `MinAttestationSlashablePeriod` is the number of consecutive blocks in which an elected Attestation Service is required to complete at least one attestation.
+- `AttestationSlasherPenalty` is the penalty amount that is slashed in a slashing event.
+- `AttestationSlasherReward` is the reward amount that a slasher is granted.
+- `MinAttestationSlashablePeriod`, `AttestationSlasherPenalty`, and `AttestationSlasherReward` are easily configurable via governance proposals.
```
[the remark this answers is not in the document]
codyborn: Please see the 'Defining parameters' section below for the proposed initial values.

CIPs/cip-0032.md
```diff
@@ -0,0 +1,120 @@
...
+// Tracks the last time an attestation was completed by an issuer
+mapping(address => uint256) public lastAttestationCompleted;
+// Tracks the last time an issuer was last slashed
+mapping(address => uint256) public lastSlashed;
+```
+
+A slashing event may occur when the slashing initiator calls `slash()` with the slashable address. Before slashing, the slashing contract checks the following:
+
+1. The validator has been elected continuously for the past `MinAttestationSlashablePeriod` blocks.
+2. The issuer was selected for at least one attestation in the past `MinAttestationSlashablePeriod` blocks.
```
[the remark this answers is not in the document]
codyborn: The `MinAttestationSlashablePeriod` is the variable we would use to change the "difficulty" of the min-bar. The number of required attestations is the numerator and the `MinAttestationSlashablePeriod` is the denominator. We can adjust either to change the difficulty. I'm proposing holding the numerator constant since it also simplifies the state management in the contract.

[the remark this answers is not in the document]
codyborn: @YazzyYaz the initial values are proposed below. Are you looking for something like this?
> 1. The validator has been elected continuously for the past `MinAttestationSlashablePeriod` (190,080) blocks.

CIPs/cip-0032.md
codyborn: 2 is saying that if that the contract's min-bar is very low. Based on a snapshot of traffic in November, there should be 10 attestations randomly assigned to an AS in the time-frame that the slasher only requires one.

3 is saying that in the case that there happens to be zero attestations requested from an AS in the time-frame, then there's a safety mechanism that will prevent the AS from being slashed. In other words, the contract won't incorrectly slash an AS just because there wasn't any traffic.

CIPs/cip-0032.md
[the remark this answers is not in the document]
YazzyYaz: Sorry, I just meant going from `one` to `1` for any values assigned to variables in a spec.

CIPs/cip-0032.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Nothing at stake problem
+
+Unlike other slashing contracts which use `SlasherUtil.performSlashing()` to both slash _and_ remove the validator, this `AttestationSlasher` would directly call `LockedGold.slash()`, only slashing stake and not removing the validator. This decreases the potential risk of having multiple validators removed simultaneously due to an overly aggressive `AttestationSlasher` policy. If the validator is slashed it may be that their locked stake drops below the `validatorLockedGoldRequirements` (currently 10K CELO). By dropping below `validatorLockedGoldRequirements` alone, the validator is not removed from the validator set but they will cease to earn epoch rewards. If they continue to get slashed by the `AttestationSlasher`, it may be that they drop to a near-zero stake while still remaining to be a validator. This is a dangerous situation, since this validator is performing an important role for the network but has nothing at stake.
```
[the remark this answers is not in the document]
codyborn: Added the following line to clarify:
> Additionally, by not using `SlasherUtil.performSlashing()`, the _slashing multiplier_ will not be impacted for the slashed validator (see `halveSlashingMultiplier()` for more details).

YazzyYaz (approved):

PR #142, opened 2021-01-07
Adds explanation of baselineQuorumFactor mechanics to cgp16

alecps:


CGPs/0016.md
```diff
@@ -11,6 +11,8 @@
 
 This CGP is an effort to lower the barriers to passing governance proposals given current levels of voter engagement. We propose a decrease in the `baselineQuorumFactor`, the proportion of the historical average voter participation level that constitutes quorum for governance proposals, from `1` to `0.5`.
 
+The mechanics of the `baselineQuorumFactor` are as follows. The `baseline` is an exponential moving average of final participation in past governance proposals, and the `baselineQuorumFactor` is simply the proportion of that average participation level that is required for future proposals to have a quorum of votes. For example, if 1M CELO is the `baseline` then a `baselineQuorumFactor` of `0.5` means that at least 0.5M CELO must participate in voting for the next governance proposal for it to pass, whereas a `baselineQuorumFactor` of `1` would mean that at least 1M CELO must participate for the proposal to pass.
```
asaj, line 14: This is close to correct but not exactly how it operates. If participation is less than `baseline * baselineQuorumFactor`, that difference is treated as "no" votes. The proposal then passes if the Yes/No ratio is higher than the threshold specified by the governance constitution given the contents of the proposal.

alecps: Thanks!

YazzyYaz (approved):

PR #148, opened 2021-01-19
Complete specification for  cip-21

mcortesi:
Add new scenario considerations and changes to cip-21

### DIFF
--- a/CIPs/cip-0021.md
+++ b/CIPs/cip-0021.md
@@ -42,17 +42,17 @@ In practice, the lookback window has penalized validators for performing routine
 
 The proposed solution is to migrate the parameter to the EVM state (layer 2).
 
-Starting at the first epoch after `HARDFORK_BLOCK`, block creation and validation MUST read the lookback window from chain state by calling the `BlockchainParameters` contract, rather than reading chain paramaters. Reward calculations are unaffected.
+Starting at the first full post-fork epoch (i.e. the same epoch as `HARDFORK_BLOCK` if it's the first block of its epoch, or the next epoch if it isn't), block creation and validation MUST read the lookback window from chain state by calling the `BlockchainParameters` contract, rather than using the value from the genesis config. Reward calculations are unaffected.
 
 In order for `uptimeLookbackWindow` to remain constant during an epoch, the following rules are applied:
 
  1. The `setter` on `BlockchainParameters.sol` for `uptimeLookbackWindow` MUST apply the new value to the next epoch (not the current one)
- 2. The `getter` on `BlockchainParameters.sol` for `uptimeLookbackWindow` MUST revert if there's no value for current epoch. This error signals the blockchain client to use the `defaultValue` (pre Hardfork value).
+ 2. The `getter` on `BlockchainParameters.sol` for `uptimeLookbackWindow` MUST revert if there's no value for the current epoch. This error signals the blockchain client to use the `defaultValue` (pre Hardfork value).
  
 As the `uptimeLookbackWindow` is now governable. Work must be done to verify the value is within safe boundaries; which are defined by the following rules:
 
  1. `uptimeLookbackWindow` MUST BE within the range `[MIN_SAFE_LOOKBACK_WINDOW, MAX_SAFE_LOOKBACK_WINDOW]`
- 2. `uptimeLookbackWindow` MUST BE less or equal to `epoch size - 2`
+ 2. `uptimeLookbackWindow` MUST BE less than or equal to `epoch size - 2`
 
 where:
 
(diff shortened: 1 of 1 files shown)


CIPs/cip-0021.md
```diff
@@ -35,36 +35,37 @@ In practice, the lookback window has penalized validators for performing routine
...
 - The lookback window parameter must be adjustable.
+- The lookback window parameter must be constant during an epoch
 - The community must decide on the value.
 
 ## Specification
 
 The proposed solution is to migrate the parameter to the EVM state (layer 2).
 
-We define the following constants:
+Starting at the first epoch after `HARDFORK_BLOCK`, block creation and validation MUST read the lookback window from chain state by calling the `BlockchainParameters` contract, rather than reading chain paramaters. Reward calculations are unaffected.
```
oneeman, line 45:
```suggestion
Starting at the first full post-fork epoch (i.e. the same epoch as `HARDFORK_BLOCK` if it's the first block of its epoch, or the next epoch if it isn't), block creation and validation MUST read the lookback window from chain state by calling the `BlockchainParameters` contract, rather than using the value from the genesis config. Reward calculations are unaffected.
```

This change makes it correct also for the edge case where `HARDFORK_BLOCK` happens to be the first block on an epoch.

CIPs/cip-0021.md
```diff
@@ -35,36 +35,37 @@ In practice, the lookback window has penalized validators for performing routine
...
 
-- `MIN_SAFE_LOOKBACK_WINDOW = 3`
-- `MAX_SAFE_LOOKBACK_WINDOW = 720`
+In order for `uptimeLookbackWindow` to remain constant during an epoch, the following rules are applied:
 
-Before `HARDFORK_BLOCK` The `BlockchainParameters.sol` contract is updated to add a new public state variable `uptimeLookbackWindow` with a governable setter.
+ 1. The `setter` on `BlockchainParameters.sol` for `uptimeLookbackWindow` MUST apply the new value to the next epoch (not the current one)
+ 2. The `getter` on `BlockchainParameters.sol` for `uptimeLookbackWindow` MUST revert if there's no value for current epoch. This error signals the blockchain client to use the `defaultValue` (pre Hardfork value).
```
oneeman, line 50:
```suggestion
 2. The `getter` on `BlockchainParameters.sol` for `uptimeLookbackWindow` MUST revert if there's no value for the current epoch. This error signals the blockchain client to use the `defaultValue` (pre Hardfork value).
```

CIPs/cip-0021.md
```diff
@@ -35,36 +35,37 @@ In practice, the lookback window has penalized validators for performing routine
...
+ 
+As the `uptimeLookbackWindow` is now governable. Work must be done to verify the value is within safe boundaries; which are defined by the following rules:
 
-```solidity
-  uint256 public uptimeLookbackWindow;
-  event UptimeLookbackWindowSet(uint256 window);
+ 1. `uptimeLookbackWindow` MUST BE within the range `[MIN_SAFE_LOOKBACK_WINDOW, MAX_SAFE_LOOKBACK_WINDOW]`
+ 2. `uptimeLookbackWindow` MUST BE less or equal to `epoch size - 2`
```
oneeman, line 55:
```suggestion
 2. `uptimeLookbackWindow` MUST BE less than or equal to `epoch size - 2`
```

CIPs/cip-0021.md
```diff
@@ -82,7 +83,7 @@ def get_lookback_window(chain: ChainState) -> int:
         window = chain.evm.get_blockchain_param("lookbackWindow")
```
oneeman, line 83: github won't let me comment on the right line.  But the comment above: `# default to 12 in case of error` needs to be changed to reflect the changes.  Also this function is missing the `epochSize - 2` part

CIPs/cip-0021.md
```diff
@@ -96,15 +97,80 @@ def get_lookback_window(chain: ChainState) -> int:
...
+  }
+```
+
+### Scenario Considerations
+
+In the current specification, the following migration scenarios are considered:
+
+ 1. **Hardfork is activated before Smart Contract is upgraded:** On this scenario, on the first epoch after hardfork the blockchain client will fail to obtain `uptimeLookbackWindow` from the contract; and will use `DEFAULT_VALUE`. Then, when smart contract is deployed (mid-epoch), it will continue to fail until next epoch where initialized value will be used. Thus, the value remains constant during the epoch.
+ 2. **Smart Contract is upgraded before Hardfork is activated:** On this scenario, on the first epoch after hardfork the blockchain client will obtain `uptimeLookbackWindow` successfully from the smart contract, and use the same value during the epoch.
+ 3. **setUptimeLookbackWindow() is called mid-epoch:** On this scenario, the new value will only be used after epoch change, thus value will remain constant during the epoch.
```
oneeman, line 163:
```suggestion
 1. **The first post-fork epoch starts before a lookback window from the smart contract is available:** This scenario includes several cases: the smart contract not having been upgraded yet, the smart contract having been upgraded but `setUptimeLookbackWindow()` not having been called, or `setUptimeLookbackWindow()` having been called but its value not having come into effect yet.  In all these cases, in the first epoch after the hard fork, the blockchain client attempt to obtain `uptimeLookbackWindow` from the contract, but will get an error, and will then use `DEFAULT_VALUE`. This will continue until the first block of the first epoch after the contract has been upgraded and `setUptimeLookbackWindow()` has been called successfully, from which point onwards the smart contract call will succeed and return the lookback window which the client will then use. Thus, the value does not change within an epoch.
 2. **The smart Contract is upgraded and initialized before, or in the same epoch as, the hard fork activation:** In this scenario, in the first post-fork epoch, the blockchain client will obtain `uptimeLookbackWindow` successfully from the smart contract, and use that value during the epoch.
 3. **setUptimeLookbackWindow() is called mid-epoch:** On this scenario, the new value come into effect at the beginning of the next epoch, thus the value returned by the contract will remain constant during the epoch.
```

Making this more complicated, but more correct.  There be better ways to phrase things, though.

oneeman (commented):
Made some suggestions to cover the cases more fully, and a comment about the pseudo-code.

prestwich (commented):
Agree with Or's suggestions. Will approve once they are included

prestwich (approved):

YazzyYaz (approved):

Issue #144, opened 2021-01-12
CIP 32 - Attestation Node Incentives - Discussions


codyborn:
CIP 32 https://github.com/celo-org/celo-proposals/pull/121 proposes introducing Attestation Service operation incentives via slashing. This is the first step toward a more discerning incentive system. As such, it's focused on a min uptime bar (one attestation over 11 days) and will not increase validator rewards. Future proposals will likely support more granular uptime measures and the potential for operators to earn additional reward.

aaronmboyd:
A little bit devil's advocate here, but I'm not sure I agree with implementing slashing for downtime on a service ultimately unrelated to block proposals and consensus of the protocol. This is not an incentive, it's a disincentive.

I definitely prefer a reward rather than punishment model. Once there is a positive economic incentive wont validators have a real desire to keep their service online? (Especially as new users are increasing).

codyborn:
Thanks for starting the discussion @aaronmboyd. I think this is a matter of perspective. One could view the existing validator reward to cover all services that a validator provides, including running an attestation service. In this case, by default the protocol assumes the operator is running the attestation service properly, and will only withhold some of the reward (via slashing) if it finds out later that this is not the case. To make it more explicit, we could cap the slashable amount to some limit per epoch OR instead of slashing from the LockedGold, we could reduce the block rewards by some capped amount. WDYT?

I went with this perspective since a majority of validators are running Attestation Services (96/100) and a majority are doing the job well. Making this the default case made sense to me, but I'm definitely open to other thoughts. I also believe that we should increase the block reward to those that run the Attestation Service exceptionally well. Since measuring this on-chain introduces more complexity, it was reserved for a future proposal.

nategraf:
Just read through the proposal and have some comments

## Attack on proposed slashing method

A validator that has a downed attestation service, or simply does not want to run one, could do the following in every slashing period. 
1. Request 100 attestation requests (more generally, the number of validators with registered attestation keys), which guarantees exactly one request will request will be assigned to themselves when selecting issuers.
2. Fulfill the request by generating the required signature and posting it to the chain.

Note that the validator does not need to have possession of the phone number associated with the identifier they request against, nor does the identifier need to correspond to a known phone number and pepper.

This attack can prevent slashing for ~$5 spent every slashing period, about $15/month under the proposal.

## Alternative to MinStakeSlasher

[...]

PR #161, opened 2021-02-10
Attestation Service Incentives v2

codyborn:
Update to Attestation Service incentives design taking into account [community feedback](https://github.com/celo-org/celo-proposals/issues/144).


CIPs/cip-0032.md
```diff
@@ -13,81 +13,105 @@ license: Apache 2.0
...
 
-## Specification
+The Attestation Reward will be affected directly by the completion rate of the Attestation Service during each epoch. Since a 100% completion rate is unrealistic, a governable target completion rate will be used to determine the rewards. To ensure that an Attestation Service isn't incorrectly penalized by a bug outside of their responsibilities (ex. Valora), a minimum global attestation completion rate for this epoch will be first considered. If the minimum global attestation completion rate is not met, all elected Attestation Services will be rewarded with 75% the maximum amount, regardless of their score. A registered Attestation Service can be forcibly deregistered if they don't meet the minimum number of attestations over a governable sliding window. The Attestation Reward will be forfeited if the validator does not have a registered Attestation Service.
```
nategraf, line 30: >  a minimum global attestation completion rate for this epoch will be first considered

I'd love for us to consider automatic mechanisms for setting this rate. I am worried that any static rate would run the risk of unfairly penalizing all validators in case of coordinated failure, such as a natural disaster (e.g. hurricane, military coup, Twilio outage, etc...) or bug in an attestation client (e.g. Valora) that causes attestations to fail.

One example would be to take the mean completion rate and define the threshold as some fraction of it. (e.g. set the fraction to `1/3`) This would prevent an attacker or coordinated failure from bring down the rewards of all validators. This fraction would, by [Markov's inequality](https://en.wikipedia.org/wiki/Markov%27s_inequality), define an upper bound on the number of validator's that can fall below it at any given time. (It would not however, prevent an attacker motivated to _prevent_ a particular subset of validators from losing rewards)

CIPs/cip-0032.md
```diff
@@ -13,81 +13,105 @@ license: Apache 2.0
...
+// These values are reset per issuer from Validators.distributeEpochPaymentsFromSigner()
+mapping(address => uint256) public attestationRequestCount;
+mapping(address => uint256) public attestationCompletionCount;
+
+// Track global completion rate
+// currEpochIndex = getEpochNumberOfBlock(block.number) % 2
+// These values get reset when epoch changes (during the first attestation request)
+uint256 public epochRequestCount;
+uint256 public epochIndexToCompletionCount;
+uint8 private currEpochIndex;
```
nategraf, line 54: Because it is not grouped with other state variables, this variable will still take up a full 256-bit word. So unless there is a natural grouping, we might as well make this `uint256`. (We could certainly turn `epochRequestCount` and `epochIndexToCompletionCount` into `uint64` though.)

CIPs/cip-0032.md
```diff
@@ -13,81 +13,105 @@ license: Apache 2.0
...
 ```js
-// Tracks the last time an attestation was requested from an issuer (AS node)
-mapping(address => uint256) public lastAttestationRequested;
+// Tracks per issuer completion rate
+// These values are reset per issuer from Validators.distributeEpochPaymentsFromSigner()
+mapping(address => uint256) public attestationRequestCount;
+mapping(address => uint256) public attestationCompletionCount;
+
+// Track global completion rate
+// currEpochIndex = getEpochNumberOfBlock(block.number) % 2
```
nategraf, line 50: Why use `% 2` when a full 8 bits are available? We could change this to `% 256`. It's probably the same either way, because the only case where this would matter is if not requests are seen for a full epoch, which is unlikely, but it's also not impossible. (Also see comment below as there may actually be no advantage to using `uint8` here anyway.)

nategraf (commented):
Added a comment about using an automatically adjusting threshold. I am personally in favor of doing so. Also some comments on the code snippets included. I don't want to get caught on that though, as implementation and optimizations can be discussed later.

CIPs/cip-0032.md
```diff
@@ -13,81 +13,105 @@ license: Apache 2.0
...
+- `MinGlobalAttestationCompletionRate` - The global completion rate rarely dips below 70%. We can set this safety net at `.5` and adjust it lower if it was ever necessary.
+- `GlobalAttestationFailureFallbackReward` - The starting point will be `.75`. This is low enough to encourage Attestation Services to come back online in the case of a total outage, but high enough to not significantly impact rewards.
 
-- `MinStakeSlasherPenalty` The penalty is only the slasher reward (no additional penalty) since the validator has already been slashed to get to this point.
-- `MinStakeSlasherReward` Similarly, the reward for slashing should cover both the gas costs as well as the infrastructure costs to monitor for offenses.
+- `AttestationMinPeriod` - An Attestation Service should complete at least 1 attestation over 2 days. This number can be more aggressive since there's no penalty and the validator can always re-register to get back in.
+- `AttestationMinRequested` - 8.
+- `MinNumberOfAttestationIssuers` - 40. Any lower than this and the random selection of issuers is too easy to game.
```
nategraf, line 92: Do we have anywhere the attack model for this? It may be worth adding that here is nowhere else. I'd be happy to write up such an analysis.

CIPs/cip-0032.md
codyborn: Love the idea @nategraf! In this case, we could set the threshold as 1/3 of the moving average over the last n epochs? I  like how it would auto increase/decrease the threshold as things outside of the AS node's control changed. In the case of an abrupt, sustained outage (over n epochs), the threshold will gradually drop leading to reduced payouts. I suppose if this behavior were not desirable, we could always lower the target completion rate via governance; however the time-delayed penalty seems like a benefit as it allows everyone enough time to recover before penalizing. WDYT?

CIPs/cip-0032.md
codyborn: I was thinking the same.. I left it as `uint8` in case we decide to add more state that is sub-256. Will leave it up to the implementer to decide.

codyborn: Good point on the other vars. I'll update them.

CIPs/cip-0032.md
codyborn: Good point!

CIPs/cip-0032.md
codyborn: I haven't seen any model for this, but it's a good idea to call it out explicitly. If you're interested in adding it, it'd be much appreciated 🙏 It'd be interesting to see the expected cost to "hijack" a phone number with different numbers of colluders.

CIPs/cip-0032.md
codyborn: @nategraf updated the docs to reflect this. I chose a 30 day sliding window to measure the baseline to ensure that a multi-day outage doesn't pull the baseline down too much.

CIPs/cip-0032.md
```diff
@@ -1,7 +1,7 @@
 ---
 cip: 32
 title: Attestation Incentives
-author: Cody Born <cody@clabs.co>, Zviad Metreveli
+author: Amy Slawson <amy@clabs.co>, Cody Born <cody@clabs.co>, Victor Graf <victor@celo.org>, Zviad Metreveli
```
nategraf, line 4:
```suggestion
author: Amy Slawson <amy@clabs.co>, Cody Born <cody@clabs.co>, Victor Graf <victor@clabs.co>, Zviad Metreveli
```

CIPs/cip-0032.md
```diff
@@ -43,23 +44,22 @@ The `Attestations` contract will be upgraded to register the request and complet
 ```js
 // Tracks per issuer completion rate
 // These values are reset per issuer from Validators.distributeEpochPaymentsFromSigner()
-mapping(address => uint256) public attestationRequestCount;
-mapping(address => uint256) public attestationCompletionCount;
+mapping(address => uint64) public attestationRequestCount;
+mapping(address => uint64) public attestationCompletionCount;
```
nategraf, line 48: Because [mappings store their values at sparse memory locations](https://docs.soliditylang.org/en/v0.4.24/miscellaneous.html), the value storage cannot benefit from packing. As a result, using `uint64` as the mapping value will use the same about of storage and slightly more computation to access it.

CIPs/cip-0032.md
```diff
@@ -27,11 +27,12 @@ Despite Validator uptime on Celo being phenomenal, the maintenance of the AS nod
...
 
 - `TargetCompletionRate` - The target completion rate for Attestation Services, taking into account standard user drop-off and other factors outside of the AS control.
 - `AttestationServiceRewardPercentage` - The percentage of max rewards that are allocated to running an Attestation Service. Initially `.20`.
```
nategraf, line 33: Just a callout that based on feedback we have gotten on contract release 3, we should aim to have the smart contract changes deployed with parameters that do not change the system and then pass another governance proposal to set the new parameters. In this case, we would want to set this to 0, and make a proposal that lands after the release to set it to `0.20`

CIPs/cip-0032.md
nategraf: On the topic of hijacking a phone number, I include some analysis in https://github.com/celo-org/celo-monorepo/issues/7057#issuecomment-779639780

nategraf:
I have a gut feeling that there are too many parameters in the proposed design. It's just an intuition, but I figured I'd propose an alternative that might be a bit of a simpler direction.

With the target completion rate, I am concerned the the steady-state completion rate may be either lower or higher for well-performing nodes. In the lower case, which would be true if we see an uptick in spurious requests or requests from a country we can't actually reliably deliver SMS, the result will be that validators will see their pay reduced without recourse. In the higher case, the system loses any pressure to continue raising completion rates until a governance proposal is passed, which may take time.

As for the global circuit-breaker, we lose the incentive for nodes to be more resilient than average such as by taking action to setup failover SMS providers. It's kind of the opposite of how Eth2 slashing works [where the penalties are designed to disincentivize coordinated failure](https://blog.ethereum.org/2020/01/13/validated-staking-on-eth2-1-incentives/#anti-correlation-and-decentralisation). As I was discussing earlier, we do need to have the system not overly punish validators for events outside their control, such as phone service interruption in an autocratic country, but we should create an incentive for validators to overcome those difficulties where possible. (In the extreme case, a validator could decide to shut down it's attestations API whenever the circuit breaker is on, as it will no longer have an ill effect on their rewards)

My alternative proposal is to calculate the completion rate on each epoch, then make the threshold for full rewards be a fixed difference from average. For example, if the completion rate for an epoch is `70/100` and the threshold difference is  `0.1`, any validators with more than a 60% completion rate will get full rewards. Below that I would suggest the rewards be a linear function of the completion rate in relation to the threshold. In this example, a validator with 30% completion for the epoch will receive 50% of the max rewards. When the threshold difference is greater than `0`, it is possible for all validators to receive full rewards, but there is still pressure to push their individual completion rate up to ensure it is safely above the threshold in case it rises. It would also be possible to set the difference below `0` to increase the pressure to raise completion rates, but then it would be impossible for all validators to receive full rewards. In the scenario of an adverse external event, overall completion rates may drop and the rewards metric will have no delay in adjusting its target. Imagining that an outage or bug causes an overall 30% completion rate, the threshold would be 20% allowing validators to receive full rewards and providing an incentive to be as resilient as possible. It also adjusts well in cases that the overall completion rate goes up as an overall completion rate of 90% would set an 80% threshold.

Many of these points come more from a place of intuition than any drawn-out analysis, so I'd love to have some discussion of the pros and cons.

CIPs/cip-0032.md
codyborn: Good to know. Do you have any pointer to the discussion? I'm curious why that is.

CIPs/cip-0032.md
codyborn: Oh thanks, good to know. The quirks of Solidity...

CIPs/cip-0032.md
nategraf: https://forum.celo.org/t/core-contracts-release-3/755/2?u=victor

codyborn: Got it, this is more about governance and community acceptance than any technical limitation. I totally agree that we should break apart CGPs wherever possible

CIPs/cip-0032.md
```diff
@@ -1,7 +1,7 @@
 ---
 cip: 32
 title: Attestation Incentives
-author: Cody Born <cody@clabs.co>
+author: Amy Slawson <amy@clabs.co>, Cody Born <cody@clabs.co>, Victor Graf <victor@clabs.co>, Zviad Metreveli
```
YazzyYaz, line 4: Zviad needs an email

YazzyYaz (approved):
Will merge it as soon as Zviad's email is added per the comment

PR #162, opened 2021-02-17
[cip] flash swap support

zviadm:


prestwich:
Generally speaking, I'm in favor of this. I'd like to see more specification (should we follow the uniswap interface precisely? can we achieve the same guarantees uniswap does?)

YazzyYaz (approved):

PR #163, opened 2021-02-17
[CIP 35] Draft: supporting ethereum-compatible transactions

oneeman:
- CIP number to be determined (edit: assigned CIP 35)
- Especially seeking feedback regarding the `ethCompatible` JSON field, the uses of "SHOULD" relating to it, and whether there may be alternative approaches that would be preferable.


CIPs/cip-eth-compat.md
```diff
@@ -0,0 +1,95 @@
...
+
+```
+feecurrency: nil
+gatewayfeerecipient: nil
+gatewayfee: 0
+```
+
+### JSON representation
+
+When representing or parsing transactions in JSON format, clients SHOULD support an additional boolean field called `ethCompatible`, which indicates whether the transaction is an Ethereum-compatible transaction (`true`) or a normal Celo transaction (`false`).  This is necessary in order to know what should be used as the transaction's "signing data" (as described above).  If the field is absent in the JSON being parsed, clients SHOULD treat the transaction as if `ethCompatible` were `false`.
```
prestwich, line 64: what should node rpc behavior in `signTransaction` and `sendTransaction` be when `ethCompatible` is set to `true` but one of the 3 celo-specific fields is specified as a non-default value

prestwich: can we specify which RPC methods accept or return JSON transactions?

oneeman: Good points!  I will:

(a) add `ethCompatible` as an option for `eth_sendTransaction` and `eth_signTransaction`
(b) add validation in these methods, in the tx pool, and in the state transition to reject transactions with `ethCompatible` set to `true` if they have non-(nil or 0) values for the 3 Celo-only fields
(c) Update the CIP with these details
(d) Look into and call out these and other methods rather accept or receive JSON transactions (or 'partial' transaction JSON objects such as the arguments to `eth_signTransaction`)

oneeman: Added a commit making this section more complete and (I hope) clearer.  I didn't list all the RPC methods that include transactions in their results, but gave a couple as examples.

I also pushed to the celo-blockchain branch some commits to add (a) handling for `ethCompatible` in RPC methods which take in a transaction in JSON format, (b) validation against `ethCompatible` being true when the transaction has non-nil-or-0 values for the Celo-only fields.

prestwich (approved):
I think this is good to merge. may have more comments or clarifications later

YazzyYaz (approved):

PR #165, opened 2021-02-22
[CGP 0023] Draft: cEUR oracle activation

martinvol:
Crafting the proposal for cEUR oracle activation.

Still early work, please don't review at least requested.

# Staging:
Are the ones listed [here](https://github.com/celo-org/celo-monorepo/issues/6922#issuecomment-782178997)

# Baklava:
Oracles listed [here](https://github.com/celo-org/celo-monorepo/issues/7190#issuecomment-785721088)
StableTokenEUR address from the proposal:
```
{
    "contract": "Registry",
    "function": "setAddressFor",
    "args": [
      "StableTokenEUR",
      "0xf9ecE301247aD2CE21894941830A2470f4E774ca"
    ],
    "value": "0",
    "description": "Registry: StableTokenEUR -> 0xf9ecE301247aD2CE21894941830A2470f4E774ca"
  },
```
The file `CGPs/0036/staging.json` is meant to be the one used for this issue https://github.com/celo-org/celo-monorepo/issues/6922

Was proposed with ID=14

# Alfajores


CGPs/0036/staging.json
```diff
@@ -0,0 +1,152 @@
...
+      },
+      {
+        "contract": "SortedOracles",
+        "function": "addOracle",
+        "args": ["0x7Ba21D1f769Ae9feCF58dEfB5bD08905Bd3B707B"], // stableTokenEUR address
+        "value": "0"
+      },
+
+
+      // fund the oracles?
```
martinvol, line 48: how would the governance proposal would work to send CELO from the community fund to each oracle?
If risky or unnecessary, can be removed and we test manually

CGPs/0036/staging.json
```diff
@@ -0,0 +1,152 @@
...
+        "args": ["0x7Ba21D1f769Ae9feCF58dEfB5bD08905Bd3B707B"],
+        "value": "0"
+      },
+
+      // sent constitution parameters (for exchange) (same as the ones from governanceConstitution.js in monorepo)
+
+      {
+        "contract": "GovernanceProxy",
+        "function": "setConstitution",
+        "args": ["0x001E515b1902cAdAb233781eC73E9643B6c5012C", "", "0.9"], // setRegistry
```
martinvol, line 85: The second parameter is `bytes4 functionId`, how would I get that?

CGPs/0036/staging.json
```diff
@@ -0,0 +1,152 @@
...
+        "value": "0"
+      },
+      {
+        "contract": "GovernanceProxy",
+        "function": "",
+        "args": ["0x7Ba21D1f769Ae9feCF58dEfB5bD08905Bd3B707B"],
+        "value": "0"
+      },
+
+      // sent constitution parameters (for exchange) (same as the ones from governanceConstitution.js in monorepo)
```
martinvol, line 80: Shall this be in this proposal or should send somewhere else?

CGPs/0036/staging.json
```diff
@@ -0,0 +1,152 @@
...
+        "value": "0"
+      },
+      {
+        "contract": "GovernanceProxy",
+        "function": "setConstitution",
+        "args": ["0xcCA78ae41fe8F83eE7F37B0110eb6FAe3E9Ca0F0", "", "0.6"], // approve
+        "value": "0"
+      },
+    ]
+
```
martinvol, line 152: Is there a risk this proposal could get too big and may fail?

CGPs/0036/staging.json
```diff
@@ -0,0 +1,152 @@
...
+      "value": "0"
+    },
+    {
+        "contract": "Freezer",
+        "function": "freeze",
+        "args": ["0xcCA78ae41fe8F83eE7F37B0110eb6FAe3E9Ca0F0"], // stableTokenEUR address
+        "value": "0"
+      },
+
+
```
martinvol, line 15:
```suggestion
    // Add oracles for stableToenEUR
```

CGPs/0036/staging.json
```diff
@@ -0,0 +1,152 @@
+[
```
martinvol, line 1: Will delete all comments before making this an open PR (now it's a draft)

YazzyYaz:
This isn't a CIP so it shouldn't be called CIP-36. It's a CGP which is separate

CGPs/0036/staging.json
asaj: You just need the destination to be the oracle address, the function to be empty, the args to be empty, and the value to be whatever value you want to fund the oracle address with. And then one such tx for each oracle being funded

CGPs/0036/staging.json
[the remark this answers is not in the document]
asaj: The arguments are the arguments in the `setConstitution` function:
https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/governance/Governance.sol#L417

So it would be the address of the (proxy) contract for which we are setting the constitution

asaj: Docs here:
https://docs.soliditylang.org/en/v0.5.3/abi-spec.html#function-selector

CGPs/0036/staging.json
asaj: The block gas limit is a hard limit but I don't think we're in danger of getting close to that

CGPs/0023/staging.json
```diff
@@ -0,0 +1,140 @@
...
+    "args": ["0xcCA78ae41fe8F83eE7F37B0110eb6FAe3E9Ca0F0", "0xe1d6aceb", "600000000000000000000000"],
+    "value": "0"
+  },
+  {
+    "contract": "Governance",
+    "function": "setConstitution",
+    "args": ["0xcCA78ae41fe8F83eE7F37B0110eb6FAe3E9Ca0F0", "0x095ea7b3", "600000000000000000000000"],
+    "value": "0"
+  }
+]
```
martinvol, line 140: Todo add reserve and other new methods to the list of set constitution

YazzyYaz (approved):

PR #174, opened 2021-02-25
CIP-35: Add paragraph explaining replay protection through EIP-155

oneeman:
This adds a paragraph to CIP-35 clarifying that Celo and Ethereum both have replay-protection through EIP-155, so that any individual signed transaction is specific to one chain and can't later also be used on the other chain.

### DIFF
--- a/CIPs/cip-0035.md
+++ b/CIPs/cip-0035.md
@@ -94,6 +94,8 @@ The rest of the specification follows from that:
 
 This would not create new security issues, aside from the possibility (inherent in any change) of vulnerabilities resulting from bugs.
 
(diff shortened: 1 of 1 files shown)


prestwich (changes requested):
It's not accurate to say that a tx is never valid for both. Users can opt out of using EIP-155 and create replayable txns.

Is the network id correct? I thought it was 52752 (0xCE10)

prestwich:
> 52752 (0xCE10)

Huh. Just looked it up. This is the slip44 designation. I didn't know it was different from the network id

oneeman:
> It's not accurate to say that a tx is never valid for both. Users can opt out of using EIP-155 and create replayable txns.
> 
> Is the chain id correct? I thought it was 52752 (0xCE10)

My bad, I didn't notice it was `SHOULD` in EIP-155.  Do any/many Ethereum wallets expose ways to opt out of EIP-155?  I'm wondering how it should be phrased given that this actually is a possibility.

oneeman:
Actually, looking at the code, it appears Celo has never supported non-EIP155 transactions (though you could create a Celo testnet supporting it, but mainnet, alfajores, and baklava never supported it).  I will double-check.

prestwich:
> Actually, looking at the code, it appears Celo has never supported non-EIP155 transactions (though you could create a Celo testnet supporting it, but mainnet, alfajores, and baklava never supported it). I will double-check.

Very interesting. Might be good to update the language to call this out specifically, as it is a departure from expected behavior

in practice, nobody opts out

oneeman:
> > Actually, looking at the code, it appears Celo has never supported non-EIP155 transactions (though you could create a Celo testnet supporting it, but mainnet, alfajores, and baklava never supported it). I will double-check.
> 
> Very interesting. Might be good to update the language to call this out specifically, as it is a departure from expected behavior
> 
> in practice, nobody opts out

Actually, I spoke too soon based on the signer used being the EIP155 signer.  But of course when it comes to signature verification the signer supports verifying non-EIP155 signatures as well.

oneeman:
@prestwich Updated the EIP-155 section, please have a look.

prestwich (approved):

prestwich:
LGTM

PR #182, opened 2021-03-07
[CIP-0037] - Unambiguous name for ERC-20 on Celo chain

aaronmgdr:
Unambiguous name for ERC-20 on Celo chain


zviadm:
Yes! Big +1 to this effort. I have already ran into issues trying to figure out how to message about standardized tokens in Celo Terminal. Calling them ERC20 directly is just going to create further confusion. 

Some notes on actual name:
* Does it need to have "-20" in it? I am guessing most people don't know or care what either "ERC" or "20" is, so not sure if the name needs to have "-20" in it. 

Some other naming options to help with the discussion:
* Celo Standard Tokens (short: CSTs)
  - Celo Core Tokens (or Core Celo Tokens, short: CCTs) can refer to CELO + cUSD + cEUR....
* Celo Tokens (short: cTokens)
 - This can be confusing since cTokens sounds like it would just mean: cUSD/cEUR/...
 - This can also be confused with Compound cTokens
* Celo Fungible Tokens (short: CFTs)
 - This can be useful since it paves way for CNFTs, for Celo Non-Fungible Token standard.

aaronmgdr:
yes very much not set on any particular name. 

like @zviadm suggested I also was thinking abut Celo Standard Token or Celo Token Standard but could see it getting confused with CELO native token. 

I think the benefit of `20` is mostly because the implementations of the standard on other chains have it and it seems like a nice convention to stick with. but Its by no means required in my opinion.

patricknbaron:
I am in favor of keeping the naming convention as close to the original as possible, but clarifying that we're talking about Celo.   For example, ERC20 and ERC721 would be CRC20 and CRC721 respectively.  This approach is intuitive because it doesn't require a new user to learn a new naming convention.

aaronmgdr:
Going to bring this up at the core dev call this week.

CIPs/CIP-0037.md
```diff
@@ -0,0 +1,49 @@
+---
+cip: ?
+title: Unambiguous name for ERC-20 on Celo chain
+author: aaron deruvo, @aaronmgdr, Martin Chrzanowski, etc
```
YazzyYaz: Need Martin's email or github handle, whichever is more preferable.

YazzyYaz (changes requested):

YazzyYaz:
@zviadm @aaronmgdr @patricknbaron would you mind copying over your comments to the issue ticket? When this PR gets merged, your comments will be less easily viewed by the community https://github.com/celo-org/celo-proposals/issues/185

CIPs/CIP-0037.md
aaronmgdr: added

YazzyYaz (approved):

PR #192, opened 2021-03-15
Added cEUR mainnet proposals

martinvol:


CGPs/0023.md
```diff
@@ -10,43 +10,42 @@
...
+Following [CR3](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0022.md), cEUR will be in an status where it would be imposible to mint or burn due empty buckets that are not possible to update (because Mento buckets are zero and there are no oracles enable to updated), but cEUR will be possible to move. Freezing is meant to prevent cEUR from moving before enabling the stability mechanism.
 
-cEUR in circulation with no stability mechanism in place can lead to a depledge. Although the risk of cEUR moving before its full activation it's very low, this proposal will add an extra ledger of security. The only to address holding cEUR after CR3 is deployed will by the community fund (Governance Proxy) and a Celo Foundation wallet.
+cEUR in circulation with no stability mechanism in place can lead to a depledge. Although the risk of cEUR moving before its full activation it's very low, this proposal will add an extra layer of security. The only two addresses holding cEUR after CR3 is deployed will by the [Community Fund](https://celocommunityfund.org/) (Governance Proxy) and a Celo Foundation wallet.
```
bowd, line 15: nit: I might be wrong but I think the correct term is `depeg` / `de-peg`. Depledge is something a bit different from my understanding:


> depeg
> To remove a previously instituted peg on a currency. For example, if Currency A is pegged to Currency B at a 1:1 ratio, but the central bank for Currency A decides to let it float, it is said to be depegged from Currency B. Depegging may occur if the peg is causing inflation or if the central bank is unable to sustain the peg for other reasons.

---

> depledge
> (transitive, intransitive) To revoke one's pledge of allegiance to a college fraternity.

CGPs/0023.md
```diff
@@ -10,43 +10,42 @@
...
 [CGP #26](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0022.md) was merged into this proposal after Baklava deployment.
 
-## Proposed Changes
 
-Fill out the following template for each transaction in the proposal
+## Proposed Changes
 
-1. The first X transactions enable a specific address to report to StableTokenEUR
+1. The first ten transactions enable a specific address to report to StableTokenEUR
   - Destination: SortedOragles.addOracle
```
tkporter, line 25: SortedOragles -> SortedOracles

CGPs/0023.md
```diff
@@ -10,43 +10,42 @@
 
 This governance proposal enables oracle addresses to make them able to report the value of CELO/EUR for the cEUR contract, deployed on [CR3](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0022.md). It also proposed to send 100 CELO to each of the oracles to cover gas fees for the first couple of years and freezing StableTokenEUR (cEUR) and ExchangeEUR.
```
tkporter, line 11: I think `It also proposed` -> `It also proposes`

tkporter: Not super passionate but it may be good to rationalize the 100 CELO amount, eg how much we've seen existing oracles use and why 100 makes sense

CGPs/0023.md
```diff
@@ -10,43 +10,42 @@
...
+- Adding new oracles will add load to the network, which is not a problem at this stage but could become one as new stable assets and other projects deploy new oracles and network adoption grows.
+- Oracle rate risk: In case of imprecise CELO/USD oracle rates and/or manipulated CELO/USD market rates, users could exchange cUSD or CELO with the reserve at a price that does not reflect the current market valuation. This could negatively affect the reserve collateralization ratio and therefore long-term cUSD stability in two ways: First, a user could mint X cUSD while providing less than X USD worth of reserve collateral and second, a user could redeem Y cUSD for more than Y USD worth of reserve collateral. The major mitigant of this risk is the initial reduction of the reserveFraction parameter as explained in the overview section which comes at the cost of the risks described further below. Additional mitigants include oracle client features like a circuit breaker protecting against unreasonably large price movements. (Same as [CGP #0007](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0007.md))
```
bowd, line 47: Should this reference CELO/EUR cEUR, or if not make it clear why we're referencing that here, is it because they share the reserver?

CGPs/0023.md
tkporter, line 15: `CR3 is deployed will by` -> `CR3 is deployed will be`

bowd (approved):
Looking good! 💶

CGPs/0023.md
```diff
@@ -10,43 +10,42 @@
...
+ 
 [CGP #26](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0022.md) was merged into this proposal after Baklava deployment.
 
-## Proposed Changes
 
-Fill out the following template for each transaction in the proposal
+## Proposed Changes
 
-1. The first X transactions enable a specific address to report to StableTokenEUR
+1. The first ten transactions enable a specific address to report to StableTokenEUR
```
tkporter, line 24: nit: make it clear that this is enabling multiple addresses to report to SortedOracles for the StableTokenEUR rate

tkporter (approved):
All the proposal values look good to me

CGPs/0023.md
nvtaveras, line 11:
```suggestion
This governance proposal enables oracle addresses to make them able to report the value of CELO/EUR for the cEUR contract, deployed on [CR3](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0022.md). It also proposes to send 100 CELO to each of the oracles to cover gas fees for the first couple of years and freezing StableTokenEUR (cEUR) and ExchangeEUR.
```

CGPs/0023.md
martinvol, line 15:
```suggestion
cEUR in circulation with no stability mechanism in place can lead to a de-peg. Although the risk of cEUR moving before its full activation it's very low, this proposal will add an extra layer of security. The only two addresses holding cEUR after CR3 is deployed will be the [Community Fund](https://celocommunityfund.org/) (Governance Proxy) and a Celo Foundation wallet.
```

CGPs/0023.md
martinvol: Thanks! 🥴

CGPs/0023.md
```diff
@@ -10,43 +10,42 @@
 
 This governance proposal enables oracle addresses to make them able to report the value of CELO/EUR for the cEUR contract, deployed on [CR3](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0022.md). It also proposed to send 100 CELO to each of the oracles to cover gas fees for the first couple of years and freezing StableTokenEUR (cEUR) and ExchangeEUR.
 
-Following [CR3](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0022.md), cEUR will be in an status where it would be imposible to mint or burn due empty buckets that are not possible to update (assuming no oracle reports, but cEUR will be possible to move. Freezing is meant to prevent cEUR from moving before enabling the stability mechanism.
+Following [CR3](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0022.md), cEUR will be in an status where it would be imposible to mint or burn due empty buckets that are not possible to update (because Mento buckets are zero and there are no oracles enable to updated), but cEUR will be possible to move. Freezing is meant to prevent cEUR from moving before enabling the stability mechanism.
```
nvtaveras, line 13:
```suggestion
Following [CR3](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0022.md), cEUR will be in a status where it will be imposible to mint or burn due to empty buckets that are not possible to update (because Mento buckets are zero and there are no oracles enabled to update them), but cEUR will be transferable. Freezing is meant to prevent cEUR being transferred before enabling the stability mechanism.
```

CGPs/0023.md
martinvol: True!

martinvol, line 47:
```suggestion
- Oracle rate risk: In case of imprecise CELO/EUR oracle rates and/or manipulated CELO/EUR market rates, users could exchange cEUR or CELO with the reserve at a price that does not reflect the current market valuation. This could negatively affect the reserve collateralization ratio and therefore long-term cEUR stability in two ways: First, a user could mint X cEUR while providing less than X EUR worth of reserve collateral and second, a user could redeem Y cEUR for more than Y EUR worth of reserve collateral. The major mitigant of this risk is the initial reduction of the reserveFraction parameter as explained in the overview section which comes at the cost of the risks described further below. Additional mitigants include oracle client features like a circuit breaker protecting against unreasonably large price movements. (Same as [CGP #0007](https://github.com/celo-org/celo-proposals/blob/master/CGPs/0007.md))
```

YazzyYaz (approved):

PR #204, opened 2021-04-08
Added Celo Forum discussion link

albertclabs:


albertclabs:
Added to Pedro's branch instead.

PR #222, opened 2021-05-07
[TBD] Reserve Voting

asaj:


zviadm:
Two quick notes on first glance:
* There is already too much locked CELO which is leading to ever decreasing yield on locked and voting CELO. This will significantly speed up inflation and cause rewards to decrease further. That doesn't seem great. Also this seems to be pretty bad for people who have locked CELO and good for people who don't have locked CELO. So I don't think people will vote against their direct interests to get this to pass?

* I think this will either make it very difficult to elect a new validator group, or it will open up opportunities for shenanigans to get validators elected right before epoch switches: 
-> Here is on example:
--> Lets say I have direct control over votes to elect 2 validators. I create two validator groups A0, A1.
--> First i use these votes to elect A0 with 2 validators and i get all the proportional reserve votes applied to my group.
--> Right before next epoch switch, I shift all my votes to not-yet elected A1 group, I also get reserve to move proportional votes from some other lowest voted groups to A1.
--> Before epoch switches, I shift back all my own votes back to A0. So now I have A0 with my own votes + reserve votes. A1 with reserve votes. It is very likely that I will get A1 group elected too with at least with 1 validator, assuming reserve is casting significant amount of votes overall.
--> This attack gets worse if attacker controls more votes directly. And you actually don't have to even stop at just having 1 extra group, you can have many other groups like A2, A3, A4 and get them all proportional reserve votes too right before epoch switches and get them all elected. 

If you try to patch this attack, I think it will then cause another issue. It will become very difficult to get new validator group elected because you wont be able to shift reserve votes until your new group is already electted.

asaj:
Thanks @zviadm! There are other compelling reasons why we shouldn't do this that we can argue regardless of the implementation.

The reason that the protocol offers rewards to voters is to help secure the network by raising the barrier to entry into the validator set. If our goal is to design a reserve voting protocol that meets the definition of impartiality laid out in the CIP, reserve voting is actually providing no utility to the network. And so if our goal is to protect the reserve assets from inflation, there are less complicated ways of achieving that, e.g. by reducing the target yield to voters or by redirecting epoch rewards to the reserve directly.

asaj:
> There is already too much locked CELO which is leading to ever decreasing yield on locked and voting CELO. This will significantly speed up inflation and cause rewards to decrease further. That doesn't seem great. Also this seems to be pretty bad for people who have locked CELO and good for people who don't have locked CELO. So I don't think people will vote against their direct interests to get this to pass?


This is an interesting comment. I'm actually of the belief that voting rewards *should* be decreased; rewards are only useful in so far as they incentivize people to vote, and it's not clear to me that we wouldn't see a similar amount of voting with lower yields. In effect, holders of CELO that don't vote subsidize this yield, and by far the largest of those is the reserve. 

But perhaps that's a discussion to save for another time.

PR #224, opened 2021-05-12
[Draft] Granda Mento

tkporter:
Very incomplete start of Granda Mento

### DIFF
--- a/CIPs/cip-granda_mento.md
+++ b/CIPs/cip-granda_mento.md
@@ -20,7 +20,7 @@ use an abbreviated title in the filename, `cip-draft_title_abbrev.md`.
 ## Simple Summary
 > If you can't explain it simply, you don't understand it well enough. Provide a simplified and layman-accessible explanation of the CIP.
 
-Granda Mento is a mechanism to facilitate large `CELO <-> stable token` trades that aren't suitable for Mento or OTC. A new contract is created that has the authorization by Governance to trade a limited amount of `CELO <-> stable token`. Trades via this contract must be approved by a multisig and can be vetoed by Governance.
+Granda Mento is a mechanism to facilitate large CELO <-> stable token trades that aren't suitable for Mento or OTC. A new contract is created that has the authorization by Governance to trade a limited amount of `CELO <-> stable token`. Trades via this contract must be approved by a multisig and can be vetoed by Governance.
 
 ## Abstract
 > A short (~200 word) description of the technical issue being addressed.
@@ -28,7 +28,7 @@ Granda Mento is a mechanism to facilitate large `CELO <-> stable token` trades t
 ## Motivation
 > The motivation is critical for CIPs that want to change the Celo protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the CIP solves. CIP submissions without sufficient motivation may be rejected outright.
 
-The existing implementation of Mento is suitable for facilitating low volume trades, e.g. up to tens of thousands at a time, without incurring slippage greater than a few percent. OTC trading satisfies the needs of medium volume trades, eg ~100k for cUSD and likely much less for stable tokens with lower total supply (like cEUR), but is unable to satisfy high volume `CELO <-> stable token trades` on the order of millions.
+The existing implementation of Mento is suitable for facilitating low volume trades, e.g. up to tens of thousands at a time, without incurring slippage greater than a few percent. OTC trading satisfies the needs of medium volume trades, eg ~100k for cUSD and likely much less for stable tokens with lower total supply (like cEUR), but is unable to satisfy high volume CELO <-> stable token trades on the order of millions.
 
 Mento is able to facilitate smaller `CELO <-> stable token` trades with minimal slippage for smaller trade sizes. However, significant slippage of over 2% starts to occur on both Mento and on centralized exchanges at trades sizes of ~$50k+. OTC trading satisfies the needs of medium volume trades, eg ~100k, but similarly takes a hit of around 2-3%.
 
@@ -50,9 +50,9 @@ The proposed implementation is being considered in the short-term for large scal
 
 At a high level, the design involves:
 1. Anyone can submit a trade proposal.
-   1. The size of the trade is required to be sufficiently large.
-   2. The assets being sold in the trade are deposited.
-   3. The current oracle price for the trade is recorded.
+   * The size of the trade is required to be sufficiently large.
+   * The assets being sold in the trade are deposited.
+   * The current oracle price for the trade is recorded.
 2. The proposed trade must be approved by a multisig that has previously been authorized by Governance. At this point, the trade still cannot be executed.
 3. A forced waiting period of X days must elapse before the trade can be executed. During this time, Governance can choose to veto the trade, refunding the trader.
 4. After the waiting period, the trade is executed.
@@ -73,35 +73,35 @@ The contract has the following configurable parameters:
 The contract has the following functions:
 
 1. **Propose trade** - Called by a trader to initiate a trade.
-   1. Callable by anyone.
-   2. Requires the trade to be larger than the minimum stable token buy/sell quantity.
-   3. Deposits the full amount of the asset being sold into the contract.
-   4. Records:
-     1. Which asset is being sold
-     2. The amount of the asset being sold
+   * Callable by anyone.
+   * Requires the trade to be larger than the minimum stable token buy/sell quantity.
+   * Deposits the full amount of the asset being sold into the contract.
+   * Records:
+     * Which asset is being sold
+     * The amount of the asset being sold
 2. **Approve trade proposal** - Approves a proposed trade.
-   1. Only callable by the approver address.
-   2. Marks a proposed trade as approved.
-   3. Records:
-     1. The timestamp at which the approval for the proposed trade occurred.
+   * Only callable by the approver address.
+   * Marks a proposed trade as approved.
+   * Records:
+     * The timestamp at which the approval for the proposed trade occurred.
 3. **Veto an approved trade proposal** - Vetoes a previously approved proposed trade.
    1. Only callable by governance.
    2. Refunds the proposed trade's deposited sell asset.
    3. Records:
      1. The trade as vetoed.
 4. **Execute an approved and ready trade** - Executes a trade.
-   1. Callable by anyone.
-   2. Requires the trade to have been approved.
-   3. Requires the required waiting period for an approved trade to have elapsed since the time it was approved.
-   4. Makes the trade:
-     1. If selling CELO and purchasing stable token:
-       1. The CELO is sent to the Reserve.
-       2. Stable token is minted to the trader according to the rate originally recorded when the trade was proposed.
-     2. If selling stable token and purchasing CELO:
-       1. The stable token is burned.
-       1. CELO is transferred from the Reserve to the trader according to the rate originally recorded when the trade was proposed.
-   5. Records:
-     1. The trade as executed.
+   * Callable by anyone.
+   * Requires the trade to have been approved.
+   * Requires the required waiting period for an approved trade to have elapsed since the time it was approved.
+   * Makes the trade:
+     * If selling CELO and purchasing stable token:
+       * The CELO is sent to the Reserve.
+       * Stable token is minted to the trader according to the rate originally recorded when the trade was proposed.
+     * If selling stable token and purchasing CELO:
+       * The stable token is burned.
+       * CELO is transferred from the Reserve to the trader according to the rate originally recorded when the trade was proposed.
+   * Records:
+     * The trade as executed.
 
 #### Modifications of existing contracts:
 
@@ -113,11 +113,11 @@ The following modifications will be made:
    2. Records:
    3. The new mint/burn allowance for the specified address
 2. **Modification to [mint()](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L224)**
-   1. If `msg.sender` is not the Exchange or `Validators.sol`, require `msg.sender` to have sufficient mint/burn allowance.
-   2. Update the mint/burn allowance of `msg.sender` if appropriate
+   * If `msg.sender` is not the Exchange or `Validators.sol`, require `msg.sender` to have sufficient mint/burn allowance.
+   * Update the mint/burn allowance of `msg.sender` if appropriate
 3. **Modification to [burn()](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L273)**
-   1. If `msg.sender` is not the Exchange, require `msg.sender` to have sufficient mint/burn allowance.
-   2. Update the mint/burn allowance of `msg.sender` if appropriate
+   * If `msg.sender` is not the Exchange, require `msg.sender` to have sufficient mint/burn allowance.
+   * Update the mint/burn allowance of `msg.sender` if appropriate
 
 ## Rationale
 The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.
--- a/CIPs/cip-granda_mento.md
+++ b/CIPs/cip-granda_mento.md
@@ -85,10 +85,10 @@ The contract has the following functions:
    * Records:
      * The timestamp at which the approval for the proposed trade occurred.
 3. **Veto an approved trade proposal** - Vetoes a previously approved proposed trade.
-   1. Only callable by governance.
-   2. Refunds the proposed trade's deposited sell asset.
-   3. Records:
-     1. The trade as vetoed.
+   * Only callable by governance.
+   * Refunds the proposed trade's deposited sell asset.
+   * Records:
+     * The trade as vetoed.
 4. **Execute an approved and ready trade** - Executes a trade.
    * Callable by anyone.
    * Requires the trade to have been approved.
@@ -109,9 +109,9 @@ Currently, `StableToken.sol` only allows its `Exchange.sol` or `Validators.sol` 
 
 The following modifications will be made:
 1. **Set mint/burn allowance function** - Sets an allowance for an address to mint or burn an amount of tokens.
-   1. Only callable by governance.
-   2. Records:
-   3. The new mint/burn allowance for the specified address
+   * Only callable by governance.
+   * Records:
+   * The new mint/burn allowance for the specified address
 2. **Modification to [mint()](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L224)**
    * If `msg.sender` is not the Exchange or `Validators.sol`, require `msg.sender` to have sufficient mint/burn allowance.
    * Update the mint/burn allowance of `msg.sender` if appropriate
--- a/CIPs/cip-granda_mento.md
+++ b/CIPs/cip-granda_mento.md
@@ -20,7 +20,7 @@ use an abbreviated title in the filename, `cip-draft_title_abbrev.md`.
 ## Simple Summary
 > If you can't explain it simply, you don't understand it well enough. Provide a simplified and layman-accessible explanation of the CIP.
 
-Granda Mento is a mechanism to facilitate large CELO <-> stable token trades that aren't suitable for Mento or OTC. A new contract is created that has the authorization by Governance to trade a limited amount of `CELO <-> stable token`. Trades via this contract must be approved by a multisig and can be vetoed by Governance.
+Granda Mento is a mechanism to facilitate large `CELO <-> stable token` trades that aren't suitable for Mento or OTC. A new contract is created that has the authorization by Governance to trade a limited amount of `CELO <-> stable token`. Trades via this contract must be approved by a multisig and can be vetoed by Governance.
 
 ## Abstract
 > A short (~200 word) description of the technical issue being addressed.
@@ -28,7 +28,7 @@ Granda Mento is a mechanism to facilitate large CELO <-> stable token trades tha
 ## Motivation
 > The motivation is critical for CIPs that want to change the Celo protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the CIP solves. CIP submissions without sufficient motivation may be rejected outright.
 
-The existing implementation of Mento is suitable for facilitating low volume trades, e.g. up to tens of thousands at a time, without incurring slippage greater than a few percent. OTC trading satisfies the needs of medium volume trades, eg ~100k for cUSD and likely much less for stable tokens with lower total supply (like cEUR), but is unable to satisfy high volume CELO <-> stable token trades on the order of millions.
+The existing implementation of Mento is suitable for facilitating low volume trades, e.g. up to tens of thousands at a time, without incurring slippage greater than a few percent. OTC trading satisfies the needs of medium volume trades, eg ~100k for cUSD and likely much less for stable tokens with lower total supply (like cEUR), but is unable to satisfy high volume `CELO <-> stable token trades` on the order of millions.
 
 Mento is able to facilitate smaller `CELO <-> stable token` trades with minimal slippage for smaller trade sizes. However, significant slippage of over 2% starts to occur on both Mento and on centralized exchanges at trades sizes of ~$50k+. OTC trading satisfies the needs of medium volume trades, eg ~100k, but similarly takes a hit of around 2-3%.
 
@@ -50,9 +50,9 @@ The proposed implementation is being considered in the short-term for large scal
 
 At a high level, the design involves:
 1. Anyone can submit a trade proposal.
-  a. The size of the trade is required to be sufficiently large.
-  b. The assets being sold in the trade are deposited.
-  c. The current oracle price for the trade is recorded.
+   1. The size of the trade is required to be sufficiently large.
+   2. The assets being sold in the trade are deposited.
+   3. The current oracle price for the trade is recorded.
 2. The proposed trade must be approved by a multisig that has previously been authorized by Governance. At this point, the trade still cannot be executed.
 3. A forced waiting period of X days must elapse before the trade can be executed. During this time, Governance can choose to veto the trade, refunding the trader.
 4. After the waiting period, the trade is executed.
@@ -73,35 +73,35 @@ The contract has the following configurable parameters:
 The contract has the following functions:
 
 1. **Propose trade** - Called by a trader to initiate a trade.
-  a. Callable by anyone.
-  b. Requires the trade to be larger than the minimum stable token buy/sell quantity.
-  c. Deposits the full amount of the asset being sold into the contract.
-  d. Records:
-    a. Which asset is being sold
-    b. The amount of the asset being sold
+   1. Callable by anyone.
+   2. Requires the trade to be larger than the minimum stable token buy/sell quantity.
+   3. Deposits the full amount of the asset being sold into the contract.
+   4. Records:
+     1. Which asset is being sold
+     2. The amount of the asset being sold
 2. **Approve trade proposal** - Approves a proposed trade.
-  a. Only callable by the approver address.
-  b. Marks a proposed trade as approved.
-  c. Records:
-    a. The timestamp at which the approval for the proposed trade occurred.
+   1. Only callable by the approver address.
+   2. Marks a proposed trade as approved.
+   3. Records:
+     1. The timestamp at which the approval for the proposed trade occurred.
 3. **Veto an approved trade proposal** - Vetoes a previously approved proposed trade.
-  a. Only callable by governance.
-  b. Refunds the proposed trade's deposited sell asset.
-  c. Records:
-    a. The trade as vetoed.
+   1. Only callable by governance.
+   2. Refunds the proposed trade's deposited sell asset.
+   3. Records:
+     1. The trade as vetoed.
 4. **Execute an approved and ready trade** - Executes a trade.
-  a. Callable by anyone.
-  b. Requires the trade to have been approved.
-  c. Requires the required waiting period for an approved trade to have elapsed since the time it was approved.
-  d. Makes the trade:
-    a. If selling CELO and purchasing stable token:
-      a. The CELO is sent to the Reserve.
-      b. Stable token is minted to the trader according to the rate originally recorded when the trade was proposed.
-    b. If selling stable token and purchasing CELO:
-      a. The stable token is burned.
-      b. CELO is transferred from the Reserve to the trader according to the rate originally recorded when the trade was proposed.
-  e. Records:
-    a. The trade as executed.
+   1. Callable by anyone.
+   2. Requires the trade to have been approved.
+   3. Requires the required waiting period for an approved trade to have elapsed since the time it was approved.
+   4. Makes the trade:
+     1. If selling CELO and purchasing stable token:
+       1. The CELO is sent to the Reserve.
+       2. Stable token is minted to the trader according to the rate originally recorded when the trade was proposed.
+     2. If selling stable token and purchasing CELO:
+       1. The stable token is burned.
+       1. CELO is transferred from the Reserve to the trader according to the rate originally recorded when the trade was proposed.
+   5. Records:
+     1. The trade as executed.
 
 #### Modifications of existing contracts:
 
@@ -109,15 +109,15 @@ Currently, `StableToken.sol` only allows its `Exchange.sol` or `Validators.sol` 
 
 The following modifications will be made:
 1. **Set mint/burn allowance function** - Sets an allowance for an address to mint or burn an amount of tokens.
-  a. Only callable by governance.
-  b. Records:
-    a. The new mint/burn allowance for the specified address
+   1. Only callable by governance.
+   2. Records:
+   3. The new mint/burn allowance for the specified address
 2. **Modification to [mint()](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L224)**
-  a. If `msg.sender` is not the Exchange or `Validators.sol`, require `msg.sender` to have sufficient mint/burn allowance.
-  b. Update the mint/burn allowance of `msg.sender` if appropriate
+   1. If `msg.sender` is not the Exchange or `Validators.sol`, require `msg.sender` to have sufficient mint/burn allowance.
+   2. Update the mint/burn allowance of `msg.sender` if appropriate
 3. **Modification to [burn()](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L273)**
-  a. If `msg.sender` is not the Exchange, require `msg.sender` to have sufficient mint/burn allowance.
-  b. Update the mint/burn allowance of `msg.sender` if appropriate
+   1. If `msg.sender` is not the Exchange, require `msg.sender` to have sufficient mint/burn allowance.
+   2. Update the mint/burn allowance of `msg.sender` if appropriate
 
 ## Rationale
 The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.
--- a/CIPs/cip-granda_mento.md
+++ b/CIPs/cip-granda_mento.md
@@ -1,0 +1,138 @@
+---
+cip: <to be assigned>
+title: Granda Mento
+author: <a list of the author's or authors' name(s) and/or username(s), or name(s) and email(s), e.g. (use with the parentheses or triangular brackets): FirstName LastName (@GitHubUsername), FirstName LastName <foo@bar.com>, FirstName (@GitHubUsername) and GitHubUsername (@GitHubUsername)>
+discussions-to: <URL>
+status: Draft
+type: <Standards Track, Meta or Informational>
+category (*only required for Standards Track): <Ring 0, 1, 2, 3>
+created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
+requires (*optional): <CIP number(s)>
+replaces (*optional): <CIP number(s)>
+license: Apache 2.0
+---
+
+This is the suggested template for new CIPs.
+
+Note that a CIP number will be assigned by an editor. When opening a pull request to submit your CIP, please
+use an abbreviated title in the filename, `cip-draft_title_abbrev.md`.
+
+## Simple Summary
+> If you can't explain it simply, you don't understand it well enough. Provide a simplified and layman-accessible explanation of the CIP.
+
+Granda Mento is a mechanism to facilitate large CELO <-> stable token trades that aren't suitable for Mento or OTC. A new contract is created that has the authorization by Governance to trade a limited amount of `CELO <-> stable token`. Trades via this contract must be approved by a multisig and can be vetoed by Governance.
+
+## Abstract
+> A short (~200 word) description of the technical issue being addressed.
+
+## Motivation
+> The motivation is critical for CIPs that want to change the Celo protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the CIP solves. CIP submissions without sufficient motivation may be rejected outright.
+
+The existing implementation of Mento is suitable for facilitating low volume trades, e.g. up to tens of thousands at a time, without incurring slippage greater than a few percent. OTC trading satisfies the needs of medium volume trades, eg ~100k for cUSD and likely much less for stable tokens with lower total supply (like cEUR), but is unable to satisfy high volume CELO <-> stable token trades on the order of millions.
+
+Mento is able to facilitate smaller `CELO <-> stable token` trades with minimal slippage for smaller trade sizes. However, significant slippage of over 2% starts to occur on both Mento and on centralized exchanges at trades sizes of ~$50k+. OTC trading satisfies the needs of medium volume trades, eg ~100k, but similarly takes a hit of around 2-3%.
+
+Users looking to make larger trades are faced with 3 options:
+1. Exchange via Mento slowly over a longer period of time
+2. Arrange one or many OTC trades
+3. Create a large limit order in hopes it gets filled over time
+
+Increasing Mento bucket sizes permanently to allow larger trades via Mento will reduce slippage for (1), but it puts a more significant portion of the Reserve at risk.
+
+Building a process to exchange large (~$1 million+) amounts of CELO for stable tokens will enable large entities to make significant block purchases of stable tokens at scale. While minting of the stable token is expected to be the main use case, reversibility (ie trading `stable token -> CELO`) is desirable to instill higher confidence when trading large quantities.
+
+While Granda Mento will help facilitate `CELO <-> stable token` trades, this is ultimately meant as a mechanism to help purchasers with USD who are seeking to own cUSD. Purchasers would come to an agreement with a broker (eg the Celo Foundation) who owns an existing amount of CELO. The broker would be the one to make the `CELO -> cUSD` trade on-chain.
+
+The proposed implementation is being considered in the short-term for large scale trades until Mento is reworked to support more efficient large-scale minting on a sustained basis.
+
+## Specification
+> The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow for an implementation on the Celo platform (go-celo).
+
+At a high level, the design involves:
+1. Anyone can submit a trade proposal.
+  a. The size of the trade is required to be sufficiently large.
+  b. The assets being sold in the trade are deposited.
+  c. The current oracle price for the trade is recorded.
+2. The proposed trade must be approved by a multisig that has previously been authorized by Governance. At this point, the trade still cannot be executed.
+3. A forced waiting period of X days must elapse before the trade can be executed. During this time, Governance can choose to veto the trade, refunding the trader.
+4. After the waiting period, the trade is executed.
+
+### Specific Changes:
+
+#### `GrandaMento.sol`
+
+A new contract, `GrandaMento.sol`, is created. One deployment of `GrandaMento` exists per stable token. It must have the ability to mint/burn the corresponding stable token and be a spender of `Reserve.sol`.
+
+The contract has the following configurable parameters:
+
+1. **Approver address** - Set by governance. Intended to be a multisig and has the authority to approve trades.
+2. **Minimum stable token buy/sell quantity for trades** - Set by governance. Minimum amount of stable token a trade must buy/sell.
+3. **Required waiting period for an approved trade before execution** - Set by governance. The minimum amount of time that must elapse between a proposed trade being approved and when the trade can be executed. Should give sufficient time for Governance to veto an approved trade.
+4. **Spread imposed on trades** - Set by governance. The percent fee imposed upon the trade.
+
+The contract has the following functions:
+
+1. **Propose trade** - Called by a trader to initiate a trade.
+  a. Callable by anyone.
+  b. Requires the trade to be larger than the minimum stable token buy/sell quantity.
+  c. Deposits the full amount of the asset being sold into the contract.
+  d. Records:
+    a. Which asset is being sold
+    b. The amount of the asset being sold
+2. **Approve trade proposal** - Approves a proposed trade.
+  a. Only callable by the approver address.
+  b. Marks a proposed trade as approved.
+  c. Records:
+    a. The timestamp at which the approval for the proposed trade occurred.
+3. **Veto an approved trade proposal** - Vetoes a previously approved proposed trade.
+  a. Only callable by governance.
+  b. Refunds the proposed trade's deposited sell asset.
+  c. Records:
+    a. The trade as vetoed.
+4. **Execute an approved and ready trade** - Executes a trade.
+  a. Callable by anyone.
+  b. Requires the trade to have been approved.
+  c. Requires the required waiting period for an approved trade to have elapsed since the time it was approved.
+  d. Makes the trade:
+    a. If selling CELO and purchasing stable token:
+      a. The CELO is sent to the Reserve.
+      b. Stable token is minted to the trader according to the rate originally recorded when the trade was proposed.
+    b. If selling stable token and purchasing CELO:
+      a. The stable token is burned.
+      b. CELO is transferred from the Reserve to the trader according to the rate originally recorded when the trade was proposed.
+  e. Records:
+    a. The trade as executed.
+
+#### Modifications of existing contracts:
+
+Currently, `StableToken.sol` only allows its `Exchange.sol` or `Validators.sol` to [mint](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L225) new stable tokens. Similarly, only its `Exchange.sol` can [burn](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L275) stable tokens.
+
+The following modifications will be made:
+1. **Set mint/burn allowance function** - Sets an allowance for an address to mint or burn an amount of tokens.
+  a. Only callable by governance.
+  b. Records:
+    a. The new mint/burn allowance for the specified address
+2. **Modification to [mint()](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L224)**
+  a. If `msg.sender` is not the Exchange or `Validators.sol`, require `msg.sender` to have sufficient mint/burn allowance.
+  b. Update the mint/burn allowance of `msg.sender` if appropriate
+3. **Modification to [burn()](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L273)**
+  a. If `msg.sender` is not the Exchange, require `msg.sender` to have sufficient mint/burn allowance.
+  b. Update the mint/burn allowance of `msg.sender` if appropriate
+
+## Rationale
+The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.
+
+## Backwards Compatibility
+All CIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The CIP must explain how the author proposes to deal with these incompatibilities. CIP submissions without a sufficient backwards compatibility treatise may be rejected outright.
+
+## Test Cases
+Test cases for an implementation are mandatory for CIPs that are affecting consensus changes. Other CIPs can choose to include links to test cases if applicable.
+
+## Implementation
+The implementations must be completed before any CIP is given status "Final", but it need not be completed before the CIP is accepted. While there is merit to the approach of reaching consensus on the specification and rationale before writing code, the principle of "rough consensus and running code" is still useful when it comes to resolving many discussions of API details.
+
+## Security Considerations
+All CIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. CIP submissions missing the "Security Considerations" section will be rejected. A CIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.
+
+## License
+This work is licensed under the Apache License, Version 2.0.


CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,138 @@
...
+       * The stable token is burned.
+       * CELO is transferred from the Reserve to the trader according to the rate originally recorded when the trade was proposed.
+   * Records:
+     * The trade as executed.
+
+#### Modifications of existing contracts:
+
+Currently, `StableToken.sol` only allows its `Exchange.sol` or `Validators.sol` to [mint](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L225) new stable tokens. Similarly, only its `Exchange.sol` can [burn](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L275) stable tokens.
+
+The following modifications will be made:
```
tkporter, line 110: An alternative implementation could have the "mint/burn" limit be set in Granda Mento by Governance.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,140 @@
...
+#### `GrandaMento.sol`
+
+A new contract, `GrandaMento.sol`, is created. One deployment of `GrandaMento` exists per stable token. It must have the ability to mint/burn the corresponding stable token and be a spender of `Reserve.sol`.
+
+The contract has the following configurable parameters:
+
+1. **Approver address** - Set by governance. Intended to be a multisig and has the authority to approve trades.
+2. **Minimum stable token buy/sell quantity for trades** - Set by governance. Minimum amount of stable token a trade must buy/sell.
+3. **Required waiting period for an approved trade before execution** - Set by governance. The minimum amount of time that must elapse between a proposed trade being approved and when the trade can be executed. Should give sufficient time for Governance to veto an approved trade.
+4. **Spread imposed on trades** - Set by governance. The percent fee imposed upon the trade.
```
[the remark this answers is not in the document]
tobikuhlmann: In theory we could just choose a period e.g. days from proposal to execution, however I gave it some more thought and it would be unnecessarily complex. My intent with that was just to suggest it's not cheaper than Mento out of fairness to regular users as it is probably a temporary feature.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,140 @@
...
+### Specific Changes:
+
+#### `GrandaMento.sol`
+
+A new contract, `GrandaMento.sol`, is created. One deployment of `GrandaMento` exists per stable token. It must have the ability to mint/burn the corresponding stable token and be a spender of `Reserve.sol`.
+
+The contract has the following configurable parameters:
+
+1. **Approver address** - Set by governance. Intended to be a multisig and has the authority to approve trades.
+2. **Minimum stable token buy/sell quantity for trades** - Set by governance. Minimum amount of stable token a trade must buy/sell.
```
tobikuhlmann, line 69: **Does it make sense to add a maximum stable token buy/sell quantity?** I don't know too much about attacks and I am not sure how the multisig approval works and would look like, but a couple of thoughts and questions come into my mind: 

- That would catch fat-finger trades automatically before approval or governance
- If approval and governance veto would fail ... this would protect the reserve 

Might there be downsides that I don't see?

We could suggest a maximum quantity to governance such that trade sizes in that region would not happen anyways in the next months (e.g. 50 or 100mio cUSD), but it would still protect the reserve from CELO depletion

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,140 @@
...
+2. Arrange one or many OTC trades
+3. Create a large limit order in hopes it gets filled over time
+
+Increasing Mento bucket sizes permanently to allow larger trades via Mento will reduce slippage for (1), but it puts a more significant portion of the Reserve at risk.
+
+Building a process to exchange large (~$1 million+) amounts of CELO for stable tokens will enable large entities to make significant block purchases of stable tokens at scale. While minting of the stable token is expected to be the main use case, reversibility (ie trading `stable token -> CELO`) is desirable to instill higher confidence when trading large quantities.
+
+While Granda Mento will help facilitate `CELO <-> stable token` trades, this is ultimately meant as a mechanism to help purchasers with USD who are seeking to own cUSD. Purchasers would come to an agreement with a broker (eg the Celo Foundation) who owns an existing amount of CELO. The broker would be the one to make the `CELO -> cUSD` trade on-chain.
+
+The proposed implementation is being considered in the short-term for large scale trades until Mento is reworked to support more efficient large-scale minting on a sustained basis.
```
asaj, line 46: Probably more medium term, while I'm super excited to improve Mento I can't imagine substantial changes coming in the short term.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Specification
+> The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow for an implementation on the Celo platform (go-celo).
+
+At a high level, the design involves:
+1. Anyone can submit a trade proposal.
```
asaj, line 52: nit: suggest "exchange" for consistency

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,140 @@
...
+   * The size of the trade is required to be sufficiently large.
```
asaj, line 53: What does this mean? Be more specific as to how "large" is measured

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,140 @@
...
+   * The assets being sold in the trade are deposited.
+   * The current oracle price for the trade is recorded.
+2. The proposed trade must be approved by a multisig that has previously been authorized by Governance. At this point, the trade still cannot be executed.
+3. A forced waiting period of X days must elapse before the trade can be executed. During this time, Governance can choose to veto the trade, refunding the trader.
```
asaj, line 57: nit: proposer

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,140 @@
...
+The proposed implementation is being considered in the short-term for large scale trades until Mento is reworked to support more efficient large-scale minting on a sustained basis.
+
+## Specification
+> The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow for an implementation on the Celo platform (go-celo).
+
+At a high level, the design involves:
+1. Anyone can submit a trade proposal.
+   * The size of the trade is required to be sufficiently large.
+   * The assets being sold in the trade are deposited.
+   * The current oracle price for the trade is recorded.
```
asaj, line 55: Should there be more discretion here? e.g. the contract enforces that the proposed price is within x% of the current oracle rate

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,140 @@
...
+2. The proposed trade must be approved by a multisig that has previously been authorized by Governance. At this point, the trade still cannot be executed.
+3. A forced waiting period of X days must elapse before the trade can be executed. During this time, Governance can choose to veto the trade, refunding the trader.
+4. After the waiting period, the trade is executed.
+
+### Specific Changes:
+
+#### `GrandaMento.sol`
+
+A new contract, `GrandaMento.sol`, is created. One deployment of `GrandaMento` exists per stable token. It must have the ability to mint/burn the corresponding stable token and be a spender of `Reserve.sol`.
```
asaj, line 64: How will this work? I believe `StableToken`s are currently only configured to support a single minter.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,140 @@
...
+
+The contract has the following configurable parameters:
+
+1. **Approver address** - Set by governance. Intended to be a multisig and has the authority to approve trades.
+2. **Minimum stable token buy/sell quantity for trades** - Set by governance. Minimum amount of stable token a trade must buy/sell.
+3. **Required waiting period for an approved trade before execution** - Set by governance. The minimum amount of time that must elapse between a proposed trade being approved and when the trade can be executed. Should give sufficient time for Governance to veto an approved trade.
+4. **Spread imposed on trades** - Set by governance. The percent fee imposed upon the trade.
+
+The contract has the following functions:
```
asaj, line 73: Consider defining the ABI explicitly.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,140 @@
...
+       * CELO is transferred from the Reserve to the trader according to the rate originally recorded when the trade was proposed.
+   * Records:
+     * The trade as executed.
+
+#### Modifications of existing contracts:
+
+Currently, `StableToken.sol` only allows its `Exchange.sol` or `Validators.sol` to [mint](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L225) new stable tokens. Similarly, only its `Exchange.sol` can [burn](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L275) stable tokens.
+
+The following modifications will be made:
+1. **Set mint/burn allowance function** - Sets an allowance for an address to mint or burn an amount of tokens.
```
asaj, line 111: Does this mean that governance would need to "top up" granda mento from time to time?

asaj, line 111: Why not a more general approach? e.g. store a list of registry IDs of contracts allowed to mint stable tokens?

CIPs/cip-granda_mento.md
tkporter: Is this because there is a desire for the exchange proposer to specify the exact price rather than to rely upon the current oracle median as the price?

CIPs/cip-granda_mento.md
tkporter: with this approach, yes

tkporter: That definitely works-- I came out of our brainstorm under the impression that Granda Mento would have a governance-set allowance. I think there are a few questions:

1. Should Granda Mento have a cap for how many stable tokens it can mint/burn?
2. Should that cap be an allowance or just a constant to avoid fat fingers or crazy large exchanges?
3. Should the cap be imposed by StableToken or GrandaMento?

Imo the easiest/safest would be different that what I wrote in the CIP:
1. Yes
2. A governable constant
3. Set in GrandaMento

But I also see the value of having a mint/burn "allowance" so that people have a better impression of what to expect with GrandaMento

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,184 @@
...
+license: Apache 2.0
+---
+
+## Simple Summary
+
+Granda Mento is a mechanism to facilitate large CELO <-> stable token exchanges that aren't suitable via Mento or OTC. A new contract is created that has the authorization by Governance to exchange `CELO <-> stable token`. Exchanges via this contract must be approved by a multisig and can be vetoed by Governance.
+
+## Abstract
+
+There are no existing solutions for making large exchanges ($1m+) involving stable tokens. Apart from cUSD minted for validator payments every epoch, stable tokens can only be minted via Mento. Large volume exchanges via Mento experience high slippage due to the limited bucket sizes in the constant product market maker. Granda Mento is a proposed a mechanism for large-volume stable token exchanges.
```
albertclabs, line 21: Let's characterize "large exchange" ($50k+) and "high slippage" (maybe reframe this a higher slippage of 1-2%)

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Motivation
+
+Users looking to make larger exchanges involving stable tokens are faced with 3 options:
+1. Exchange `CELO <-> stable token` via Mento slowly over a longer period of time.
+2. Arrange one or many `XXX <-> stable token` OTC trades.
+3. Create a large limit order of `fiat <-> stable token` on a centralized exchange in hopes it gets filled over time.
+
+The existing implementation of Mento is suitable for facilitating low volume exchanges, e.g. up to thousands of stable tokens at a time, with minimal slippage. However, significant slippage of over 2% starts to occur on both Mento and on centralized exchanges at trades sizes of ~$50k+. OTC trading satisfies the needs of medium volume exchanges, eg ~100k for cUSD and likely much less for stable tokens with lower total supply (like cEUR), but similarly takes a hit of around 2-3%. There are no existing avenues that are able to satisfy high volume stable token exchanges on the order of millions.
```
albertclabs, line 30: It's worth calling out that options 2 and 3 above may not be possible with smaller stablecoins like cEUR and future coins with lower liquidity close to initialization. Option 3 can be done in theory, but rarely fills due to market makers capturing the spread and has too much time uncertainty to full fill.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,184 @@
...
+#### New contracts:
+
+##### `GrandaMento.sol`
+
+A new contract, `GrandaMento.sol`, is created. One deployment of `GrandaMento` exists per stable token. It must have the ability to mint/burn the corresponding stable token and be a spender of `Reserve.sol`.
+
+The contract is owned Governance, is freezable, and has the following configurable parameters:
+
+1. **`address public approver;`** - Set by governance. Intended to be a multisig and has the authority to approve exchanges.
+2. **`uint256 public minStableExchangeAmount;`** - Set by governance. Minimum amount of the stable token an exchange can buy/sell.
```
albertclabs, line 80: Should we propose what a min/max exchange amount will be when we init this contract? Let's say 200k - 10M?
200k is a bit more than what we've seen people try to (and successfully) convert and make it seem accessible to individuals. 10M is the upper cap of what I've heard *may* we attempted at the very upper end of possibilities. 5M is more realistically what may be done for the future UBI program, so would be a more conservative number.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,184 @@
...
+While Granda Mento will help facilitate `CELO <-> stable token` exchanges, this is ultimately meant as a mechanism to help purchasers with fiat (eg USD) who are seeking to own Celo's corresponding stable token (eg cUSD). Purchasers would come to an agreement with a broker (eg the Celo Foundation) who owns an existing amount of CELO. The broker would be the one to make the `CELO -> cUSD` exchange on-chain.
+
+The proposed implementation is being considered in the medium-term for large scale exchanges until Mento is reworked to support more efficient large-scale minting on a sustained basis.
+
+## Specification
+
+At a high level, the design involves:
+1. Anyone can submit an exchange proposal.
+   * The amount of stable tokens being bought/sold in the exchange is required to be within a governable range.
+   * The assets being sold in the exchange are deposited.
```
albertclabs, line 45: Should we call out CELO and cUSD, cEUR here as the exact assets which are eligible?

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,184 @@
...
+
+#### New contracts:
+
+##### `GrandaMento.sol`
+
+A new contract, `GrandaMento.sol`, is created. One deployment of `GrandaMento` exists per stable token. It must have the ability to mint/burn the corresponding stable token and be a spender of `Reserve.sol`.
+
+The contract is owned Governance, is freezable, and has the following configurable parameters:
+
+1. **`address public approver;`** - Set by governance. Intended to be a multisig and has the authority to approve exchanges.
```
albertclabs, line 79: Would also prefer to see a proposed set of initial approvers. Would this be part of the CGP instead of in the CIP? Perhaps someone from community (Zviad?), someone from cLabs, someone from Foundation. I could also check if the Community Fund signers (Patrick, Dee, Deepak) would do this.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,184 @@
...
+2. **`uint256 public minStableExchangeAmount;`** - Set by governance. Minimum amount of the stable token an exchange can buy/sell.
+3. **`uint256 public maxStableExchangeAmount;`** - Set by governance. Maximum amount of the stable token an exchange can buy/sell.
+4. **`uint256 public exchangeWaitPeriod;`** - Set by governance. The minimum amount of time that must elapse between a proposed exchange being approved and when the exchange can be executed. Should give sufficient time for Governance to veto an approved exchange.
+5. **`uint256 public spread;`** - Set by governance. The percent fee imposed upon an exchange execution.
+
+The contract has the following functions:
+
+1. **`function proposeExchange(uint256 amount, bool sellCelo) external returns (uint256)`** - Called by an exchange proposer to propose an exchange.
```
albertclabs, line 87: "sellCelo = false" would be buying Celo with a stablecoin instead? Needs some clarification to the use of this bool

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,184 @@
...
+       * The amount of the asset being sold.
+   * Returns:
+     * The `id` of the struct in the mapping.
+2. **`function approveExchangeProposal(uint256 id) external`** - Approves a proposed exchange.
+   * Only callable by the approver address.
+   * Marks the proposed exchange with the given `id` as approved.
+   * Records:
+     * The exchange as in the Approved state.
+     * The timestamp at which the approval occurred (ie `block.timestamp`).
+3. **`function cancelExchangeProposal(uint256 id) external`** - Cancels a proposed exchange.
```
albertclabs, line 104: Cancelling or filling a proposal are the only ways for the proposer to get something back. Should we have some kind of time expiry on these proposals (~1-2 weeks)? I'm imagining a worst case where 2/3 of the multisig holders aren't available, on vacation, etc.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,184 @@
...
+4. TWAP between exchange proposal and exchange execution.
+   * Involves smart contract changes to have the TWAP available on chain.
+   * Exchanger doesn't know the price at the proposal time.
+
+It's desirable for both the exchanger and for the Celo community to know what price will be used for the exchange-- this way, the exchanger knows what they're committing to, and the community can decide if they agree with the price. Approaches (1) and (2) the only options that involve knowledge of the price at the start of the trade.
+
+Another question is whether Granda Mento should have an "allowance" that must be granted by Governance indicating how many stable tokens it can mint/burn. Each exchange through Granda Mento would "spend" some of that allowance, and eventually a top-up would be required. This allows reasoning about the worst-case scenario if the approver multisig is compromised and Governance is unable to rally in time. Because this would involve more active involvement from Governance for top-ups and some additional implementation complexity, instead a configurable value in Granda Mento is proposed to restrict how many stable tokens a single exchange can mint/burn.
```
albertclabs, line 158: I think we can leave this "allowance" out. Imo the max exchange amount largely addresses this concern and an allowance is unnecessary.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,184 @@
...
+   * This is vulnerable to oracle attacks.
+   * This puts more control in the hands of the exchange proposer.
+3. Oracle price at the time of exchange execution.
+   * This is vulnerable to oracle attacks.
+   * Exchanger doesn't know the price at the proposal time.
+4. TWAP between exchange proposal and exchange execution.
+   * Involves smart contract changes to have the TWAP available on chain.
+   * Exchanger doesn't know the price at the proposal time.
+
+It's desirable for both the exchanger and for the Celo community to know what price will be used for the exchange-- this way, the exchanger knows what they're committing to, and the community can decide if they agree with the price. Approaches (1) and (2) the only options that involve knowledge of the price at the start of the trade.
```
albertclabs, line 156: I think it's important to highlight here that you're looking for feedback on 1 vs 2 and it's not been decided yet. This may be the most important point of feedback we're looking for in this CIP as the struct created by the proposeExchange function isn't fully defined until this is decided.

CIPs/cip-granda_mento.md
tkporter: Because the CIP is just the specification (that's how I think of it at least) and Governance will have to set the approver multisig on-chain, I think we can leave the decision/proposal of exactly who would be in the multisig to the CGP. But I can specify more generally it would involve community members or members on behalf of large organizations in the Celo ecosystem

CIPs/cip-granda_mento.md
tkporter: If the min amount of multisig signers isn't available, then the proposal hasn't yet been approved by the multisig, which means the proposer is still able to cancel their proposal. Is that sufficient to cover the cases you were thinking of?

If the proposal has been approved, then it will either have to be vetoed by the community or be executed once the delay has passed

CIPs/cip-granda_mento.md
tkporter: Wonder if we should leave this decision to the CGP and leave the CIP as a specification, wdyt?

CIPs/cip-granda_mento.md
tkporter: From a user need perspective, is (2) more desirable than (1)? If users are indifferent then I think it makes the most sense to go with 1.

CIPs/cip-granda_mento.md
tkporter: great idea, added

CIPs/cip-granda_mento.md
albertclabs: (1) is the simplest and most analogous to an OTC price quote at time of request. (2) would require more explanation of how the final price within that range is determined.

CIPs/cip-granda_mento.md
albertclabs: Ah that makes sense. Resolved.

CIPs/cip-granda_mento.md
albertclabs: Sounds good. Let's keep these parameters in mind for the CGP.

CIPs/cip-granda_mento.md
albertclabs: Can leave for CGP.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,183 @@
...
+       * Stable token is minted to the exchange proposer according to the rate originally recorded when the exchange was proposed.
+     * If selling stable token and purchasing CELO:
+       * The stable token is burned.
+       * CELO is transferred from the Reserve to the exchange proposer according to the rate originally recorded when the exchange was proposed.
+   * Records:
+     * The exchange as in the Executed state.
+
+## Rationale
+
+Granda Mento is not intended to undermine the market or provide a way for large players to "dump" CELO for cUSD. It's also not intended to be an exceptionally elegant or complicated solution-- instead, this is looking to be something that is simple, straightforward, and works alongside the existing implementation of Mento. Any larger-scale Mento design modifications to facilitate large stable token mints are out of scope, and should be considered for a medium/long term design change to the stability protocol. The proposed implementation attempts to navigate the tradeoffs of complexity, centralization, economic safety, and ease of use for all actors in the Celo ecosystem.
```
tobikuhlmann, line 129: Do we want to create Granda Mento for both cUSD and cEUR? If yes, a generalization from cUSD to e.g. cXXX might be clarifying and increase consistency (As already used elsewhere here).
Another question that comes to my mind: What will happen for new stable tokens? Will they directly get a Granda Mento? @albertclabs

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,183 @@
...
+Implementation Risks:
+* Modifications to logic in `StableToken.sol` around which addresses can mint/burn.
+* `GrandaMento.sol`'s logic around minting/burning stable tokens or spending Reserve funds must be accurate.
+* Exchanges in `GrandaMento.sol` must only be executable if they meet the required criteria.
+
+Operational Risks:
+* The multisig may become compromised or approve exchanges that are not favorable from the community's perspective. Governance must be on the lookout for exchange proposals to veto.
+
+Economic Risks:
+* The stability protocol has never dealt with large mints/burns apart from the initial allocations of cUSD and cEUR. An extremely large mint/burn quantity could potentially affect stability.
```
tobikuhlmann, line 180: That's an interesting risk. Also if someone's plans with a large amount of cXXX minted through Granda Mento change and they dump them on the market (improbable though? And that could come from any large cXXX holder). Mento might need a while to contract or expand that much supply. Probably a larger Mento discussion

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,183 @@
...
+category (*only required for Standards Track): <Ring 0, 1, 2, 3>
+created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
+requires (*optional): <CIP number(s)>
+replaces (*optional): <CIP number(s)>
+license: Apache 2.0
+---
+
+## Simple Summary
+
+Granda Mento is a mechanism to facilitate large CELO <-> stable token (e.g. cXXX) exchanges that aren't suitable via Mento or OTC. A new contract is created that has the authorization by Governance to exchange `CELO <-> stable token`. Exchanges via this contract must be approved by a multisig and can be vetoed by Governance.
```
tobikuhlmann, line 17: Albert mentioned Granda Mento would be for cUSD and cEUR for now, and not automatically for new stable coins. I think defining that here and using it consistently would be clarifying

tkporter: That's a great point. I was imagining that Granda Mento would be deployed for cUSD & cEUR now, but also for future stable tokens. @albertclabs does that align with what you were thinking?

CIPs/cip-granda_mento.md
tkporter: If there's no strong user need for (2) then I'm going to commit to (1)

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,186 @@
...
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): <Ring 0, 1, 2, 3>
+created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
+license: Apache 2.0
+---
+
+## Simple Summary
+
+Granda Mento is a mechanism to facilitate large CELO <-> stable token (e.g. cXXX) exchanges that aren't suitable via Mento or OTC. A new contract is created authorization by Governance to exchange `CELO <-> stable token` for any stable token Governance explicitly enables. Exchanges via this contract must be approved by a multisig and can be vetoed by Governance.
```
martinvol, line 15: > is created authorization by Governance to exchange

What do you mean here?

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,186 @@
...
+
+While Granda Mento will help facilitate `CELO <-> stable token` exchanges, this is ultimately meant as a mechanism to help purchasers with fiat (e.g. USD) who are seeking to own Celo's corresponding stable token (e.g. cUSD). Purchasers would come to an agreement with a broker (e.g. the Celo Foundation) who owns an existing amount of CELO. The broker would be the one to make the `CELO -> cUSD` exchange on-chain.
+
+The proposed implementation is being considered in the medium-term for large scale exchanges until Mento is possibly reworked to support more efficient large-scale minting on a sustained basis.
+
+## Specification
+
+At a high level, the design involves:
+1. Anyone can submit an exchange proposal.
+   * The amount of stable tokens being bought/sold in the exchange is required to be within a governable range.
```
martinvol, line 42: Interested about what should we consider for an upper hard cap. I sthis meant for security or there is some other reason we would like a limit?

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,186 @@
...
+   * The assets being sold (ie either the stable token or CELO) in the exchange are deposited.
+   * The current oracle price for the exchange is recorded.
```
martinvol, line 44: Although this is significantly more complicated to implement on-chain, would it be more accurate to take a multi-day price average? This is often how large trades are priced OTC.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,186 @@
...
+2. The proposed exchange must be approved by a multisig that has previously been authorized by Governance. At this point, the exchange still cannot be executed.
```
martinvol, line 45: It'd be interesting if governance itself could be signer of this multisig, in case of an emergency.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,186 @@
...
+3. A forced waiting period of X days must elapse before the exchange can be executed. During this time, Governance can choose to veto the exchange, refunding the exchange proposer.
```
martinvol, line 46: Is governance fast enough for this, considering that a proposal takes more than a week? That means that the proposal can has a couple of days to start the veto if the whole exchange process takes 2 days. 

Maybe at some point we could implement an ad-hoc governance process that finishes as soon as it has more nay votes than CELO to be traded.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,186 @@
...
+4. After the waiting period, the exchange is executed according to the oracle price recorded in step 1.
+
+An exchange can have the following states:
+1. Proposed - The exchange has been proposed, but not yet approved by the approver. The proposer may still cancel their proposal and be refunded their deposit.
+2. Approved - The exchange has been proposed and approved by the approver, but not yet executed or vetoed. The proposer may not cancel their proposal.
```
martinvol, line 51:
```suggestion
2. Approved - The exchange has been proposed and approved by the approver, but not yet executed or vetoed. The proposer may no longer cancel their proposal at this point.
```

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Modifications of existing contracts:
+
+##### `StableToken.sol`
+
+Currently, `StableToken.sol` only allows its `Exchange.sol` or `Validators.sol` to [mint](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L225) new stable tokens. Similarly, only its `Exchange.sol` can [burn](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L275) stable tokens.
+
+The following modifications will be made:
+1. **Modification to [mint()](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L224)**
+   * Allow `msg.sender` to be the stable token's `GrandaMento` in addition to the existing permitted senders, Exchange and `Validators.sol`.
```
martinvol, line 65: interesthing if this should be enabled for al StableTokens or we would like to enable it ad-hoc for cEUR or  other exchanges later on.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,186 @@
...
+2. **Modification to [burn()](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L273)**
+   * Allow `msg.sender` to be the stable token's `GrandaMento` in addition to the existing permitted sender, the Exchange.
+
+#### New contracts:
+
+##### `GrandaMento.sol`
```
martinvol, line 71: So far the naming of our contracts have been straightforward, I'd leave GrandaMento as a codename of this effort but I'd use something like "BigExchange.sol" for it.

CIPs/cip-granda_mento.md
martinvol: I feel like StableTokenX should have a governable list of contracts allowed to burn/mint.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,186 @@
...
+       * The amount of the asset being sold.
+       * Whether CELO is being sold.
+   * Returns:
+     * The `id` of the struct in the mapping.
+2. **`function approveExchangeProposal(uint256 id) external`** - Approves a proposed exchange.
+   * Only callable by the approver address.
+   * Marks the proposed exchange with the given `id` as approved.
+   * Records:
+     * The exchange as in the Approved state.
+     * The timestamp at which the approval occurred (ie `block.timestamp`).
```
martinvol, line 102: Maybe just blockNumber so it's easier to look up.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,186 @@
...
+## Specification
+
+At a high level, the design involves:
+1. Anyone can submit an exchange proposal.
+   * The amount of stable tokens being bought/sold in the exchange is required to be within a governable range.
+   * The assets being sold (ie either the stable token or CELO) in the exchange are deposited.
+   * The current oracle price for the exchange is recorded.
+2. The proposed exchange must be approved by a multisig that has previously been authorized by Governance. At this point, the exchange still cannot be executed.
+3. A forced waiting period of X days must elapse before the exchange can be executed. During this time, Governance can choose to veto the exchange, refunding the exchange proposer.
+4. After the waiting period, the exchange is executed according to the oracle price recorded in step 1.
```
martinvol, line 47: Shall it include a fee? I see no reason why this wouldn't include a fee of 0.5%, because it's not paying any spread and indeed bigger trades would be more favorable than small ones.

There could be an argument that the fee should be lower than 0.5%, since the exchanger is bearing risk in case of a big price move against their favor, but the whole exchange can be vetoed in case the price moves on their favor.

CIPs/cip-granda_mento.md
tkporter: I think it has a number of benefits, security like you mentioned, but also as a general safety check for some protection against fat-fingers adding an extra zero

CIPs/cip-granda_mento.md
tkporter: typo oops, thanks

CIPs/cip-granda_mento.md
tkporter: This was discussed, ultimately because of the human "counterparties" that make up the approver multisig & governance, it was deemed that the approver/governance could disagree with the oracle price if they want, and that it's not worth the implementation effort & risk of implementing an on-chain TWAP. I feel like it could also be argued that just a straight up TWAP isn't a totally "fair" price, eg if the CELO price just jumped (which it has done in the past) then the TWAP may be low and the exchanger is trying their best to get a better deal than they would've if the normal oracle price was used
I'll beef up the `Rationale` section with this information, wdyt of the decision?

CIPs/cip-granda_mento.md
tkporter: I think a likely waiting period will be ~2 weeks, which gives a week to prep the governance veto and a week to vote on it. The user need is less about needing the funds quickly but more knowing the price they'll be given. But I agree timelines are tight-- I think this would be a good candidate for the "governance 2.0" idea of having some proposals be fast-tracked

CIPs/cip-granda_mento.md
tkporter: That's the purpose of `FixidityLib.Fraction public spread;` further down -- I'll make this clearer here

CIPs/cip-granda_mento.md
tkporter: This line used to make more sense when I had a 1 GrandaMento per StableToken, but I recently moved it to 1 GrandaMento for all StableTokens (fixing now). The current proposal would be to have all StableTokens allow GrandaMento to mint/burn, but the only way for GrandaMento to actually mint/burn is if governance explicitly whitelists a specific StableToken inside GrandaMento

Re: governable list of contracts, I'm a little torn-- I see the value in there being a governable list of contracts, but at the same time I worry adding this feature could up the complexity/surface area of the proposed change. @yorhodes thoughts?

CIPs/cip-granda_mento.md
tkporter: @yorhodes is there a preference from CAP on contract naming? I'd very slightly prefer to keep granda mento to avoid confusion, but I'm ok either way

CIPs/cip-granda_mento.md
tkporter: Not sure I follow why blockNumber is more advantageous?

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+   * Deposits the full amount of the asset being sold into the contract.
+   * Records:
+     * The exchange as in the Proposed state.
+     * A struct with the following info is stored in a mapping with an `id` key:
+       * The stable token address.
+       * The amount of the asset being sold.
+         * Behind the scenes, StableToken keeps track of balances via "units," which do not change with time. However, the value returned by a call to `balanceOf` is the "value," which can change with time due to the inflation/demurrage feature. Because GrandaMento must refund a deposit for a Cancelled exchange proposal, GrandaMento should be sure to refund the correct amount adjusted by any inflation that has been incurred. To account for this, the amount of the StableToken being sold is stored in "units," and the "value" at a given time is calculated using [StableToken.unitsToValue](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L415).
```
tkporter, line 94: @yorhodes What do you think of us changing this to `public`, or adding a getter or something?

https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L49

That way instead of doing:
```
units = stableToken.valueToUnits(
  stableToken.balanceOf(...) // this has a `unitsToValue` call in there
)
```
we could just do w/ less gas
```
units = stableToken.balanceOfUnits(...)
```

CIPs/cip-granda_mento.md
martinvol: Oh, missed that

CIPs/cip-granda_mento.md
martinvol: From the implementation perspective doesn't seem that complex to make contracts able to allow a particular contract to burn/mint an it can be used in future use cases. I'm not saying we should do it that way, not 100% sold myself, but I think that the decision should be around better UX rather than code simplicity.

CIPs/cip-granda_mento.md
martinvol: you can just paste it on the explorer and it's easier to get the approve tx and tx that happened in the same block that could be relevant. This is definitely a nit though.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+
+The contract has the following configurable parameters:
+
+1. **`address public approver;`** - Set by Governance. Intended to be a multisig and has the authority to approve exchanges. The signers for the multisig are at the discretion of Governance, but it would comprise of community members or members on behalf of organizations that are aligned with the Celo network/ecosystem.
+2. **`uint256 public exchangeWaitPeriodSeconds;`** - Set by Governance. The minimum amount of time in seconds that must elapse between a proposed exchange being approved and when the exchange can be executed. Should give sufficient time for Governance to veto an approved exchange.
+3. **`FixidityLib.Fraction public spread;`** - Set by Governance. The percent fee imposed upon an exchange execution.
```
martinvol, line 79: Do we have any consensus about how much that fee should be?

tobikuhlmann: I think we didn't talk much about this yet (maybe @tkporter and @albertclabs have more details meanwhile). @albertclabs and I had a brief chat about it and he mentioned users would probably not care about fees like 2-3 % what they would get OTC, which would be a neat way to bolster the reserve. My thoughts were at least the costs (fees + slippage) that they would have with Mento when they would execute slowly in batches. What do you think?

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+Building a process to exchange large (~$1 million+) amounts of CELO for stable tokens will enable large entities to make significant block purchases of stable tokens at scale. While minting of the stable token is expected to be the main use case, reversibility (ie exchanging `stable token -> CELO`) is desirable to instill higher confidence when exchanging large quantities.
+
+While Granda Mento will help facilitate `CELO <-> stable token` exchanges, this is ultimately meant as a mechanism to help purchasers with fiat (e.g. USD) who are seeking to own Celo's corresponding stable token (e.g. cUSD). Purchasers would come to an agreement with a broker (e.g. the Celo Foundation) who owns an existing amount of CELO. The broker would be the one to make the `CELO -> cUSD` exchange on-chain.
+
+The proposed implementation is being considered in the medium-term for large scale exchanges until Mento is possibly reworked to support more efficient large-scale minting on a sustained basis.
+
+## Specification
+
+At a high level, the design involves:
+1. Anyone can submit an exchange proposal.
```
zviadm, line 41: It would be nice if every proposal was transparent about where the CELO is coming from and at what price it was acquired at.
Examples:
* CELO acquired through Celo Foundation at price of X$ (then it would make sense to use no more than that price when converting CELO -> cUSD)
* CELO acquired through Market order (this is going to be more contentious. Because 1M$ CELO order can move market price by 10%, so it wouldn't make sense to allow selling CELO->cUSD at current 10% increased market price).
* CELO acquired through OTC trade, similar case as acquiring directly from Celo Foundation

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Security Considerations
+
+Implementation Risks:
+* Modifications to logic in `StableToken.sol` around which addresses can mint/burn.
+* `GrandaMento.sol`'s logic around minting/burning stable tokens or spending Reserve funds must be accurate.
+* Exchanges in `GrandaMento.sol` must only be executable if they meet the required criteria.
+
+Operational Risks:
+* The multisig may become compromised or approve exchanges that are not favorable from the community's perspective. Governance must be on the lookout for exchange proposals to veto.
```
asaj, line 181: A twitter bot could be interesting like the one running for governance proposals

asaj (commented):
Very nicely written! Would love to have a discussion on the two approaches for setting a price but after that LGTM.

CIPs/cip-granda_mento.md
martinvol: It really makes sense, as the protocol can reject if the pricing goes really bad. Regarding the TWAP being "too good" in case of a price jump, that is really not a problem due liquidity constraints, for example: 
* 90 days TWAP is $3
* CELO is currently trading at $3.5
* Someone is making a proposal to buy 1M celo @TWAP = 3 cUSD

Although that scenario seems unfair, in reality the purchaser would in fact struggle to get a price of $3 if it wanted to liquidate that whole position on a CEX. I wouldn't personally push to veto a trade of that kind.

Maybe something we could do is allow the exchanger to submit their own price along with the exchange proposal, as long as it is within a certain range from the oracle price?

Without this "price discovery" mechanism, it could be that parties are reluctant to use Granda Mento in times of high volatility, or try to time the submission to get a good price, or even manipulate the price.

CIPs/cip-granda_mento.md
tkporter: ahhh this isn't intended to be for looking up the approval tx, this records the timestamp so there can be a check when executing that the appropriate amount of time has elapsed since the exchange proposal was approved (so that Governance has time to veto).
This check here:
https://github.com/celo-org/celo-proposals/pull/224/files#diff-fdeefbb76f60238f3814cb1238cb13ba0914ae709ec9856f368d49399900fa00R115

CIPs/cip-granda_mento.md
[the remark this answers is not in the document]
tkporter: that's a good idea, though it would definitely consume more gas. But given the expected frequency of proposals this is probably not an issue.

I feel like GrandaMento won't be the last contract to have to handle this, so there might be value in serving as an example that's sort of gas efficient

The only other ideas I have are:
* The value of deposits is kept track of as a percentage of the balance of `GrandaMento.sol`- this feels too complex/confusing
* Move the logic into a new wrapper token of StableToken that only keeps track of units, ie its `balanceOf` returns the amount in units rather than the value. That way balances of the wrapper are unchanging and it abides by the ERC20 interface. Also feels a little complex/confusing

CIPs/cip-granda_mento.md
albertclabs: I think a TWAP method would discourage anyone from using it during long periods of time when the TWAP is lower than the market price. The human gatekeepers should ensure that the "market price" at time of request isn't too far from something that is unreasonable (such as a random market fluctuation to the downside). I'd keep this implementation simple for now, as it will likely not be used too frequently.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+## Specification
+
+At a high level, the design involves:
+1. Anyone can submit an exchange proposal.
+   * The amount of stable tokens being bought/sold in the exchange is required to be within a governable range.
+   * The assets being sold (ie either the stable token or CELO) in the exchange are deposited.
+   * The current oracle price for the exchange is recorded.
+2. The proposed exchange must be approved by a multisig that has previously been authorized by Governance. At this point, the exchange still cannot be executed.
+3. A forced waiting period of X days must elapse before the exchange can be executed. During this time, Governance can choose to veto the exchange, refunding the exchange proposer.
+4. After the waiting period, the exchange is executed according to the oracle price recorded in step 1. A fee is imposed upon the exchange through the use of a "spread".
```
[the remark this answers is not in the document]
albertclabs: This is an extra need for the proposer to set parameters correctl, leaves more room for human error, and would be another avenue for trying to "game the system" for slightly less slippage. I'm in favor of the simplest implementation here as proposed and applying a spread (equal to current 0.5% spread in Mento?)

CIPs/cip-granda_mento.md
albertclabs: My feeling here is that 0.5% similar to Mento today would be appropriate. It could likely be higher up to 1% and still be similar to what they would experience writing a block purchasing script. I'd shy away from much more than that as the primary aim of this is not the bolster the Reserve, but to increase liquidity in the greater ecosystem.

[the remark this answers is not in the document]
albertclabs: I would lean towards giving as little control to the proposer as possible. More parameters add more complexity and could be a barrier to entry.

Some of the comparable solutions for large buyers also have very few input variables...
OTC trade: Similar intended conversion amount as input, variable conversion rate depending on OTC desk
Large limit order on CEX: variable limit price and quantity, uncertain (long or forever) execution time
Granda Mento: Should limit user controlled factors to just the total conversion amount. Conversion certainty in time is a benefit for the user over CEXes, conversion fee/rate certainty closer to proposal time is a benefit over OTC trades.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+   * Only callable by Governance.
+   * Sets the minimum and maximum amounts of stable token that can be minted/burned in an exchange by updating `stableTokenExchangeLimits[stableToken]`.
+     * Effectively "enables" exchanging a stable token if the min & max amounts were previously 0.
+     * Effectively "disables" exchanging a stable token if setting the min & max amounts to 0.
+
+## Rationale
+
+Granda Mento is not intended to undermine the market or provide a way for large players to "dump" CELO for cXXX. It's also not intended to be an exceptionally elegant or complicated solution-- instead, this is looking to be something that is simple, straightforward, and works alongside the existing implementation of Mento. Any larger-scale Mento design modifications to facilitate large stable token mints are out of scope, and should be considered for a medium/long term design change to the stability protocol. The proposed implementation attempts to navigate the tradeoffs of complexity, centralization, economic safety, and ease of use for all actors in the Celo ecosystem.
```
albertclabs, line 133: I might actually call out here that it should be intended to be much simpler or understandable than some of the comparable solutions for users. This would summarize of the comparative analysis and justify a few of the design decisions above.

vs OTC: Fees are fixed, constant, and lower
vs. Exchange Limit Orders: Fill or cancellation time is a predictible
vs. Trading bot: Fixed slippage, single transaction execution, less technical knowledge needed

CIPs/cip-granda_mento.md
martinvol: Seems like 0.75% would be a good number to present to the community?

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+3. **`FixidityLib.Fraction public spread;`** - Set by Governance. The percent fee imposed upon an exchange execution.
+4. **`mapping(address => ExchangeLimits) stableTokenExchangeLimits`** - Set by Governance via `setStableTokenExchangeLimits`. A mapping of stable token address to a struct containing two `uint256`s, `minExchangeAmount` and `maxExchangeAmount`.
+
+The contract has the following functions:
+
+1. **`function proposeExchange(address stableToken, uint256 sellAmount, bool sellCelo) external returns (uint256)`** - Called by an exchange proposer to propose an exchange.
+   * Callable by anyone.
+   * If `sellCelo` is true, CELO is the asset being sold and `stableToken` is the asset being bought. If `sellCelo` is false, `stableToken` is the asset being sold and CELO is the asset being bought.
+   * Requires the amount of `stableToken` being bought/sold to be within the token-specific range `[stableTokenExchangeLimits[stableToken].minExchangeAmount, stableTokenExchangeLimits[stableToken].maxExchangeAmount]` that is set by Governance via `setStableTokenExchangeLimits` (described later).
```
yorhodes, line 87: Why are parameters are a governance concern but exchange approvals an elected multisig concern?

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+2. **`uint256 public exchangeWaitPeriodSeconds;`** - Set by Governance. The minimum amount of time in seconds that must elapse between a proposed exchange being approved and when the exchange can be executed. Should give sufficient time for Governance to veto an approved exchange.
+3. **`FixidityLib.Fraction public spread;`** - Set by Governance. The percent fee imposed upon an exchange execution.
+4. **`mapping(address => ExchangeLimits) stableTokenExchangeLimits`** - Set by Governance via `setStableTokenExchangeLimits`. A mapping of stable token address to a struct containing two `uint256`s, `minExchangeAmount` and `maxExchangeAmount`.
+
+The contract has the following functions:
+
+1. **`function proposeExchange(address stableToken, uint256 sellAmount, bool sellCelo) external returns (uint256)`** - Called by an exchange proposer to propose an exchange.
+   * Callable by anyone.
+   * If `sellCelo` is true, CELO is the asset being sold and `stableToken` is the asset being bought. If `sellCelo` is false, `stableToken` is the asset being sold and CELO is the asset being bought.
```
yorhodes, line 86: I know we use this `sellCelo` switch elsewhere but I have always found this confusing
Can we just expose 2 endpoints? Can you remind me of the rationale here.

CIPs/cip-granda_mento.md
yorhodes: I don't really understand this limitation
Could we make the `sellAmount` denominated in units?

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+     * If selling stable token and purchasing CELO:
+       * The stable token is burned.
+       * CELO is transferred from the Reserve to the exchange proposer according to the rate originally recorded when the exchange was proposed.
+   * Records:
+     * The exchange as in the Executed state.
+5. **`function setStableTokenExchangeLimits(address stableToken, uint256 minExchangeAmount, uint256 maxExchangeAmount) external`**
+   * Only callable by Governance.
+   * Sets the minimum and maximum amounts of stable token that can be minted/burned in an exchange by updating `stableTokenExchangeLimits[stableToken]`.
+     * Effectively "enables" exchanging a stable token if the min & max amounts were previously 0.
+     * Effectively "disables" exchanging a stable token if setting the min & max amounts to 0.
```
yorhodes, line 129: do these need to be denominated in `units` for different exchange directions?

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+
+Some ideas other than the proposed approach include:
+
+1. Exchanges have a delay and require a deposit of the sold asset. Any exchange must be explicitly approved by a governance proposal to be executed.
+   * This relies upon the community as the arbiter of what exchanges are "safe."
+   * This could exacerbate the already-present voter fatigue.
+2. Large-volume exchanges could be made by anyone without Governance approval.
+   * The most compelling approach to this would involve an auction. This is complicated to implement, requires careful design, and was intentionally removed from the original Mento design.
+
+Approach (2) was deemed risky and complicated. Approach (1) felt more in line with the goal of having Granda Mento be simple and safe, but ultimately the involvement of Governance for every exchange felt unsustainable and exhausting for voters. The proposed implementation involving an exchange delay, a multisig for approvals, and Governance in the (likely rare) event the community wishes to veto an exchange, aims to relieve voter exhaustion while still providing the security of community scrutiny.
```
yorhodes, line 143: This requires operational monitoring of the `GrandaMento` exchanges for governance participants
Could this lead to even more fatigue?

yorhodes (approved):

YazzyYaz (changes requested):

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): <Ring 0, 1, 2, 3>
+created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
+license: Apache 2.0
+---
+
+## Simple Summary
+
+Granda Mento is a mechanism to facilitate large CELO <-> stable token (e.g. cXXX) exchanges that aren't suitable via Mento or OTC. A new contract is created that can exchange `CELO <-> stable token` for any stable token Governance explicitly enables. Exchanges via this contract must be approved by a multisig and can be vetoed by Governance.
```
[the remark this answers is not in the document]
tkporter: Like you mentioned earlier, the multisig approver was intended to alleviate fatigue for voters. Governance will be able to set the `approver` address. There was some pushback against having Governance being required to approve every exchange. There were concerns that it would require too much participation if we expected Granda Mento to exist for more than a few months / exchanges. As it became more clear that this is something that'll exist for a decent amount of time (especially for smaller total supply stablecoins like cEUR or those that may be added in the future), the multisig approach became more attractive. I believe as long as the process is transparent, community members can give their input, and governance has the ability to step in if the multisig doesn't execute the will of the community, there are very similar benefits to exchanges being approved by governance

Fwiw there's a (very small) precedent of governance voting for vetoes I read about [here](https://ournetwork.substack.com/p/our-network-deep-dive-2)

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,187 @@
...
+
+##### `GrandaMento.sol`
+
+A new contract, `GrandaMento.sol`, is created, added to the Registry with the identifier `GrandaMento`, and owned by Governance. `GrandaMento` is a singleton contract that is capable of facilitating exchanges for multiple stable tokens-- that is, with the approval of Governance, `GrandaMento` could support `CELO <-> cUSD`, `CELO <-> cEUR`, and future `CELO <-> cXXX` exchanges. `GrandaMento` must be a spender of `Reserve.sol`, and supported stable tokens must allow `GrandaMento` to mint/burn them.
+
+The contract has the following configurable parameters:
+
+1. **`address public approver;`** - Set by Governance. Intended to be a multisig and has the authority to approve exchanges. The signers for the multisig are at the discretion of Governance, but it would comprise of community members or members on behalf of organizations that are aligned with the Celo network/ecosystem.
+2. **`uint256 public exchangeWaitPeriodSeconds;`** - Set by Governance. The minimum amount of time in seconds that must elapse between a proposed exchange being approved and when the exchange can be executed. Should give sufficient time for Governance to veto an approved exchange.
```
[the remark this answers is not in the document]
tkporter: This would be cool, it'd definitely avoid issues if the governance parameters are made longer.

Something I was thinking would be cool is if governance had the ability to assume the role of the approver if it desires. It could set the `approver` to the GovernanceProxy address and then set `exchangeWaitPeriodSeconds` to zero (to avoid an exchange requiring basically 2 periods of a governance vote to be executed). Do you think that's compelling enough a reason to not go with the sum of governance parameters?

CIPs/cip-granda_mento.md
tkporter: I wasn't involved with the rationale in `Exchange.sol` but I feel like the 2 endpoint approach would end up looking like (excuse the bad fn names):
```
function proposeExchange(address stableToken, uint256 sellAmount, bool sellCelo) internal returns (uint256) {
  // ...
}

function proposeCeloSell(address stableToken, uint256 sellAmount) external returns (uint256) {
  return proposeExchange(stableToken, sellAmount, true);
}

function proposeStableTokenSell(address stableToken, uint256 sellAmount) external returns (uint256) {
  return proposeExchange(stableToken, sellAmount, false);
}
```

And imo there isn't much to gain from having those functions on the contract level. Though I usually think of contracts as something that's ok to have a slightly rough interface because they're usually going to be called by another piece of code (like contractkit or the cli) that can abstract away the rough interface and give a better UX. Does CAP think of contracts differently?

CIPs/cip-granda_mento.md
tkporter: Because the inflation feature of StableToken doesn't depeg 1 cUSD value from being pegged to 1 USD, it just causes the balance of someone holding cUSD to go down as time goes on, I think these should be denominated in value.

CIPs/cip-granda_mento.md
tkporter: Hmmm interesting point. In my opinion the fatigue around governance is less about being active in the community (eg discussing proposals or keeping up to date) but more around bigger players having to go through key management processes to use their voting keys frequently. So while there will be some responsibility on Governance now to keep an eye on exchange proposal approvals, I don't think it will dramatically affect fatigue

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,188 @@
...
+   * This puts more control in the hands of the exchange proposer.
+3. Oracle price at the time of exchange execution.
+   * This is vulnerable to oracle attacks.
+   * Exchanger doesn't know the price at the proposal time.
+4. TWAP between exchange proposal and exchange execution.
+   * Involves smart contract changes to have the TWAP available on chain.
+   * Exchanger doesn't know the price at the proposal time.
+
+It's desirable for both the exchanger and for the Celo community to know what price will be used for the exchange-- this way, the exchanger knows what they're committing to, and the community can decide if they agree with the price. Approaches (1) and (2) the only options that involve knowledge of the price at the start of the trade. While these are both vulnerable to oracle attacks, the proposed implementation's approver and Governance veto serve as safeguards against an exchange with a manipulated price being executed. Because there is not a strong user need behind (2), (1) has been chosen. While the use of a TWAP for the price may be nice, the implementation complexity of implementing a TWAP on-chain is high, and the benefit is slim.
```
nvtaveras, line 162:
```suggestion
It's desirable for both the exchanger and for the Celo community to know what price will be used for the exchange-- this way, the exchanger knows what they're committing to, and the community can decide if they agree with the price. Approaches (1) and (2) are the only options that involve knowledge of the price at the start of the trade. While these are both vulnerable to oracle attacks, the proposed implementation's approver and Governance veto serve as safeguards against an exchange with a manipulated price being executed. Because there is not a strong user need behind (2), (1) has been chosen. While the use of a TWAP for the price may be nice, the implementation complexity of implementing a TWAP on-chain is high, and the benefit is slim.
```

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,188 @@
...
+The contract has the following functions:
+
+1. **`function proposeExchange(address stableToken, uint256 sellAmount, bool sellCelo) external returns (uint256)`** - Called by an exchange proposer to propose an exchange.
+   * Callable by anyone.
+   * If `sellCelo` is true, CELO is the asset being sold and `stableToken` is the asset being bought. If `sellCelo` is false, `stableToken` is the asset being sold and CELO is the asset being bought.
+   * Requires the amount of `stableToken` being bought/sold to be within the token-specific range `[stableTokenExchangeLimits[stableToken].minExchangeAmount, stableTokenExchangeLimits[stableToken].maxExchangeAmount]` that is set by Governance via `setStableTokenExchangeLimits` (described later).
+   * Deposits the full amount of the asset being sold into the contract.
+   * Records:
+     * The exchange as in the Proposed state.
+     * A struct with the following info is stored in a mapping with an `id` key:
```
nvtaveras, line 92: Can we expand a bit here on how will the oracle rate be recorded? Will it also be part of this struct?

CIPs/cip-granda_mento.md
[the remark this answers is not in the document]
nvtaveras: Since the overall Granda Mento process will be gated by the community, are there any scenarios where two entities would want to propose a trade at a different spread after going through some sort of offline community discussion/pre-approval? e.g. an NGO vs a larger corporation looking to acquire cXXX.

Just thinking a bit here if we are sure we want to impose the same spread to all trades.

cc: @tkporter

CIPs/cip-granda_mento.md
yorhodes: For the core contracts, usage by another contract is probably the most pertinent. Much of this proposal seems motivated by mitigating misuse of the exchange proposals. User understanding of verified interfaces is one of the last lines of defense against malicious application signatures, so I generally prefer a more specified interface where possible. For example, the user probably knows nothing about the address of `address stableToken`, but could more reasonably sign a `proposeCELOToCUSDExchange(amount)`.

yorhodes: I have advocated for registry ids in favor of addresses in the past and will continue to do so for this reason, among others.

tobikuhlmann (approved):
LGTM! Just Nelson's typo fixes should be committed

CIPs/cip-granda_mento.md
albertclabs: It looks like most of the discussion has been around the amount of the fee (as something desired to be greater than zero), and not the inclusion of this parameter. We can discuss going into the CGP what value to set it to, as it looks like there's broad agreement this should be a governable value.

albertclabs (approved):

martinvol (approved):
Amazing work @tkporter!

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,185 @@
+---
+cip: <to be assigned>
+title: Granda Mento
+author: Trevor Porter (@tkporter), Albert Wang (@albertclabs)
+discussions-to: https://forum.celo.org/t/discussion-on-granda-mento-enabling-larger-stablecoin-mints/966
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): <Ring 0, 1, 2, 3>
```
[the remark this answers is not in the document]
asaj: It is touching the core contracts as it allows a new way for cXXX to be minted or burned, I think by the definition in CIP-0000 it would be ring 1.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,188 @@
...
+Granda Mento is a mechanism to facilitate large CELO <-> stable token (e.g. cXXX) exchanges that aren't suitable via Mento or OTC. A new contract is created that can exchange `CELO <-> stable token` for any stable token Governance explicitly enables. Exchanges via this contract must be approved by a multisig and can be vetoed by Governance.
+
+## Abstract
+
+There are no existing solutions for making large exchanges ($1m+) involving stable tokens. Apart from cUSD minted for validator payments every epoch, stable tokens can only be minted via Mento. Large volume exchanges ($50k+) via Mento experience higher slippage (1-2%) due to the limited sizes of the constant product market maker buckets. Granda Mento is a proposed a mechanism for large volume stable token exchanges, primarily to enable minting of large quantities of stable tokens.
+
+## Motivation
+
+There are no existing avenues that are able to satisfy high volume stable token exchanges on the order of millions. Users looking to make larger exchanges involving stable tokens are faced with 3 options:
+1. Exchange `CELO <-> stable token` via Mento slowly over a longer period of time.
```
asaj, line 24:
```suggestion
1. Exchange `CELO <-> stable token` via Mento slowly over a longer period of time to minimize slippage.
```

asaj (approved):

CIPs/cip-granda_mento.md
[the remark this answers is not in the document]
tkporter: Hmm that's an interesting point, my feeling is that having the same spread is desirable from the community's perspective and that any large exchanger would be okay with having the same flat spread as that's sort of common in OTC. I think this is something that @albertclabs can probably attest to?

CIPs/cip-granda_mento.md
tkporter: For posterity, ended up using the registry ID of the stable token as a parameter. Also moved to using the registry ID as the key for the exchange limits of a stable token.

CIPs/cip-granda_mento.md
```diff
@@ -0,0 +1,188 @@
...
+1. Anyone can submit an exchange proposal.
+   * The amount of stable tokens being bought/sold in the exchange is required to be within a governable range.
+   * The assets being sold (ie either the stable token or CELO) in the exchange are deposited.
+   * The current oracle price for the exchange is recorded.
+   * Off-chain, each proposal should be shared with the community in a forum post. For transparency, the post should share where the token being sold is coming from, when it was acquired, and at what price it was acquired.
+2. The proposed exchange must be approved by a multisig that has previously been authorized by Governance. At this point, the exchange still cannot be executed.
+3. A forced waiting period of X days must elapse before the exchange can be executed. During this time, Governance can choose to veto the exchange, refunding the exchange proposer.
+4. After the waiting period, the exchange is executed according to the oracle price recorded in step 1. A fee is imposed upon the exchange through the use of a "spread".
+
+An exchange can have the following states:
```
YazzyYaz, line 50: This feels like 2 proposals, one for enabling Granda Mento and another for defining the process around submitting an exchange proposal for the Granda Mento. Also, who will be coordinating those proposals?

CIPs/cip-granda_mento.md
YazzyYaz: I think it'll be the opposite, more proposals to keep track off means coordination and participation from others, even if it's not necessarily be the same as the regular governance process.

CIPs/cip-granda_mento.md
YazzyYaz: @tkporter what is "Governance must be on the lookout" in this case? Who comprises the Governance team that needs to be be aware of another class of proposals that will need to be tracked?

CIPs/cip-granda_mento.md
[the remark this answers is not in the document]
tkporter: 100% agree. The multisig owners haven't been decided yet, but I think that can be a convo that's had when the CGP is proposed that will activate GrandaMento & actually set the multisig.

CIPs/cip-granda_mento.md
tkporter: I like that - do you think that template should live in this CIP or elsewhere?

CIPs/cip-granda_mento.md
tkporter: Around the next contract release when the implementation is deployed, cLabs (probably me) will propose the CGP to activate Granda Mento for cUSD and cEUR, which will probably be a single one-time proposal. As subsequent stabletokens are released, I imagine activating Granda Mento for the new stable token will be included in the process of launching the stable token

As for the process for submitting an exchange proposal, currently I feel like the CIP doesn't give much info on that, but I would expect the exchange proposer would be responsible for some process like a forum post, possibly chatting about it in a governance call, etc.

Issue #233, opened 2021-06-15
[WIP] - Discussions for extension to ODIS for password hashing


nategraf:
[CIP current WIP]

YazzyYaz:
PR of discussion https://github.com/celo-org/celo-proposals/pull/234#pullrequestreview-695541766

ankushswar1:
@nategraf Moving the discussion here per @YazzyYaz's recommendation:

Thanks for clarifying all of my questions 🙏 I think a lot of my feedback/confusion was repeated throughout multiple comments, so I'll try to consolidate some of my lingering thoughts here:

1. I think it would be helpful to clarify upfront that there are a finite set of domain `name`s, each with its own ruleset, and a client cannot arbitrarily specify a new domain `name`. You mention that an unrecognized domain `name` would lead to an error at the HTTP transport layer, but I think it may be helpful to mention this at a high-level in the introduction of what the domain struct does. 

2. I still have a little bit of confusion on this line from the `Key management` section:

> Because ODIS' current OPRF function operates with the same cryptographic primitives as the proposed function, and because its a bad idea in general to reuse keys, implementation of this proposal MUST use a new set of keys, not used in any other service.

I initially interpreted this as each domain requires a new set of keys - I think I may have assumed this because you mentioned that different domains should produce different outputs on the same input, but it wasn't clear whether this was accomplished through a salt or simply different keys - it may be helpful to suggest how that might work as you did [here](https://github.com/celo-org/celo-proposals/pull/234#discussion_r660877917). 

My understanding after reading your responses is that this proposal would only require **one** new DKG, and the same keys will be used for each domain (and thus new domains don't require a new DKG). The new DKG is simply necessary because the proposal modifies ODIS significantly, and because it's a bad idea to reuse keys. Is that correct? It might be helpful to explicitly state that the same keys are used for different domains, and that the distinct outputs of different domains comes from something inherent to the domain struct (e.g. a salt), as opposed to using different signing keys for each domain.

nategraf:
1. Good callout. I've tried to address this in 93a66cca80bdfb3dacb561271bea15a305ca9c03
2. I've opened up a PR to add a note to try to make this a bit more clear. PTAL https://github.com/celo-org/celo-proposals/pull/236

PR #234, opened 2021-06-16
[DRAFT] Extension to ODIS for Password Hashing

nategraf:
Draft proposal for an extension to the ODIS phone number privacy service to allow for its use for
hardening passwords such that they can be used as encryption and authorization keys.

Please send discussion to https://github.com/celo-org/celo-proposals/issues/233

### DIFF
--- a/CIPs/cip-40.md
+++ b/CIPs/cip-40.md
@@ -1,5 +1,5 @@
 ---
-cip: <to be assigned>
+cip: 40
 title: Extension to ODIS for Password Hashing
 author: Victor Graf <victor@clabs.co>, Cody Born <cody@clabs.co>, Alec Schaefer <alec@clabs.co>
 discussions-to: https://github.com/celo-org/celo-proposals/issues/233
--- a/CIPs/cip-draft_poprf.md
+++ b/CIPs/cip-draft_poprf.md
@@ -296,8 +296,6 @@ type SmartContractDomainOptions = {
 
 ### Domain serialization
 
-<!-- TODO(victor): Refactor this section a bit to describe a function to map a subset of TS types to
-EIP-712 types -->
 Because any variation in the domain string, by design, results in an unpredictable change in the
 output, there must be a canonical (i.e. deterministic) serialization for each given domain string.
 Further, there must not be a collision in the encoded domain strings between two semantically
@@ -349,6 +347,7 @@ const domain: LinearBackoffDomain = {
 }
 ```
 
+<!-- TODO(victor) Verify that there are no mistakes in this example. -->
 ```text
 domainSeparator = keccak256(
   keccak256("EIP712Domain(string name,string version)") ||
@@ -386,10 +385,13 @@ Domains that make use of signatures for authentication of requests (e.g. the `Pe
 example below) should include a field in the domain options for specification of a signature.
 Additional fields may also be added, such as a nonce to prevent replaying requests.
 
-<!-- TOOD(victor): Add a specification of the EIP-712 domain to be used for here -->
 Signatures should be verified over the entire request, excluding the signature itself. Specifically,
 the message against which a signature is verified should be obtained by first removing the signature
 from the full `GetDomainRestrictedSigRequest` object, then serializing it via EIP-712.
+
+Serializing the request with EIP-712 should use the domain `{ name: "ODIS Domain Restricted
+Signature Request", version: "1" }` and type defined by applying the type mapping [defined
+above](#mapping-typescript-to-eip-712-types) to the `GetDomainRestrictedSigRequest` type.
 
 ### HTTP Transport
 
--- a/CIPs/cip-draft_poprf.md
+++ b/CIPs/cip-draft_poprf.md
@@ -22,40 +22,46 @@ described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.html).
 
 ## Simple Summary
 
-An extension to the ODIS phone number privacy service to allow for its use for hardening passwords
-such that they can be used as encryption and authorization keys.
+An extension to the ODIS phone number privacy service to support granular rate limiting with the
+goal of hardening passwords for use as encryption and authorization keys.
 
 ## Abstract
 
 <!-- TODO(victor): Once we have a roughly agreed upon proposal for the pOPRF, link it here -->
 Given the required cryptographic functionality (to be discussed in more detail in a future
-proposal), this proposal describes an interface to be exposed by the ODIS service to allow for
-rate-limited hash queries against user-defined domains.
+proposal), this proposal describes an interface to be exposed by the ODIS service allowing the user
+to specify a domain string (i.e. a salt value) with rate-limiting rules to be combined with their
+blinded input (e.g.  phone number or password).
 
 Central to the proposed interface are domains. A domain is a struct which specifies rate limiting
-information and uniquely identifies the context of the secret to be hashed. 
+information and uniquely identifies the context of the secret to be hashed. As an example, a domain
+for hashing an account password might specify the username (context) and a cap of 10 password
+attempts (rate-limiting parameter). These would be combined with their password to form the final
+output.
 
 Queries with distinct domain specifiers will receive uncorrelated output. For example, output from
-ODIS when requesting `{ domain: "tel", message: "+18002738255" }` will be distinct from and
-unrelated to the output when requesting `{ domain: "password", message: "+18002738255" }`.
+ODIS with a phone number domain and message `+18002738255` will be distinct from and
+unrelated to the output when requesting with a password domain and message `+18002738255`.
 
-Additionally, domain specifiers will control what rate limiting rules are applied. Combined with the
-fact that distinct domains produce uncorrelated output, this allows for the domains to be used as
-distinct OPRFs with custom rate limiting rules. For example the `tel` domain could use the current
-ODIS rate limiting rules, while the `password` domain may use much stricter rules.
+Additionally, domain specifiers will control what rate limiting rules are applied. For example the
+phone number domain could use the current ODIS rate limiting rules, while the password domain may
+use much stricter rules. Combined with the fact that distinct domains produce uncorrelated output,
+this allows for the domains to be used as distinct OPRFs with custom rate limiting rules. 
 
 In order to make this scheme flexible, allowing for user-defined tuning of rate-limits and the
 introduction of new rate limiting and authorization rules in the future, domains are defined as
-serializeable structures.
+serializeable structs.
 
 ## Motivation
 
 Effective [offline password cracking](https://en.wikipedia.org/wiki/Password_cracking) techniques
 limit the use of passwords to derive encryption or authentication keys. Rate-limited hashing can be
-used to make it much more difficult to crack a password. ODIS, the phone number privacy service on
-Celo, implements a rate-limited hash to protect users' phone numbers from being extracted from their
-on-chain identifiers. With some extensions, ODIS can additionally be used to provide a rate-limited
-hash for passwords, as well as many more use cases.
+used to make it much more difficult to crack a password. Computationally expensive password hashing
+functions, such as `PBKDF` and `scrypt`, are commonly used for this purpose, but provide [limited
+protection](https://arxiv.org/abs/2006.05023) and are expensive to run on end-user devices. ODIS,
+the phone number privacy service on Celo, implements a rate-limited hash to protect users' phone
+numbers from being extracted from their on-chain identifiers. With some extensions, ODIS can
+additionally be used to provide a rate-limited hash for passwords, as well as many more use cases.
 
 [Phone Number Privacy](https://docs.celo.org/celo-codebase/protocol/identity/phone-number-privacy)
 
@@ -67,9 +73,9 @@ input (e.g. phone number or password). Unfortunately, it only allows for a singl
 scheme.
 
 In order to support secure password hashing, granular rate limits are invaluable. In order to apply
-rate limits on a per-user basis (e.g. restricting the number of guesses on a user's password to a
+rate limits on a per-context basis (e.g. restricting the number of guesses on a user's password to a
 constant number) ODIS' OPRF scheme must be extended to allow the inclusion of a domain specifier
-(a.k.a salt) that is visible to the service.
+that is visible to the service.
 
 ## Specification
 
@@ -133,14 +139,14 @@ interface Domain {
   // Major version number. Allows for backwards incompatible changes.
   version: number
   // Arbitrary key-value pairs.
-  // Must be serializable to EIP-712 encoding.
+  // MUST be serializable to EIP-712 encoding.
   [key: string]: EIP712Value
 }
 
 interface DomainOptions {
   // Arbitrary key-value pairs.
   // Valid keys and values depend on the domain specification.
-  // Must be serializable to EIP-712 encoding.
+  // SHOULD be serializable to EIP-712 encoding.
   [key: string]: EIP712Value
 }
 ```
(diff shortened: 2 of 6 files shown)


CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,474 @@
...
+
+```solidity
+interface OdisAccessController {
+  function shouldAllowOdisRequest(bytes calldata domainData, bytes calldata authorization) external view returns(bool);
+}
+```
+
+For example, this contract could check that a user owns a certain NFT or has burned a fee before
+being allowed access to the signature. The signature would be applied to the domain and the returned
+value from the contract function.
```
alecps, line 277: Not something we need to worry about now but we should think about how we could ensure backwards compatibility of the smart contract logic (i.e. prevent the logic from changing without a version increase)

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,474 @@
...
+- Specify the rate limiting function to be used in deciding whether to serve the request.
+
+In order to provide flexibility and support future extensions, the domain specifier is structured
+and allows for the usage of multiple domain-specific rate limiting rules, as support by the ODIS
+service.
+
+In the domain specifiers, two fields MUST be provided: `name` and `version`. Together, these select
+which set of rules the service will apply to the request. Rule sets are included in the ODIS
+implementation, and new rule sets may be added over time. Updating a rule set in a way that would
+change the rules applied to existing domains MUST be accompanied by an increase in the `version`
```
alecps, line 168: Currently we have the freedom to tweak rate limits as needed. I.e. if we want to add cEUR balances to quota calculation, we can do that easily. I worry that this requirement may be too strict if it implies that we would need to bump the version number for the domain to make such a change, since bumping the version number changes the output. That is, users who queried ODIS for idenitifiers before we added support for cEUR would not be able to use the updated rate limit without switching to a new identifier. Maybe we can say that any change which _restricts_ a rate limit SHOULD be accompanied by an increase in the version?

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,474 @@
...
+
+Domains that make use of signatures for authentication of requests (e.g. the `PepperedDomain`
+example below) should include a field in the domain options for specification of a signature.
+Additional fields may also be added, such as a nonce to prevent replaying requests.
+
+Signatures should be verified over the entire request, excluding the signature itself. Specifically,
+the message against which a signature is verified should be obtained by first removing the signature
+from the full `GetDomainRestrictedSigRequest` object, then serializing it via EIP-712.
+
+Serializing the request with EIP-712 should use the domain `{ name: "ODIS Domain Restricted
```
alecps, line 401: this is just an example domain, right?

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,474 @@
...
+Signatures are a useful primitive for authenticated rate limiting (e.g. allowing a higher rate limit
+to the owners of verified accounts on Celo). In order to facilitate this, the entire request struct
+should be deterministically serializeable, specifically using EIP-712.
+
+Domains that make use of signatures for authentication of requests (e.g. the `PepperedDomain`
+example below) should include a field in the domain options for specification of a signature.
+Additional fields may also be added, such as a nonce to prevent replaying requests.
+
+Signatures should be verified over the entire request, excluding the signature itself. Specifically,
+the message against which a signature is verified should be obtained by first removing the signature
```
alecps, line 398: Does it matter if "removing" the signature means removing the entire field or just setting it to its zero value?

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,474 @@
...
+When requested over HTTP, which is the primary proposed transport, the following HTTP semantics
+SHOULD be followed.
+
+If a given `name` and `version` are not recognized by an ODIS signer, it SHOULD return a `501: Not
+Implemented` error. If a given `name` and `version` are recognized, but are deprecated, it SHOULD
+return a `405: Method Not Allowed` error. Signers SHOULD keep a list of all old domain names and
+version, even if they are no longer supported.
+
+When a user's request is rejected for violating the rate limiting of a domain, the service SHOULD
+respond with `429: Too Many Requests`. If the waiting would allow for additional requests against
```
alecps, line 416: I think we currently use 403 for this but maybe 429 is more appropriate

CIPs/cip-draft_poprf.md
nategraf: That's a good callout. If we decide to implement this domain, we'd need to deal with issues like that.

CIPs/cip-draft_poprf.md
nategraf: I'd actually apply that rule in the opposite direction. I'm not worried about changes that restrict the rate limit (i.e. such that previously valid requests will not longer be valid), I am worried about changes that loosen the rules. Making a previously invalid request valid means the user is exposed to new risk, which you can argue should only happen by opting in. I agree that its beneficial to have more flexibility though, using cEUR for the phone number domain seems fairly natural, and I'll change this to a SHOULD.

CIPs/cip-draft_poprf.md
nategraf: I was thinking the field should be removed entirely from the struct, but on second thought will change it to set it to zero instead.

CIPs/cip-draft_poprf.md
nategraf: No this is the suggested EIP-712 domain. I'll try to clarify that.

CIPs/cip-draft_poprf.md
nategraf: I think so, especially because the rate limit may have a time component which can be indicated with `Retry-After`.

nategraf:
@YazzyYaz Would you ming assigning a CIP number and approving this for merge when you get a chance?

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,475 @@
...
+
+<!-- TODO(victor): Once we have a roughly agreed upon proposal for the pOPRF, link it here -->
+Given the required cryptographic functionality (to be discussed in more detail in a future
+proposal), this proposal describes an interface to be exposed by the ODIS service allowing the user
+to specify a domain string (i.e. a salt value) with rate-limiting rules to be combined with their
+blinded input (e.g.  phone number or password).
+
+Central to the proposed interface are domains. A domain is a struct which specifies rate limiting
+information and uniquely identifies the context of the secret to be hashed. As an example, a domain
+for hashing an account password might specify the username (context) and a cap of 10 password
```
ankushswar1, line 38: What constitutes a username on the Celo blockchain? Would it be an account's public key, or do we consider the phone number/other forms of social identity to be the "username"? If the latter, what would be the domain for hashing a phone number (since the domain's username is the same as the secret being hashed)?

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,475 @@
...
+protection](https://arxiv.org/abs/2006.05023) and are expensive to run on end-user devices. ODIS,
+the phone number privacy service on Celo, implements a rate-limited hash to protect users' phone
+numbers from being extracted from their on-chain identifiers. With some extensions, ODIS can
+additionally be used to provide a rate-limited hash for passwords, as well as many more use cases.
+
+[Phone Number Privacy](https://docs.celo.org/celo-codebase/protocol/identity/phone-number-privacy)
+
+ODIS' rate-limited hash function is implemented from an
+[oblivious](http://www.cs.columbia.edu/~tal/4261/F19/hugo-columbia-oprf.pdf) [pseudorandom
+function](https://en.wikipedia.org/wiki/Pseudorandom_function_family) (OPRF). In effect, this allows
```
ankushswar1: To clarify, the rate limiting is independent from the hash function, right? As in the OPRF doesn't inherently limit how frequently an input can be hashed, it just ensures that the input is blinded while being hashed?

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,475 @@
...
+<!-- TODO(victor): Once we have a roughly agreed upon proposal for the pOPRF, link it here -->
+Given the required cryptographic functionality (to be discussed in more detail in a future
+proposal), this proposal describes an interface to be exposed by the ODIS service allowing the user
+to specify a domain string (i.e. a salt value) with rate-limiting rules to be combined with their
+blinded input (e.g.  phone number or password).
+
+Central to the proposed interface are domains. A domain is a struct which specifies rate limiting
+information and uniquely identifies the context of the secret to be hashed. As an example, a domain
+for hashing an account password might specify the username (context) and a cap of 10 password
+attempts (rate-limiting parameter). These would be combined with their password to form the final
```
ankushswar1: Is the idea to have the user set a custom rate limit for each domain? Why would this be preferable to having a predetermined constant rate limit (perhaps set by Governance) for each domain type (e.g. password, phone number)? It seems like this could introduce unwanted complexity on the user's end, and could make the rate-limiting scheme ineffective if the user doesn't know what they are doing (e.g. they set a super high number of max password attempts)?

CIPs/cip-40.md
```diff
@@ -0,0 +1,475 @@
...
+respond with `429: Too Many Requests`. If the waiting would allow for additional requests against
+the same domain, the service SHOULD include the `Retry-After` header to specify how long the client
+must wait. ODIS servers MAY apply additional rate limiting, such as by IP address. Any rate limiting
+rules outside of those applied per domain MUST be counted against first, such that a request
+dropped by optional rate limiting rules will not be counted against the domain quota.
+
+### Key management
+
+Because ODIS' current OPRF function operates with the same cryptographic primitives as the [proposed
+function](https://github.com/celo-org/celo-proposals/pull/234), and because its a bad idea in
```
YazzyYaz, line 426: I'd recommend linking to the CIP number over the pull request.

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,475 @@
...
+output.
+
+Queries with distinct domain specifiers will receive uncorrelated output. For example, output from
```
ankushswar1: Is this accomplished by using separate peppers (or salts) based on the domain? As in each ODIS signer holds a set of distinct keys (one for each existing domain) and the combiner generates the pepper by using the domain-specific keys from a threshold number of signers?

CIPs/cip-draft_poprf.md
ankushswar1: Ah, seems like from the `Domain Specification` section below, that users can't specify custom rate limits and rather the ODIS determines the rate limit based on the Celo-wide rule set for the specified domain?

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,475 @@
...
+
+In order to provide flexibility and support future extensions, the domain specifier is structured
+and allows for the usage of multiple domain-specific rate limiting rules, as support by the ODIS
+service.
+
+In the domain specifiers, two fields MUST be provided: `name` and `version`. Together, these select
+which set of rules the service will apply to the request. Rule sets are included in the ODIS
+implementation, and new rule sets may be added over time. Updating a rule set in a way that would
+change the rules applied to existing domains SHOULD be accompanied by an increase in the `version`
+number. Domain names SHOULD be human-readable. Version numbers SHOULD be whole integers.
```
ankushswar1: Does `name` correspond to a specific type of identity/input (e.g. phone number, password) which the protocol can then look up the corresponding rule set/rate-limiting scheme for? Or should the `name` explicitly be the rate-limiting scheme (e.g. `Linear Backoff Domain`)?

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,475 @@
...
+
+With a given `name` and `version`, a number of additional fields MAY be specified to act as
+parameters for the rate limiting rule set (e.g. in a linear backoff rule set, a field may define the
+refresh rate) and to establish the context. In order to ensure deterministic serialization, each
+`name` and `version` MUST correspond to a single type definition.
+
+Changing any field in the domain specifier, results in a new domain with completely unrelated
```
ankushswar1: Should clients specify the `name` of a domain from a pre-determined set of domain names (like a publicly-available list of rate-limiting schemes that Celo has)? If the `name` is different and doesn't correspond to an existing rate-limiting scheme, would the protocol return some valid output (by just generating a different salt or something along those lines) or should the request just fail altogether?

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,475 @@
...
+  notBeforeBlock?: number
+  // Arbitrary bytes to be included in the smart contract call for use by the
+  // contract logic. Intended to be a signature or other authorization string.
+  authorization: string
+}
+```
+
+### Domain serialization
+
+Because any variation in the domain string, by design, results in an unpredictable change in the
```
ankushswar1: Maybe this hasn't been thought out yet (or I misunderstand ODIS), but how exactly would a new domain name correspond to new/different output? My understanding is that ODIS generates a pepper to hash phone numbers from a set of keys created during the DKG ceremony - is this pepper what essentially distinguishes the output of different domains? In that case, wouldn't there need to be a new DKG ceremony for each domain, so how would a completely new domain `name` be handled (since there may not be a set of keys/already-performed DKG corresponding to it)?

CIPs/cip-draft_poprf.md
ankushswar1: I think this is clarified later down the proposal (in `HTTP Transport`), but maybe it makes sense to clarify earlier on that domain `name`s _must_ come from a set of existing domains when getting feedback from the community?

CIPs/cip-draft_poprf.md
```diff
@@ -0,0 +1,475 @@
...
+Incorrect implementation or operation of the ODIS service can have very serious effects on users of
+the service. In particular, a bug in the rate limiting code or compromise of a threshold of nodes
+could allow an attacker to bypass rate limiting and derive a user's secret from its hash or other
+commitment.
+
+Implementing this proposal involves writing new code to be included in the ODIS service, which
+implies a potential increase in the attack surface and opportunities to introduce security
+vulnerabilities. Care should be taken to ensure the correctness of the implementation, including
+testing, careful code review, and potentially third-party audits.
+
```
ankushswar1: Maybe an open question - would the parameters of the pOPRF vary with the domain type? For instance, would password domains use the same set of signers and threshold number of signatures as the phone number domains? Or does it make more sense to vary these parameters based on the domain (for ex, password domains may require a higher threshold of signatures for stronger security guarantees/decreased risk of collusion by a few malicious signers)?

CIPs/cip-draft_poprf.md
ankushswar1: Seems to be clarified in the `HTTP Transport` section

CIPs/cip-draft_poprf.md
ankushswar1: Clarified below - rate limiting is part of the HTTP layer, not the OPRF crypto primitive.

CIPs/cip-draft_poprf.md
nategraf: Oh and to answer your direct question, both work as long as implementations are consistent.

CIPs/cip-draft_poprf.md
nategraf: It's not a global concept, simply application specific. I'll try to clarify that.

CIPs/cip-draft_poprf.md
nategraf: Right, but I might clarify further that HTTP is simply the transport (gRPC, or WebSockets could be used instead), so it's really at the "ODIS protocol" layer.

CIPs/cip-draft_poprf.md
nategraf: When I say user in this doc, I generally mean the user of ODIS / the application developer. I would expect each application to choose a rate limit appropriate their needs. It's better than setting it centrally because each application may have different security requirements and we don't want to involve Governance where there is not a collective action problem. (i.e. it's perfectly acceptable for different applications / users to use different rate limits and domains, so there is no reason to force them all to be chosen centrally)

nategraf: To clarify on your question, the supported rule sets would be developed and supported centrally as part of the ODIS software, but the parameters for those rulesets can be controlled by the ODIS user. Of course, this blurs a little when your consider some interesting use cases, such as the smart contract domain example, where the parameter is essentially the ruleset.

CIPs/cip-draft_poprf.md
nategraf: I leave this a bit vauge in the abstract, simply stating that the domain is combined with the message, but try to calrify later. Essentially though, the domain acts as a salt, and salts are often thought of as "function selectors" when working with PRFs. (i.e. You can define from function $f(s || x)$ an equivalent function $f_s(x)$ where s is the salt and x is the message.)

nategraf: Does it become clear after reading the rest of the proposal. If not, I'll try to fix that.

CIPs/cip-40.md
nategraf: Oops, that's actually supposed to be a link to Notion for now, and to the crypto CIP later.

CIPs/cip-draft_poprf.md
nategraf: I try to make this clear above, but the `name` identifies a ruleset, so it could be `Linear Backoff Domain` or `Password Domain`, but it cannot include a specific user identifier such as a phone number. (i.e. it must come from a small fixed set, included in the ODIS software). I don't specify whether the name should refer to what the ruleset does (e.g. linear backoff), or how its intended to be used (e.g. to hash passwords). I think its ok to do either and both, but I lean toward the former.

CIPs/cip-draft_poprf.md
nategraf: I'll try to clarify that.

CIPs/cip-draft_poprf.md
nategraf: A single key is used for all domains. Not sure what in the `HTTP Transport` section answers this 😅

CIPs/cip-draft_poprf.md
nategraf: Varying the signer set and/or threshold based on the domain would require a distinct keys for each configuration. We don't provide a way to do this in this proposal. In this proposal the signer set is abstracted away as part of the "cryptographic functionality". If there is a need for this in the future, we could propose a further extension to ODIS.

YazzyYaz (approved):

YazzyYaz:
@ankushswar1 I reccommend moving your questions and discussion to the discussion link here: https://github.com/celo-org/celo-proposals/issues/233

PR #236, opened 2021-06-30
CIP 40 - Address comments to clarify proposal

nategraf:
This PR includes adds a note to clarify that only a single new key-pair is needed to implement the
proposal. Clarification on a comment by @ankushswar1 https://github.com/celo-org/celo-proposals/issues/233#issuecomment-870967376

### DIFF
--- a/CIPs/cip-0040.md
+++ b/CIPs/cip-0040.md
@@ -453,6 +453,12 @@ Multiple candidate implementations exist for this functionality. Linked below is
 cryptographic design for use in the ODIS pOPRF service. Other implementation with the same interface
 may also be used to implement the following service design.
 
+Note that in the proposed interface, and in the function proposal below, only a single public key,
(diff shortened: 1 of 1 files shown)


CIPs/cip-0040.md
```diff
@@ -453,6 +453,12 @@ Multiple candidate implementations exist for this functionality. Linked below is
 cryptographic design for use in the ODIS pOPRF service. Other implementation with the same interface
 may also be used to implement the following service design.
 
+Note that in the proposed interface, and in the function proposal below, only a single public key,
```
ankushswar1, line 456: Looks good to me! I think this clarifies that the new ODIS construction would only need 1 new set of keys, not that each domain needs a new set of a keys. 

nit: I think "shareded" (line 457) is supposed to be "sharded" or "shared"

YazzyYaz (approved):

PR #285, opened 2021-09-08
adds odisRateLimit to CIP-0040 folder

alecps:
Associated Notion Doc https://www.notion.so/clabsco/ODIS-RateLimit-Structures-3719039d08db4ffc9e55a0257e4c7468

PR #287, opened 2021-09-21
Reivew edits for #285

nategraf:
In a final draft review of https://github.com/celo-org/celo-proposals/pull/285, I had the following
edits I would like to suggest.

The most significant edit is to reframe this proposal as a single domain, of potentially many that
will be implemented, which includes its own structure for rate limiting. The current framing in the
proposal puts it forward as a rate limit that is to be used across the broadest possible set of
domains, but this is not really in line with the design of CIP-40. Instead I am recomending that
this be reframed as a domain with integrated arte limit structure to solve the immediate challenge
of Cloud Backup PIN hashing, and use cases that can be additionally addressed with limited added
complexity. It is the intention of CIP-40 that additional domains with their own rate limiting rules
should be proposed and integrated as needed.

Other suggested changes are minor and include:
* Moving the mathmatical expressions extension section to the end.
* Renaming AuthenticatedRateLimitDomain to SequentialDelayDomain.
* Making the `publicKey`, and therefore `signature`, fields optional.
* Renaming `context` to `salt` and removing the bound value from the type.
* Changing the code block tags from `tsx` to `typescript`.

### DIFF
--- a/CIPs/CIP-0040/odisRateLimit.md
+++ b/CIPs/CIP-0040/odisRateLimit.md
@@ -1,14 +1,19 @@
 # CIP-40 Sequential Delay Domain
 
+## Simple Summary
+
+CIP-40 domain supporting signature-authenticated rate limits defined as a series of time-delayed stages.
+
 ## Background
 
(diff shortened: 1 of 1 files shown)


alecps (approved):
Looks great! Thanks

PR #318, opened 2023-03-03
CIP-53: Remove minimum client version

karlb:


palango:
Thanks for creating this CIP!

The discussion is available at https://github.com/celo-org/celo-proposals/discussions/361 . Please add that link to the CIP's header.

The next free CIP number is 53, please also add this to the header.

karlb:
Done. Should I also change the status to "accepted", since no objections came up since the All-Core Devs Call?

palango (approved):

PR #324, opened 2023-03-21
Repo cleanup

carterqw2:
- Retroactively updates statuses of CIPs to their real values
- Removes an undocumented tag lang


karlb (approved):
The status is hard to check, but it looks plausible.

PR #388, opened 2023-07-18
Change CIP-52 error handling case

karlb:
The original CIP requires handling three cases:
1. FeeHandler registered
2. FeeHandler not registered but Governance is
3. neither is registered

Having less cases makes the code cleaner, reduces the chances of having an error handling that does not actually work and creates less unusual situations that are hard to understand for people watching the blockchain.

Therefore I suggest removing case 2 and use the same error handler as in case 3: refunding to the sender.


palango (approved):
Makes a lot of sense to me.

PR #389, opened 2023-07-21
Create cip-0062.md

martinvol:


gastonponti:
Closing this proposal that is actually a CGP:
https://github.com/celo-org/governance/blob/main/CGPs/cgp-0087.md

PR #403, opened 2024-04-16
Fix link destination

ezdac:
The CIP-64 link pointed to the wrong blockchain's improvement proposal 🙃


karlb (approved):
Oops, that must have come up when quickly getting the URL from a search. Thanks for catching!

